PR #1, opened 2020-09-08
Update GoFSH for 0.1.0 release

cmoesel:
- Update package.json w/ new name, version, description, contributors, bin, and publisher-related tasks
- Convert GitLab CI config to GitHub CI config
- Update README


package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "fsh-ihsus",
-  "version": "0.0.1",
-  "description": "Ichthymorphic Helper Systematically Undoes Sushi (FSH Decompiler)",
+  "name": "gofsh",
+  "version": "0.1.0",
+  "description": "GoFSH is a FHIR Shorthand (FSH) decompiler, able to convert formal FHIR definitions from JSON/XML to FSH.",
```
ngfreiter, line 4: I don't think we can yet claim XML to FSH.

ngfreiter (commented):
One small change, looks good otherwise!

package.json
cmoesel: I think I was thinking of the description more as the target than the current reality.  But point taken.

cmoesel: Addressed in 53dc60d

ngfreiter (approved):
Take a hike XML!

PR #2, opened 2020-09-24
Generate configuration YAML from ImplementationGuide

mint-thompson:
Completes task [CIMPL-530](https://standardhealthrecord.atlassian.net/browse/CIMPL-530).

When an ImplementationGuide JSON file is present, try to use it to create a config file. Since canonical and fhirVersion are required in the configuration, those are also required from the JSON. Since configuration isn't stored directly in the IG or anywhere else, goFSH assumes that the IG's url is in FSH's preferred format and works from there. Optional fields for id, name, status, and version are also included when available. If more than one ImplementationGuide is found, whichever one is found first is used, and a message is logged.

### DIFF
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -4,6 +4,7 @@ import {
   ExportableInstance,
   ExportableValueSet,
   ExportableCodeSystem,
+  ExportableConfiguration,
   ExportableFlagRule,
   ExportableCardRule,
   ExportableCombinedCardFlagRule
@@ -18,6 +19,7 @@ export class Package {
   public readonly instances: ExportableInstance[] = [];
   public readonly valueSets: ExportableValueSet[] = [];
   public readonly codeSystems: ExportableCodeSystem[] = [];
+  public configuration: ExportableConfiguration;
 
   constructor() {}
 
@@ -28,6 +30,7 @@ export class Package {
       | ExportableInstance
       | ExportableValueSet
       | ExportableCodeSystem
+      | ExportableConfiguration
   ) {
     if (resource instanceof ExportableProfile) {
       this.profiles.push(resource);
@@ -39,11 +42,20 @@ export class Package {
       this.valueSets.push(resource);
     } else if (resource instanceof ExportableCodeSystem) {
       this.codeSystems.push(resource);
+    } else if (resource instanceof ExportableConfiguration) {
+      if (this.configuration) {
+        logger.warning(
+          `Multiple implementation guide resources found in input folder. Skipping implementation guide with ${
+            resource.config.id ? `id ${resource.config.id}` : 'unknown id'
+          }.`
+        );
+      } else {
+        this.configuration = resource;
+      }
     }
   }
 
   // TODO: if more optimization steps are added, break them into separate functions.
-  // TODO: Optimization step: combine CardRule and FlagRule
   // TODO: Optimization step: combine ContainsRule and OnlyRule for contained item with one type
   optimize(processor: FHIRProcessor) {
     logger.debug('Optimizing FSH definitions...');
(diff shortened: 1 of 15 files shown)


ngfreiter (commented):
This looked good to me overall, but I was wondering if the `FSHOnly` flag should be set in the configuration, and that led me down a bit of a rabbit hole of wondering how we want to structure the output of goFSH. Should the output of goFSH be a `FSHOnly` project structure, so just a `sushi-config.yaml` and some `fsh` files in `input/fsh`, or should the output be more similar to the output of `sushi --init`, where the project is ready to be used with the IG Publisher.

If we are going with the `FSHOnly` structure, it ends up raising the question (which I wish I had thought of way earlier) of if we need to generate a config at all, when an IG.json exists. Because SUSHI can extract config from IG.json, so we could just copy it to the output and let SUSHI extract.

But if we aren't going with the `FSHOnly` structure, and we would set up a project structure so that SUSHI is generating an ImplementationGuide.json file, then I do think it is necessary to extract a config, but then in this case the config would need to include more properties so that the SUSHI generated ImplementationGuide.json matches the original input to goFSH. If we are going with this approach, then I don't think it is appropriate to add that to the scope of this PR. I just wanted to bring up these questions that are swirling inside my head in case these questions have already been answered and I'm just forgetting.

cmoesel:
I think even a `FSHOnly` `config.yaml` might be helpful because it gives them a starting point if they want to use more advanced features later.  If we really wanted to, we could put a comment at the top of the file that says if you do not need IG features, you can delete this `config.yaml` (but I don't know if that's necessary).

Regarding "to IG or not to IG": How do you feel about `gofsh --ig /path/to/input`?  So, default to `FSHOnly` but generate a starter IG by request?

cmoesel (changes requested):
Looks good!  I only had one minor suggestion to tighten up the regex a bit when extracting the canonical.

src/processor/Package.ts
```diff
@@ -39,11 +42,20 @@ export class Package {
       this.valueSets.push(resource);
     } else if (resource instanceof ExportableCodeSystem) {
       this.codeSystems.push(resource);
+    } else if (resource instanceof ExportableConfiguration) {
+      if (this.configuration) {
+        logger.warning(
+          `Multiple implementation guide resources found in input folder. Skipping implementation guide with ${
+            resource.config.id ? `id ${resource.config.id}` : 'unknown id'
```
ngfreiter, line 49: Here as a backup, instead of `unknown id` we could consider outputting the `canonical`, since we know that will be defined right?

test/processor/ConfigurationProcessor.test.ts
```diff
@@ -0,0 +1,53 @@
...
+      fs.readFileSync(path.join(__dirname, 'fixtures', 'missing-url-ig.json'), 'utf-8')
+    );
+    const result = ConfigurationProcessor.process(input);
+    expect(result).toBeUndefined();
+    expect(loggerSpy.getLastMessage('warn')).toMatch(
+      /ImplementationGuide missing properties.*url/s
+    );
+  });
+
+  it('should not create a Configuration with a fhirVersion', () => {
```
ngfreiter, line 29: I think this should say: "should not create a Configuration **without** a fhirVersion".

src/processor/Package.ts
ngfreiter: I wasn't seeing a test for the logic of skipping previous IGs here. Could you add a test for that?

ngfreiter (commented):
Looking good, I just had a few very minor comments.

ngfreiter:
@mint-thompson this looks good to me. Once the conflicts are resolved I can give this one last look and approve.

ngfreiter (approved):
Looks good!

cmoesel (approved):

PR #3, opened 2020-09-29
Remove default slicing rules on choice elements

mint-thompson:
Completes task [CIMPL-540](https://standardhealthrecord.atlassian.net/browse/CIMPL-540).

When a choice element adheres to the default slicing rules provided by SUSHI, including those rules is not necessary as long as one of the choices exists. Therefore, if there is a rule with a path to one of those choices, slicing rules on the choice element can be removed if they have the default value applied.

### DIFF
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -11,7 +11,7 @@ import {
 } from '../exportable';
 import { FHIRProcessor } from './FHIRProcessor';
 import { logger } from '../utils';
-import { pullAt } from 'lodash';
+import { pullAt, groupBy, values, flatten } from 'lodash';
 import { fshtypes } from 'fsh-sushi';
 
 export class Package {
@@ -89,9 +89,10 @@ export class Package {
 
   // Choice elements have a standard set of slicing rules applied to them by SUSHI.
   // Therefore, it is not necessary to define that slicing using FSH when one of the choices exists.
+  // If the full set of four default rules exists for the same element, remove those rules.
   private suppressChoiceSlicingRules(): void {
     [...this.profiles, ...this.extensions].forEach(sd => {
-      const rulesToRemove: number[] = [];
+      const rulesToMaybeRemove: number[] = [];
       sd.rules.forEach((rule, i, allRules) => {
         if (rule instanceof ExportableCaretValueRule && rule.path.endsWith('[x]')) {
           const pathStart = rule.path.replace(/\[x\]$/, '');
@@ -110,10 +111,16 @@ export class Package {
               otherRule => otherRule.path != rule.path && otherRule.path.startsWith(pathStart)
             )
           ) {
-            rulesToRemove.push(i);
+            rulesToMaybeRemove.push(i);
           }
         }
       });
+      // if four rules to maybe remove have the same path, then that's a full set of defaults, and they are removed
+      const rulesToRemove = flatten(
+        values(groupBy(rulesToMaybeRemove, i => sd.rules[i].path)).filter(
+          ruleGroup => ruleGroup.length === 4
+        )
+      );
       pullAt(sd.rules, rulesToRemove);
     });
   }
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -6,11 +6,13 @@ import {
   ExportableCodeSystem,
   ExportableFlagRule,
   ExportableCardRule,
-  ExportableCombinedCardFlagRule
+  ExportableCombinedCardFlagRule,
+  ExportableCaretValueRule
 } from '../exportable';
 import { FHIRProcessor } from './FHIRProcessor';
 import { logger } from '../utils';
 import { pullAt } from 'lodash';
+import { fshtypes } from 'fsh-sushi';
 
 export class Package {
   public readonly profiles: ExportableProfile[] = [];
@@ -49,6 +51,7 @@ export class Package {
     logger.debug('Optimizing FSH definitions...');
     this.resolveProfileParents(processor);
     this.combineCardAndFlagRules();
+    this.suppressChoiceSlicingRules();
   }
 
   private resolveProfileParents(processor: FHIRProcessor): void {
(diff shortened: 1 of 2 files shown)


src/processor/Package.ts
```diff
@@ -83,4 +86,35 @@ export class Package {
...
+                rule.value.code === 'open')) &&
+            allRules.some(
+              otherRule => otherRule.path != rule.path && otherRule.path.startsWith(pathStart)
+            )
+          ) {
+            rulesToRemove.push(i);
+          }
+        }
+      });
+      pullAt(sd.rules, rulesToRemove);
```
cmoesel, line 117: This considers each standard choice slicing rule individually; _not_ as a set.  So if we had something like:
```
* value[x] ^slicing.discriminator[0].type = #value
* value[x] ^slicing.discriminator[0].path = $this
* value[x] ^slicing.ordered = false
* value[x] ^slicing.rules = #open
```
then this would _keep_ the first rule but remove the others.  I don't think it makes sense to output _partial_ choice slicing rules.  I think we need to consider them as a set.  If all 4 rules match, we remove them; otherwise, we keep them all.

test/processor/Package.test.ts
```diff
@@ -110,5 +112,98 @@ describe('Package', () => {
...
+
+    it('should not remove caret value rules on a choice element that apply nonstandard choice slicing', () => {
+      const profile = new ExportableProfile('SlicedProfile');
+      profile.parent = 'Observation';
+      const slicingType = new ExportableCaretValueRule('value[x]');
+      slicingType.caretPath = 'slicing.discriminator[0].type';
+      slicingType.value = new fshtypes.FshCode('value');
+      const slicingPath = new ExportableCaretValueRule('value[x]');
+      slicingPath.caretPath = 'slicing.discriminator[0].path';
+      slicingPath.value = 'Observation.code';
```
cmoesel, line 195: I know this is intentionally not using `$this` as the `path` -- but `Observation.code` is a completely invalid slicing path in this context -- because the path needs to be a valid _relative_ path to the element.  I like to use valid examples in the tests as much as possible, so I'd suggest setting the path to `id` instead (since it is a valid subpath of `value[x]`.

test/processor/Package.test.ts
```diff
@@ -110,5 +112,98 @@ describe('Package', () => {
...
+      slicingOrdered.value = true;
+      const slicingRules = new ExportableCaretValueRule('value[x]');
+      slicingRules.caretPath = 'slicing.rules';
+      slicingRules.value = new fshtypes.FshCode('closed');
+      profile.rules.push(slicingType, slicingPath, slicingOrdered, slicingRules);
+      const myPackage = new Package();
+      myPackage.add(profile);
+      myPackage.optimize(processor);
+      expect(profile.rules).toHaveLength(4);
+    });
```
cmoesel, line 207: If you agree w/ my suggestion above that rule removal should be all or nothing, we should add a test that differs from the standard choice slicing by only one rule.

cmoesel (changes requested):
I think the overall approach is sound but we should consider the rules as an "all or nothing" proposition.  Either all four slicing rules match (for a given path) and we remove them, or we don't remove them at all... See individual comments for details.

src/processor/Package.ts
ngfreiter: I'm a little shaky on this. As far as I can tell, if `value[x]` did happen to have just "slicing.ordered" defined on it and set to `true`, SUSHI would maintain that when automatically adding the type slicing. So if the user did:
```
Profile: MyObservation
Parent: Observation
* value[x] ^slicing.rules = #closed
* valueString = "hello"
```
The resulting slicing is:
```
"slicing": {
          "discriminator": [
            {
              "type": "type",
              "path": "$this"
            }
          ],
          "ordered": false,
          "rules": "closed"
        }
```
But then GoFSH is going to read that in, and not all rules will match, so it will output all four of the `CaretRules`, when really only the one that differs from the defaults, setting `rules` to `closed` is necessary.

src/processor/Package.ts
```diff
@@ -83,4 +86,42 @@ export class Package {
...
+            ((rule.caretPath === 'slicing.discriminator[0].type' &&
+              rule.value instanceof fshtypes.FshCode &&
+              rule.value.code === 'type') ||
+              (rule.caretPath === 'slicing.discriminator[0].path' && rule.value === '$this') ||
+              (rule.caretPath === 'slicing.ordered' && rule.value === false) ||
+              (rule.caretPath === 'slicing.rules' &&
+                rule.value instanceof fshtypes.FshCode &&
+                rule.value.code === 'open')) &&
+            allRules.some(
+              otherRule => otherRule.path != rule.path && otherRule.path.startsWith(pathStart)
```
ngfreiter, line 111: This line is checking that some choice is defined right? So something like `valueString` if the `[x]` element is `value[x]`? If so, I'm wondering if this is a strict enough check. What if the other rule is just something like `value[x] MS`. Wouldn't that  also trigger this case?

ngfreiter (commented):
I had one question below and one question above.

src/processor/Package.ts
mint-thompson: In that case, `otherRule` and `rule` have the same path, so the condition would not be satisfied.

src/processor/Package.ts
cmoesel: Ultimately it's a stylistic thing, I guess.  But I feel like having GoFSH output partial slicing definitions, based on inference of what SUSHI would do, could confuse users.  They might wonder, why did SUSHI only output the slicing ordered rule but not the others?  And I *feel* like if FSH Authors wanted to override parts of a choice slicing, they would likely define the whole thing -- partly because most authors probably don't even know how SUSHI handles this -- so they wouldn't know to just put in slicing.ordered and not worry about the rest.

I guess I'm just saying that I don't think partially overriding inferred slicings is a best practice or a reasonable thing to expect authors to do. Outputting the full slicing construct is not harmful and allows the full context of the slicing to be understood.  At least that's my thoughts on this. ;-)

ngfreiter: Yeah I'm fine with outputting more rules for clarity.

src/processor/Package.ts
ngfreiter: That's true, I am still a little worried by this though, because doesn't it rely on the idea that no other element starts with the same path as the choice rule without `[x]`. I know on SUSHI we've gotten into trouble with two elements with out logic having bugs when one element's name is contained within that of another. Although maybe it is impossible to have an element that starts with the same string as the choice element, because then that element would look like a specific choice. So I think this is most likely safe, but I'm just not 100% sure.

mint-thompson: Hmm, thinking about it more carefully, I think you're right. We might be looking at `value[x]` and find another rule with a path of `value[x].id`, which satisfies this condition. So this does need to be tightened up.

cmoesel:
Looks good.  Tested with output from mCODE and it works brilliantly.

cmoesel (approved):

ngfreiter (approved):
Thanks for the change, this looks good!

PR #4, opened 2020-09-30
Extract ObeysRule and Invariant

mint-thompson:
Completes task [CIMPL-513](https://standardhealthrecord.atlassian.net/browse/CIMPL-513).

When an element has a constraint array, extract one obeys rule for the element and one invariant for each constraint. In FSH, a single obeys rule may contain multiple constraints, but in SUSHI, an obeys rule contains only one constraint. Therefore, an ExportableObeysRule maintains a list of constraint keys.

Maintain invariants along with the profile or extension that refers to them in order to keep them grouped together when exporting FSH.


src/processor/StructureDefinitionProcessor.ts
```diff
@@ -58,11 +60,16 @@ export abstract class AbstractSDProcessor {
           ValueSetRuleExtractor.process(element)
         );
       }
+      const { obeysRule, invariants } = ObeysRuleExtractor.process(element);
+      newRules.push(obeysRule);
+      newInvariants.push(...invariants);
```
cmoesel, line 65: I'm not sure how I feel about this.  I understand that in formal FHIR definitions, invariants are inlined in each resource -- but in FSH, invariants have global scope.  They can be re-used between resources.

For this reason, I might consider having two extractors -- an `ObeysRuleExtractor` and a `InvariantExtractor`.  This way the `ObeysRuleExtractor` follows the same pattern as the others (returning just an array of rules) and would fit more cleanly in the `if/else` above.  It would also allow the `InvariantExtractor` (or code around the `InvariantExtractor`) to check for duplicate invariants and only push to the invariants if it's unique by key.

In fact, we might want some error handling to detect if somehow (in some way) two invariants w/ the same key had differing definitions.  In that case, I'd issue an error and indicate we're only keeping the first one.

Whether or not we split this into two extractors is probably a matter of opinion.  I'm glad to discuss it.  But either way I think we need to properly deduplicate invariants here.

cmoesel (changes requested):
I'd consider splitting this into two invariants.  Whether or not we do, however, I think we need to check for duplicate invariant definitions.

src/processor/FHIRProcessor.ts
```diff
@@ -13,22 +20,31 @@ export class FHIRProcessor {
...
       // Profiles and Extensions are both made from StructureDefinitions
+      // Invariants may be contained within StructureDefinitions
       this.structureDefinitions.push(rawContent);
       if (rawContent.type === 'Extension') {
         logger.debug(`Processing contents of ${inputPath} as Extension.`);
-        return ExtensionProcessor.process(rawContent, this.fhir);
+        const extension = ExtensionProcessor.process(rawContent, this.fhir);
+        if (extension) {
+          return [extension, ...InvariantProcessor.process(rawContent)];
+        }
```
cmoesel, line 37: This isn't exactly what I expected, but I'm not sure it's _wrong_ -- it's just not what I expected.  I think it depends on how you view the processors.  Does the name of the processor indicate what it _processes_ or what it _returns_?  I.e., does `ExtensionProcessor` (which extends `StructureDefinitionProcessor`) _process_ Extension SDs (and therefore should return whatever it finds) or does it _return_ Extensions (and therefore should return only `Extension`s)?

I was thinking that an `ExtensionProcessor` processes Extension SDs, and therefore, it would return all the things that can result from processing Extension SDs: `ExportableExtension`, `ExportableInvariant`s, (eventually) `ExportableMapping`s, etc.  So rather than having a separate `InvariantProcessor`, we would add an `extractInvariants` function to the `StructureDefinitionProcessor` and its `process` function would now call that as well, to collect all the invariants -- and eventually return them along with the extension.

This is really a design question about what classes do what things where - and will establish a pattern for future features -- so it's probably worth some thought and discussion.  Glad to hear input from @mint-thompson, @ngfreiter, and @jafeltra on this one.

src/rule-extractor/ObeysRuleExtractor.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export class ObeysRuleExtractor {
+  static process(input: ProcessableElementDefinition): ExportableObeysRule {
+    const invariantKeys: string[] = [];
+    if (input.constraint?.length > 0) {
+      input.constraint.forEach((constraint, i) => {
+        invariantKeys.push(constraint.key);
+        // The ObeysRule only contains the key, but these other attributes
+        // will be handled by the InvariantProcessor. So, they should not be
+        // used for CaretValueRules.
```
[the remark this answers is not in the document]
cmoesel: Thinking about this again after leaving a long comment about where the invariant processing code should go (sorry) -- but if the invariant processing code were in the `StructureDefinition.extractInvariants` function, then as long as the `StructureDefinition.process` function called it _before_ the `StructureDefinition.extractRules` function, then `extractInvariants` could mark these paths as _processed_ instead of having to do it here.

cmoesel (commented):
Looking good.  This, however, led to a realization about some assumptions I had made and ultimately to a design question...

src/processor/FHIRProcessor.ts
mint-thompson: Upon further reflection, I think I like your pattern of "A processor is named for what it takes as input." So invariants would get extracted, not processed (since they're not a top-level entity in the JSON). Handling them before any rules lets `InvariantExtractor` set the `processedPaths` appropriately to avoid extra caret value rules, since `ObeysRuleExtractor` doesn't care what paths have already been processed, and there's no harm in adding a duplicate entry to `processedPaths`.

A processor is named for its input and may have various outputs (Profile, Invariants, etc.) based on what exactly is found in the input. An extractor is named for its output and may have various inputs based on what exactly is needed (element, SD, FHIR defs, etc.) to generate the outputs. I feel like these can be consistent rules.

Relatedly, I think that the two separate `ProfileProcessor` and `ExtensionProcessor` can probably be safely refactored and removed in favor of a concrete `StructureDefinitionProcessor`. But, that's a completely separate task from this one.

cmoesel: I think that all makes sense to me, and you've explained it in a way that I think we can generally apply in the future.

src/extractor/ObeysRuleExtractor.ts
```diff
@@ -0,0 +1,30 @@
...
+        // The ObeysRule only contains the key, but these other attributes
+        // will be handled by the InvariantExtractor. So, they should not be
+        // used for CaretValueRules.
+        input.processedPaths.push(
+          `constraint[${i}].key`,
+          `constraint[${i}].human`,
+          `constraint[${i}].severity`,
+          `constraint[${i}].expression`,
+          `constraint[${i}].xpath`
+        );
```
cmoesel, line 20: Can't we move this to the `InvariantExtractor` (where it arguably belongs) as long as we ensure that `InvariantExtractor` gets called before the rule extractors?

src/processor/ExtensionProcessor.ts
```diff
@@ -1,16 +1,21 @@
...
+  static process(
+    input: any,
+    fhir: fhirdefs.FHIRDefinitions
+  ): [ExportableExtension, ...ExportableInvariant[]] | [] {
     if (ExtensionProcessor.isProcessableStructureDefinition(input)) {
       const extension = new ExportableExtension(input.name);
       ExtensionProcessor.extractKeywords(input, extension);
       ExtensionProcessor.extractRules(input, extension, fhir);
-      return extension;
+      const invariants = ExtensionProcessor.extractInvariants(input);
```
cmoesel, line 16: If we move this up a few lines, it will allow the `InvariantProcessor` to declare the paths it has processed.

src/processor/ProfileProcessor.ts
```diff
@@ -1,14 +1,19 @@
...
+  static process(
+    input: any,
+    fhir: fhirdefs.FHIRDefinitions
+  ): [ExportableProfile, ...ExportableInvariant[]] | [] {
     if (ProfileProcessor.isProcessableStructureDefinition(input)) {
       const profile = new ExportableProfile(input.name);
       ProfileProcessor.extractKeywords(input, profile);
       ProfileProcessor.extractRules(input, profile, fhir);
-      return profile;
+      const invariants = ProfileProcessor.extractInvariants(input);
```
cmoesel, line 14: As noted in the ExtensionProcessor, if we bump this up a few lines it will allow `InvariantExtractor` to declare its own processed paths.

cmoesel (changes requested):
I'd still prefer to have the `InvariantExtractor` mark the paths it processes rather than the `ObeysRuleExtractor`...

src/extractor/ObeysRuleExtractor.ts
mint-thompson: We could, but to do that, we'd need to do some additional refactoring to make rules and invariants get extracted from the same `ProcessableElementDefinition` instances. Right now, `extractRules` and `extractInvariants` create those instances locally, so they disappear along with their `processedPaths` at the end of each function call. Given that the solution in `ObeysRuleExtractor` is admittedly a little magical, what do you think of this idea:
- In the `process` method in `ProfileProcessor` and `ExtensionsProcessor`, build a `ProcessableElementDefinition[]` from `input.differential.element`.
- Modify the `extractRules` and `extractInvariants` methods to take a `ProcessableElementDefinition[]` as a parameter.

That way, these extractors (and any future extractors) can keep the track of `processedPaths` from start to finish. I'll try writing this and see where it gets me.

jafeltra (approved):
This makes sense to me! I admittedly did not follow all the iterations of this in real time, but reading through the comments and looking at the current branch, this makes logical sense to me and works as expected! I didn't notice anything additional!

cmoesel (commented):
I've run GoFSH 0.2.0 and this PR's branch against US Core 3.1.0 and compared the results.  The two major things I've learned from this are:

1. **US Core constraint keys are not globally unique.**  I had assumed that constraint keys should be globally unique, but a [closer reading of the spec](http://hl7.org/fhir/R4/elementdefinition-definitions.html#ElementDefinition.constraint.key) shows this is not necessarily true; they need only be unique within the StructureDefinition!  Yikes.  The FSH spec was designed w/ this assumption in mind, so the FSH spec actually forces all constraints to be unique.  So... this is a problem, because we can't properly convert US Core to FSH without changing the constraint keys (sigh).  We should discuss how we want to approach this.

2. **Without the Invariant / ObeysRule exporters, this commit is a regression.** Since it marks the paths as processed, we no longer get caret rules -- but since the exporters aren't implemented, we don't get obeys rules either.  So we lose the invariants entirely.  This is a short-term issue, but it means that if we merge this PR, GoFSH `master` branch will _not_ be in a releasable state.  We aim to always have `master` be releasable (to allow for quick bug fixes).  So...  I think I would want the exporter PRs also reviewed and approved before merging any of this (or merge the exporters first since they are harmless without the importers).

cmoesel:
Re: the first observation above (lack of global uniqueness for contraint keys), I've posted a question about this on Zulip here: https://chat.fhir.org/#narrow/stream/179177-conformance/topic/Uniqueness.20of.20constraint.20keys

src/extractor/InvariantExtractor.ts
```diff
@@ -1,31 +1,49 @@
...
           invariant.xpath = constraint.xpath;
-          input.processedPaths.push(`constraint[${i}].xpath`);
+          constraintPaths.push(`constraint[${i}].xpath`);
+        }
+
+        // if an invariant with this key already exists, don't make a new invariant with the same key.
+        // if the new invariant would be an exact match of the existing invariant, mark the paths as
+        // processed so an ObeysRule is created and no CaretValueRules are created.
+        // if the new invariant has a key match but isn't an exact match, it will be created using CaretValueRules.
+        const matchingKeyInvariant = existingInvariants.find(inv => inv.name === constraint.key);
```
cmoesel, line 38: Should we also be checking against the local `invariants` array?  I know we are not a FHIR validator, but as it is, if there is a duplicate key in this element, I think it will result in two `Invariant`s being created -- which will very likely fail a round trip.

src/extractor/ObeysRuleExtractor.ts
```diff
@@ -7,8 +7,10 @@ export class ObeysRuleExtractor {
     const invariantKeys: string[] = [];
     if (input.constraint?.length > 0) {
       input.constraint.forEach((constraint, i) => {
-        invariantKeys.push(constraint.key);
-        input.processedPaths.push(`constraint[${i}].key`);
+        // if the element's key was processed, that means an Invariant exists, and we should add an ObeysRule.
```
cmoesel, line 10: Cool.  I like that logic.

src/processor/ExtensionProcessor.ts
```diff
@@ -8,7 +8,8 @@ import { ProcessableElementDefinition } from '.';
 export class ExtensionProcessor extends AbstractSDProcessor {
   static process(
     input: any,
-    fhir: fhirdefs.FHIRDefinitions
+    fhir: fhirdefs.FHIRDefinitions,
+    existingInvariants: ExportableInvariant[] = []
```
cmoesel, line 12: Kind of a bummer that we need to pass this all around now -- but I don't really see a better option.  Oh well.

cmoesel (changes requested):
I think the approach will work!  I just found one slight issue we probably want to address (but is easy to fix).

src/extractor/InvariantExtractor.ts
cmoesel: I'm not sure if I understand the question.  Perhaps we can discuss after standup.

cmoesel (changes requested):
This looks good -- but in my manual testing of US Core, it shined light on something we missed in the writer PRs: the actual `FSHExporter` does not export the `Invariant`s -- so now we end up with FSH that has `* obeys us-core-1` but no `Invariant` definition for `us-core-1`.

Even though it is outside the scope of this PR, can you update `FSHExporter` to export the `Invariant`s?  Otherwise this produces invalid FSH and should not be in `master`.

cmoesel (approved):
Yes!

cmoesel:
@jafeltra -- did you want to re-review or are OK w/ merging based on my re-review.  Changes were pretty minimal since your last review (if you don't count the changes merged from master).

jafeltra:
I looked through the diffs and I think it makes sense, so I'm good if this gets merged!

PR #5, opened 2020-10-05
Remove default context rules from extensions

cmoesel:
If an Extension definition does not have a context specified by the author, SUSHI will add context equivalent to these rules:
```
* ^context[0].type = #element
* ^context[0].expression = "Element"
```

If GoFSH finds these rules, and they are the only context rules on the extension, they will be removed since they are generated by default by SUSHI.

If you want to test this manually, build a FSH project that has Extensions in it (without author-specified contexts).  Then run GoFSH on its output.  Using the GoFSH master branch, you'll see the `context` rules.  Using this branch, you will not see those rules.

A good manual negative test is running GoFSH against US Core.  Since US Core *does* define specific contexts for each Extension, there will be no difference between running master GoFSH and this branch's GoFSH.

### DIFF
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -6,11 +6,13 @@ import {
   ExportableCodeSystem,
   ExportableFlagRule,
   ExportableCardRule,
-  ExportableCombinedCardFlagRule
+  ExportableCombinedCardFlagRule,
+  ExportableCaretValueRule
 } from '../exportable';
 import { FHIRProcessor } from './FHIRProcessor';
 import { logger } from '../utils';
-import { pullAt } from 'lodash';
+import { fshtypes } from 'fsh-sushi';
+import { isEqual, pullAt } from 'lodash';
 
 export class Package {
   public readonly profiles: ExportableProfile[] = [];
@@ -42,13 +44,12 @@ export class Package {
     }
   }
 
-  // TODO: if more optimization steps are added, break them into separate functions.
-  // TODO: Optimization step: combine CardRule and FlagRule
   // TODO: Optimization step: combine ContainsRule and OnlyRule for contained item with one type
   optimize(processor: FHIRProcessor) {
     logger.debug('Optimizing FSH definitions...');
     this.resolveProfileParents(processor);
     this.combineCardAndFlagRules();
+    this.removeDefaultExtensionContextRules();
   }
 
   private resolveProfileParents(processor: FHIRProcessor): void {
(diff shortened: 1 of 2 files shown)


test/processor/Package.test.ts
```diff
@@ -110,5 +112,87 @@ describe('Package', () => {
...
+      expressionRule2.caretPath = 'context[1].expression';
+      expressionRule2.value = 'CodeSystem';
+      extension.rules = [typeRule, expressionRule, typeRule2, expressionRule2];
+      const myPackage = new Package();
+      myPackage.add(extension);
+      myPackage.optimize(processor);
+      expect(extension.rules).toHaveLength(4);
+    });
+
+    it('should remove default context from profiles', () => {
```
ngfreiter, line 182: This should be "should not remove" right?

src/processor/Package.ts
```diff
@@ -83,4 +84,37 @@ export class Package {
...
+          /^context\[\d+]\.type$/.test(r.caretPath)
+      ).length;
+      if (numContexts === 1) {
+        // * ^context[0].type = #element
+        const typeRuleIdx = sd.rules.findIndex(
+          r =>
+            r instanceof ExportableCaretValueRule &&
+            r.path === '' &&
+            r.caretPath === 'context[0].type' &&
+            isEqual((r as ExportableCaretValueRule).value, new fshtypes.FshCode('element'))
```
ngfreiter, line 104: I think TypeScript can infer that `r` is an `ExportableCaretValueRule`. It seems to be doing that on `112`.

ngfreiter (commented):
Looks good and worked as promised. Two minor comments.

src/processor/Package.ts
cmoesel: Yes, indeed.  That's the result of some refactoring where I initially had that `isEqual` in a separate `if` condition (without the `instanceof`, so it was necessary).  Thanks for finding that, eagle eye @ngfreiter!

cmoesel:
@ngfreiter -- thanks for the review!  Your comments have been addressed in bd9f373.

ngfreiter (dismissed):
Looks good!

mint-thompson (approved):
Looks good. The only thing I'm thinking about is whether the default rule checks in `removeDefaultExtensionContextRules` should compare to some predefined instance of `ExportableCaretValueRule` rather than manually comparing all the fields, but that's really just a matter of personal preference.

cmoesel:
Good idea, @mint-thompson.  I'll give that a try to see if it looks cleaner.

cmoesel:
Merged from master (9e60fdf) and incorporated @mint-thompson's suggestion (a0b4c62).  @mint-thompson -- can I get a follow up review on the change?

mint-thompson (approved):

PR #6, opened 2020-10-05
Use "named" syntax

ngfreiter:
Addresses https://standardhealthrecord.atlassian.net/browse/CIMPL-532.

This PR adds a step to `optimize` that will combine a `ContainsRule` and an `OnlyRule` into a `ContainsRule` that uses the `named` syntax, where appropriate.

For example, the following two rules:
```
extension contains fooExtension 0..1 MS
extension [fooExtension] only http://foo.org/fhir/StructureDefinition/fooExtension
```
can be combined to 
```
extension contains http://foo.org/fhir/StructureDefinition/fooExtension named fooExtension 0..1 MS
```

### DIFF
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -102,7 +102,7 @@ export class Package {
             // on the differential unnecessarily. Using the "named" syntax with "Extension" causes errors in SUSHI.
             // As long as the type is not "Extension", we assume it is a profile of Extension, and we can therefore
             // use the "named" syntax.
-            if (onlyRule && onlyRule.types.length === 1 && onlyRule.types[0].type !== 'Extension') {
+            if (onlyRule?.types.length === 1 && onlyRule?.types[0].type !== 'Extension') {
               item.type = onlyRule.types[0].type;
               rulesToRemove.push(onlyRuleIdx);
             }
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -6,7 +6,9 @@ import {
   ExportableCodeSystem,
   ExportableFlagRule,
   ExportableCardRule,
-  ExportableCombinedCardFlagRule
+  ExportableCombinedCardFlagRule,
+  ExportableContainsRule,
+  ExportableOnlyRule
 } from '../exportable';
 import { FHIRProcessor } from './FHIRProcessor';
 import { logger } from '../utils';
@@ -49,6 +51,7 @@ export class Package {
     logger.debug('Optimizing FSH definitions...');
     this.resolveProfileParents(processor);
     this.combineCardAndFlagRules();
+    this.constructNamedExtensionContainsRules();
   }
 
   private resolveProfileParents(processor: FHIRProcessor): void {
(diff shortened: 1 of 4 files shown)


src/processor/Package.ts
```diff
@@ -83,4 +87,30 @@ export class Package {
...
+          rule.items.forEach(item => {
+            const onlyRuleIdx = sd.rules.findIndex(
+              other =>
+                other.path === `${rule.path}[${item.name}]` && other instanceof ExportableOnlyRule
+            );
+            const onlyRule = sd.rules[onlyRuleIdx] as ExportableOnlyRule;
+            if (
+              onlyRule &&
+              onlyRule.types.length == 1 &&
+              isUri(onlyRule.types[0].type.split('|')[0])
```
cmoesel, line 105: Why does this apply only if the type is a URI?  I think that the way GoFSH works now, all types here would be URIs anyway, but... in theory, shouldn't `extension[birthSex] only uscore-birth-sex` work just as well as `extension[birthSex] only http://hl7.org/fhir/us/core/StructureDefinition/us-core-birthsex`? What is the use case we're trying to _avoid_ here?

cmoesel (commented):
Looks good and code's clean.  I tried it w/ US Core and it worked as expected.

I had one question about the restriction to only URI types.  See my individual comment.

src/processor/Package.ts
ngfreiter: The issue I was hitting was actually one from USCore. Check out the us-core-race SD, and you will see that in the differential element for the `ombCategory` slice, the type is set to `Extension`:
```
"type": [{ "code": "Extension" }]
```
This is not really necessary from USCore, but it is there in the differential and I guess not invalid. My thinking at the time was that any input to GoFSH that should be turned into a "named" rule will be a URL, since it will be specified in `type.profile`. But if we later on optimize rules to point to other defined items by name rather than URL, and we do that optimization before this one, I think it could be a problem.

I could just explicitly rule out `Extension` I think? At the time that felt wrong, but now that I think about it, if that value is anything other than `Extension`, then I think it should be fair to assume that it is referring to some profile on `Extension`. Does that make sense to you?

cmoesel: Thanks for explaining.  Yeah, that's weird -- but I agree, special-casing `Extension` seems like the best way to go.  And then you also don't need the `isValidUri` function anymore.  I'd add a comment, however, explaining why we're special-casing `Extension` (e.g., we've seen it in US Core).

ngfreiter: Ok @cmoesel, 265f269 explicitly rules out `Extension`, and we no longer use the `isUri` function.

cmoesel (dismissed):
Super duper.

src/processor/Package.ts
```diff
@@ -83,4 +86,30 @@ export class Package {
...
+            const onlyRuleIdx = sd.rules.findIndex(
+              other =>
+                other.path === `${rule.path}[${item.name}]` && other instanceof ExportableOnlyRule
+            );
+            const onlyRule = sd.rules[onlyRuleIdx] as ExportableOnlyRule;
+            // Explicitly ignore "Extension" since some IGs (ex: USCore) add a type constraint to Extension
+            // on the differential unnecessarily. Using the "named" syntax with "Extension" causes errors in SUSHI.
+            // As long as the type is not "Extension", we assume it is a profile of Extension, and we can therefore
+            // use the "named" syntax.
+            if (onlyRule && onlyRule.types.length === 1 && onlyRule.types[0].type !== 'Extension') {
```
mint-thompson, line 105: Very minor, but it should be safe to rewrite this statement with some nullish coalescing operators:
```suggestion
            if (onlyRule?.types.length === 1 && onlyRule?.types[0].type !== 'Extension') {
```

It should be safe, since if `onlyRule` is undefined, the first condition will be false.

cmoesel (approved):

PR #7, opened 2020-10-05
Remove default 0..0 rules from extensions

cmoesel:
When SUSHI exports extensions, it creates default 0..0 rules for extension or value[x], whichever path was not used by the extension.  This is because FHIR says that you can only use extension or value[x], but not both.

Now if GoFSH detects that type of situation, it will remove the 0..0 rules since SUSHI should generate them automatically (and they are therefore redundant).

To manually test this, run GoFSH from master on an IG with extensions and then run GoFSH from this branch on the same IG.  You'll note that some of the extension zeroed-out rules are no longer exported using this branch.

### DIFF
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -42,13 +42,12 @@ export class Package {
     }
   }
 
-  // TODO: if more optimization steps are added, break them into separate functions.
-  // TODO: Optimization step: combine CardRule and FlagRule
   // TODO: Optimization step: combine ContainsRule and OnlyRule for contained item with one type
   optimize(processor: FHIRProcessor) {
     logger.debug('Optimizing FSH definitions...');
     this.resolveProfileParents(processor);
     this.combineCardAndFlagRules();
+    this.removeImpliedZeroToZeroCardRules();
   }
 
   private resolveProfileParents(processor: FHIRProcessor): void {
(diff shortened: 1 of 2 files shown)


ngfreiter (approved):
This looks good to me. I think we should maybe considering breaking out the testing of the optimization functions into either their own files, or at least their own describe blocks. Because we have several PRs up right now (and I am guessing more to come) that are adding functions to `optimize`, and it isn't necessarily obvious which tests are testing which functions.

cmoesel:
I was thinking the same exact thing, @ngfreiter.  Since we're in the middle of things right now, with lots of PRs hanging out, I suggest we get them all in first and then refactor the tests into different suites.  Either different describe blocks or maybe even different files (similar to some of our constraint tests in SUSHI).

I was also wondering if we should just put each optimization in its own file -- so we'd have a folder called optimizations and each file would represent a different optimization.  And I was even thinking of something a little crazier: self-registering optimizations.  We could have a function called `registerOptimization` (or something like that) that takes a function and a weight as an argument -- and then in `Package.ts` we just call all of the registered observations.  This would make it easier to add and remove optimizations and also would pave the way for user configurations where they might or might not want certain observations.  Maybe I'm over-engineering it, but I thought it sounded like fun.  But I call dibs on getting to implement it. ;-)

ngfreiter:
Ohh that does sounds fun and like a good idea. But ok get things merged and then refactor, works for me!

mint-thompson (commented):
Requesting one minor change in a test case description.

cmoesel:
OK, @mint-thompson, @ngfreiter -- merged w/ master and comments addressed.

ngfreiter (approved):
👍

PR #9, opened 2020-10-07
Remove default slicing rules

jafeltra:
This PR addresses [CIMPL-529](https://standardhealthrecord.atlassian.net/browse/CIMPL-529) and supports removing the default slicing rules when present. The rules removed are: 

```
* extension ^slicing.discriminator[0].type = #value
* extension ^slicing.discriminator[0].path = "url"
* extension ^slicing.ordered = false
* extension ^slicing.rules = #open
```
These rules can be on extensions on any element, as well as on modifierExtensions. When they are present, there is also a contains rule at the same path, and there are no additional `slicing.discriminator` rules present, these four rules will be removed.

### DIFF
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -278,27 +278,29 @@ export class Package {
     [...this.profiles, ...this.extensions].forEach(sd => {
       const rulesToMaybeRemove: number[] = [];
       sd.rules.forEach((rule, i, allRules) => {
-        if (
-          rule instanceof ExportableCaretValueRule &&
-          (rule.path.endsWith('extension') || rule.path.endsWith('modifierExtension'))
-        ) {
-          const hasSlicingDiscriminatorTypeRule =
-            rule.caretPath === 'slicing.discriminator[0].type' &&
-            rule.value instanceof FshCode &&
-            rule.value.code === 'value';
-          const hasSlicingDiscriminatorPathRule =
-            rule.caretPath === 'slicing.discriminator[0].path' && rule.value === 'url';
-          const hasSlicingOrderedRule =
-            rule.caretPath === 'slicing.ordered' && rule.value === false;
-          const hasSlicingRulesRule =
-            rule.caretPath === 'slicing.rules' &&
-            rule.value instanceof FshCode &&
-            rule.value.code === 'open';
+        if (rule instanceof ExportableCaretValueRule && /(modifierE|e)xtension$/.test(rule.path)) {
+          // * path ^slicing.discriminator[0].type = #value
+          const DEFAULT_SLICING_DISCRIMINATOR_TYPE = new ExportableCaretValueRule(rule.path);
+          DEFAULT_SLICING_DISCRIMINATOR_TYPE.caretPath = 'slicing.discriminator[0].type';
+          DEFAULT_SLICING_DISCRIMINATOR_TYPE.value = new FshCode('value');
+          // * path ^slicing.discriminator[0].value = "url"
+          const DEFAULT_SLICING_DISCRIMINATOR_PATH = new ExportableCaretValueRule(rule.path);
+          DEFAULT_SLICING_DISCRIMINATOR_PATH.caretPath = 'slicing.discriminator[0].path';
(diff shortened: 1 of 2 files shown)


src/processor/Package.ts
```diff
@@ -228,4 +229,59 @@ export class Package {
...
     });
   }
+
+  private removeDefaultSlicingRules(): void {
+    [...this.profiles, ...this.extensions].forEach(sd => {
+      const rulesToMaybeRemove: number[] = [];
+      sd.rules.forEach((rule, i, allRules) => {
+        if (
+          rule instanceof ExportableCaretValueRule &&
+          (rule.path.endsWith('extension') || rule.path.endsWith('modifierExtension'))
```
cmoesel, line 239: Matter of preference, but you could use a regex here:
```ts
/(modifierE|e)xtension$/.test(rule.path)
```

src/processor/Package.ts
```diff
@@ -228,4 +229,59 @@ export class Package {
...
+          const hasContainsRule = allRules.some(
+            otherRule => otherRule instanceof ExportableContainsRule && otherRule.path === rule.path
+          );
+          const hasOneSlicingDiscriminatorRule = !allRules.some(
+            otherRule =>
+              otherRule instanceof ExportableCaretValueRule &&
+              otherRule.caretPath !== 'slicing.discriminator[0].type' &&
+              otherRule.caretPath !== 'slicing.discriminator[0].path' &&
+              otherRule.caretPath.startsWith('slicing.discriminator[')
+          );
```
cmoesel, line 262: Won't this return `false` if there is *any* path that has more than one slicing discriminator?  But really we only care if _this_ path has more than one slicing discriminator.  Right?

src/processor/Package.ts
```diff
@@ -228,4 +229,59 @@ export class Package {
...
+            rule.value instanceof FshCode &&
+            rule.value.code === 'value';
+          const hasSlicingDiscriminatorPathRule =
+            rule.caretPath === 'slicing.discriminator[0].path' && rule.value === 'url';
+          const hasSlicingOrderedRule =
+            rule.caretPath === 'slicing.ordered' && rule.value === false;
+          const hasSlicingRulesRule =
+            rule.caretPath === 'slicing.rules' &&
+            rule.value instanceof FshCode &&
+            rule.value.code === 'open';
```
cmoesel, line 252: The above four work, but @mint-thompson had recommended a slightly different approach in a PR I recently put up.  They suggested creating the default rules and then comparing using `isEqual`.  Here's the result: https://github.com/FHIR/GoFSH/blob/master/src/processor/Package.ts#L172-L179

If you were to consider that, then you would define:
```ts
const DEFAULT_SLICING_DISCRIMINATOR_TYPE_RULE = new ExportableCaretValueRule(rule.path);
DEFAULT_SLICING_DISCRIMINATOR_TYPE_RULE.caretPath = 'slicing.discriminator[0].type';
DEFAULT_SLICING_DISCRIMINATOR_TYPE_RULE.value = new FshCode('value');
```
Then in your `if` statement around line 265, you would have:
```ts
isEqual(rule, DEFAULT_SLICING_DISCRIMINATOR_TYPE_RULE)
```

I'm not sure if it's better in this case or not, but it's something to consider.

cmoesel (changes requested):
I found one thing that I think may be a bug and added two stylistic changes for your consideration.

src/processor/Package.ts
jafeltra: Ah yes. I'll add that check and a test to ensure it works!

src/processor/Package.ts
jafeltra: Ah yes, I'll update to use this approach. I had started off with this, but had changed in when refactoring some things, but I think it makes the rules a lot easier to understand, so I'll update!

jafeltra:
@cmoesel your comments should all be addressed in 134c902

mint-thompson (dismissed):

cmoesel (approved):
Looks.... awesome.

PR #10, opened 2020-10-07
Suppress URL Assignment Rules when possible

ngfreiter:
This suppresses URL assignment rules on extensions when appropriate. There are two cases addressed here. 

The first is that when a new inline extension is created, SUSHI will add a `fixedUri` of the `sliceName` of the extension to the `url` element on that extension. So, if GoFSH creates two rules like this:
```
* extension contains detailed 0..*
* extension[detailed].url = "detailed" (exactly)
```
the `optimize` function will now recognize that SUSHI will infer the second rule, and the second rule is removed.

The second case is for standalone extension definitions. Similar to above, SUSHI will add a `fixedUri` of the `url` of the extension to the `url` element on the extension. So, if on an extension definition, GoFSH creates a  rule like this:
```
* url = "http://hl7.org/fhir/us/core/StructureDefinition/us-core-race" (exactly)
```
And that url matches the one generated by combining the canonical and the Extensions `id`, then the `optimize` function of GoFSH will remove that rule. Note that this optimization requires that the canonical is known. If that information is not available, we cannot know for sure if SUSHI will infer the rule correctly, so we leave the rule on the Extension.

### DIFF
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -10,7 +10,8 @@ import {
   ExportableCombinedCardFlagRule,
   ExportableContainsRule,
   ExportableOnlyRule,
-  ExportableCaretValueRule
+  ExportableCaretValueRule,
+  ExportableFixedValueRule
 } from '../exportable';
 import { FHIRProcessor } from './FHIRProcessor';
 import { logger } from '../utils';
@@ -67,6 +68,7 @@ export class Package {
     this.suppressChoiceSlicingRules();
     this.removeDefaultExtensionContextRules();
     this.removeImpliedZeroToZeroCardRules();
+    this.suppressUrlAssignmentOnExtensions();
   }
 
   private resolveProfileParents(processor: FHIRProcessor): void {
@@ -227,5 +229,45 @@ export class Package {
       });
       pullAt(sd.rules, rulesToRemove);
     });
+  }
+
+  private suppressUrlAssignmentOnExtensions(): void {
+    // Loop over all profiles and extensions, removing assignment rules on inline extensions
+    [...this.profiles, ...this.extensions].forEach(sd => {
+      const rulesToRemove: number[] = [];
+      sd.rules.forEach(rule => {
+        if (rule instanceof ExportableContainsRule && rule.path.endsWith('extension')) {
+          rule.items.forEach(item => {
+            const assignmentRuleIdx = sd.rules.findIndex(
+              other =>
+                other.path === `${rule.path}[${item.name}].url` &&
+                other instanceof ExportableFixedValueRule &&
+                other.fixedValue === item.name
+            );
+            if (assignmentRuleIdx >= 0) {
+              rulesToRemove.push(assignmentRuleIdx);
+            }
+          });
+        }
+      });
+      pullAt(sd.rules, rulesToRemove);
+    });
+    // We must know the configuration to determine if a rule fixing "url" matches the url that SUSHI will assume
+    if (this.configuration?.config?.canonical) {
+      this.extensions.forEach(extension => {
+        const rulesToRemove: number[] = [];
+        extension.rules.forEach((rule, i) => {
+          if (
+            rule instanceof ExportableFixedValueRule &&
+            rule.path === 'url' &&
+            rule.fixedValue ===
+              `${this.configuration.config.canonical}/StructureDefinition/${extension.id}`
+          ) {
+            rulesToRemove.push(i);
+          }
+        });
+        pullAt(extension.rules, rulesToRemove);
+      });
+    }
   }
 }
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -233,15 +233,17 @@ export class Package {
 
   private suppressUrlAssignmentOnExtensions(): void {
     // Loop over all profiles and extensions, removing assignment rules on inline extensions
+    // NOTE: Inline extensions on a profile are allowed by SUSHI, but they are technically not
+    // valid FHIR and the IG Publisher does not like this
     [...this.profiles, ...this.extensions].forEach(sd => {
       const rulesToRemove: number[] = [];
       sd.rules.forEach(rule => {
-        if (rule instanceof ExportableContainsRule && rule.path.endsWith('extension')) {
+        if (rule instanceof ExportableContainsRule && /(modifierE|e)xtension$/.test(rule.path)) {
           rule.items.forEach(item => {
             const assignmentRuleIdx = sd.rules.findIndex(
               other =>
-                other.path === `${rule.path}[${item.name}].url` &&
                 other instanceof ExportableFixedValueRule &&
+                other.path === `${rule.path}[${item.name}].url` &&
                 other.fixedValue === item.name
             );
             if (assignmentRuleIdx >= 0) {
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -66,6 +66,7 @@ export class Package {
     this.constructNamedExtensionContainsRules();
     this.suppressChoiceSlicingRules();
     this.removeDefaultExtensionContextRules();
+    this.removeImpliedZeroToZeroCardRules();
   }
 
   private resolveProfileParents(processor: FHIRProcessor): void {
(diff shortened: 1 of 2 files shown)


src/processor/Package.ts
```diff
@@ -228,4 +230,44 @@ export class Package {
       pullAt(sd.rules, rulesToRemove);
     });
   }
+
+  private suppressUrlAssignmentOnExtensions(): void {
+    // Loop over all profiles and extensions, removing assignment rules on inline extensions
+    [...this.profiles, ...this.extensions].forEach(sd => {
+      const rulesToRemove: number[] = [];
+      sd.rules.forEach(rule => {
+        if (rule instanceof ExportableContainsRule && rule.path.endsWith('extension')) {
```
cmoesel, line 239: I think this applies to `modifierExtension` as well, so the test should probably be:
```ts
/(modifierE|e)xtension$/.test(rule.path)
```

src/processor/Package.ts
```diff
@@ -228,4 +230,44 @@ export class Package {
...
+          rule.items.forEach(item => {
+            const assignmentRuleIdx = sd.rules.findIndex(
+              other =>
+                other.path === `${rule.path}[${item.name}].url` &&
+                other instanceof ExportableFixedValueRule &&
```
cmoesel, line 244: I can't explain why, but I feel like the `instanceof` should be the first thing to check.  It feels like maybe it is less expensive?  Or it more quickly sets the context for people reading the code?  I don't know.  You can ignore this if you want.

src/processor/Package.ts
```diff
@@ -228,4 +230,44 @@ export class Package {
...
+                other.fixedValue === item.name
```
cmoesel, line 245: I'm not sure about this.  The only time it will be `item.name` is if it is a sub-extension of an `Extension` definition.  Because inline extensions are only used in that case.

Most cases will be something like:
```
* extension contains http://hl7.org/fhir/StructureDefinition/patient-religion named religion
* extension[religion].url = http://hl7.org/fhir/StructureDefinition/patient-religion (exactly)
```
which I think would fail this code.

Note this is different than the Standalone Extension case below; as it's not defining an extension, but rather _using_ one in the context of some other definition.

src/processor/Package.ts
```diff
@@ -228,4 +230,44 @@ export class Package {
...
+            if (assignmentRuleIdx >= 0) {
+              rulesToRemove.push(assignmentRuleIdx);
+            }
+          });
+        }
+      });
+      pullAt(sd.rules, rulesToRemove);
+    });
+    // We must know the configuration to determine if a rule fixing "url" matches the url that SUSHI will assume
+    if (this.configuration?.config?.canonical) {
```
cmoesel, line 256: Hmmm... if you had access to the original SD that was used as input, you would be able to just look at `sd.url` instead of having to try to construct it.  It seems like that would be more reliable.

That said, I just did a PR where it also would have been helpful to have access to the SD in the optimization.  But since it was the only case of that, I took a different approach.  But now that we have two cases...

src/processor/Package.ts
```diff
@@ -228,4 +230,44 @@ export class Package {
...
+      this.extensions.forEach(extension => {
+        const rulesToRemove: number[] = [];
+        extension.rules.forEach((rule, i) => {
+          if (
+            rule instanceof ExportableFixedValueRule &&
+            rule.path === 'url' &&
+            rule.fixedValue ===
+              `${this.configuration.config.canonical}/StructureDefinition/${extension.id}`
```
cmoesel, line 264: This will usually be right, but it will be wrong if/when the author overrides the URL with one that is outside the canonical.  In this case, we would actually have an `ExportableCaretValueRule` for that (e.g., `* ^url = "http://foo.org/MyExtension"`).

If we're not going to try to bring the SD down to the optimization code (as discussed in my previous comment) we probably need to accommodate the use case where there might be a caret rule overriding the URL.

test/processor/Package.test.ts
```diff
@@ -627,5 +627,85 @@ describe('Package', () => {
...
+      const assignmentRule = new ExportableFixedValueRule('extension[foo].url');
+      assignmentRule.fixedValue = 'foo';
+      extension.rules = [containsRule, assignmentRule];
+      const myPackage = new Package();
+      myPackage.add(extension);
+      myPackage.optimize(processor);
+      expect(extension.rules).toEqual([containsRule]);
+    });
+
+    it('should remove URL assignment rules on inline extensions on a profile', () => {
```
cmoesel, line 644: While FSH/SUSHI don't expressly forbid inline extensions on profiles, I believe we've determined it really shouldn't be supported.

cmoesel (changes requested):
Found a few things for you to look at.

src/processor/Package.ts
ngfreiter: Yes, but when the extension exists on its own and is used in context of some other definition, SUSHI does _not_ automatically infer that second assignment rule. There is no reason to, the assignment of the `url` is indicated on the definition of the extension itself.

At least that is my understanding from looking at the SUSHI code and from testing. I could be missing something here though so let me know if you have an example of SUSHI doing something I'm missing.

src/processor/Package.ts
ngfreiter: I find this tricky to think about, so I'm disclaimering that I may be off base here. But I'm trying to think that the rule can only be ignored if we know SUSHI will generate that rule. SUSHI generates the `url` assignment rule based on the canonical given in the config, or based on a `CaretValueRule` if one exists. As I understand it, GoFSH will never generate a `CaretValueRule` on `url`, since we let SUSHI handle this. In that case, the original `sd.url` is essentially ignored by GoFSH, and we _do_ really need to check the canonical.

ngfreiter: Now you've got me thinking though, the fact that GoFSH ignores all `CaretValueRules` on `url` is wrong, since we may sometimes have a canonical now. I think that would probably need to be updates so that a `CaretValueRule` is generated if the `url` does not match the one that is going to be generated by SUSHI, given the known canonical.

src/processor/Package.ts
ngfreiter: So as discussed above, my understanding is that such a rule is impossible because `url` is in `IGNORED_PROPERTIES`:
```
const IGNORED_PROPERTIES = [
  'resourceType',
  'id',
  'url', // NOTE: For now, we assume URL matches canonical and is generated
  'name',
  'title',
  'description',
  'fhirVersion',
  /*'mapping', NOTE: mapping should be ignored when a mapping extractor is implemented*/
  'baseDefinition',
  'derivation',
  'snapshot',
  'differential'
];
```

test/processor/Package.test.ts
ngfreiter: While I agree that we have decided it is bad form, I'm not sure what I should change here. It is allowed in SUSHI without any error/warning. Do you think I should remove the processing of profiles altogether? Or just remove this test?

cmoesel: How about this?  Go ahead and leave the logic/test -- but note in a comment that while SUSHI allows this, it is not actually valid FHIR and the IG Publisher will emit errors.  Does that work?

src/processor/Package.ts
cmoesel: Good point.  Although now that we support getting a canonical from the config, we should probably change that so `url` is not ignored anymore.  But that's outside the scope of this PR.

src/processor/Package.ts
cmoesel: I'm apparently reading your comments out of order -- but I agree we should no longer be suppressing the url caret rule 100% of the time -- but only when it differs from the URL SUSHI would generate based on the canonical.

Once the config is generated by inference, we should always have a config and this becomes somewhat of a moot point I guess.  So at that point, I suppose we wouldn't need the sd, but we would still need to check for a caret rule.

src/processor/Package.ts
cmoesel: I think you're right for cases where the extension is not used.  But I think if you constrain something further in the extension, we still get to this case.  For example:
```
Profile: MyPatient
Parent: Patient
* extension contains http://hl7.org/fhir/StructureDefinition/patient-religion named religion 1..1
* extension[religion].valueCodeableConcept = http://foo.org#bar
```
Will produce this differential:
```json
{
  "element": [
    {
      "id": "Patient.extension",
      "path": "Patient.extension",
      "slicing": {
        "discriminator": [
          {
            "type": "value",
            "path": "url"
          }
        ],
        "ordered": false,
        "rules": "open"
      },
      "min": 1
    },
    {
      "id": "Patient.extension:religion",
      "path": "Patient.extension",
      "sliceName": "religion",
      "min": 1,
      "max": "1",
      "type": [
        {
          "code": "Extension",
          "profile": [
            "http://hl7.org/fhir/StructureDefinition/patient-religion"
          ]
        }
      ]
    },
    {
      "id": "Patient.extension:religion.valueCodeableConcept",
      "path": "Patient.extension.valueCodeableConcept",
      "min": 0,
      "max": "1",
      "patternCodeableConcept": {
        "coding": [
          {
            "code": "bar",
            "system": "http://foo.org"
          }
        ]
      }
    }
  ]
}
```
But... now that I think about it,that's an `OnlyRule`, not an `AssignmentRule` -- so hopefully we're handling that in a different optimizer (and if not, we should).

Clearly I should try to avoid doing code reviews after 8pm. ;-)

ngfreiter:
Ok @cmoesel, comments should be addressed.

test/processor/Package.test.ts
```diff
@@ -627,5 +627,101 @@ describe('Package', () => {
...
+      myPackage.add(profile);
+      myPackage.optimize(processor);
+      expect(profile.rules).toEqual([containsRule]);
+    });
+
+    it('should not remove other URL rules on inline extensions on a profile', () => {
+      const profile = new ExportableProfile('MyProfile');
+      const containsRule = new ExportableContainsRule('extension');
+      containsRule.items.push({ name: 'foo' });
+      const assignmentRule = new ExportableFlagRule('extension[foo].url');
```
jafeltra, line 677: This is really small, but I think this variable should be `flagRule`, just for clarity.

jafeltra (commented):
This looks good. I think I followed all the discussion on this PR and I think it makes sense, but I'll still plan to let @cmoesel confirm his comments have been resolved. I didn't notice anything else, just one small suggestion.

cmoesel (approved):
Aside from the minor typo that @jafeltra found, this looks good to me!

jafeltra (approved):
Honestly just glad I passed the test

PR #11, opened 2020-10-07
Extension do not set Parent to Extension URL

jafeltra:
This PR addresses [CIMPL-533](https://standardhealthrecord.atlassian.net/browse/CIMPL-533) and suppresses the Parent on Extensions when the parent is extension, such as:

```
Extension: MyExtension
Parent: http://hl7.org/fhir/StructureDefinition/Extension
```
We were already trying to suppress the parent if it was set to `Extension`, so I added the extension base definition URL to the list to check when suppressing.


cmoesel (approved):
This is a thing of beauty.

cmoesel:
It's so simple, I'm merging it w/ one review.

PR #12, opened 2020-10-07
Don't export card rules for sliced cards that SUSHI calculates

cmoesel:
SUSHI will automatically create a card rule if all of the following is true:
* this is a sliced element
* there is a cardinality rule on at least one of the slices
* the min on this element matches the sum of the mins on all slices

If we reach this condition, don't output a card min since SUSHI will do it.

Technically this is an optimization, but the optimizations don't have access to the full SD, so I put it in the card extractor, which does have access to the full SD.  If we decide we want to give the full SD to optimizers, we can revisit this -- but they'll also need access to all the FHIR definitions.

### DIFF
--- a/test/extractor/CardRuleExtractor.test.ts
+++ b/test/extractor/CardRuleExtractor.test.ts
@@ -68,7 +68,7 @@ describe('CardRuleExtractor', () => {
       expect(element.processedPaths).toEqual([]);
     });
 
-    it('should return a null when there is only a min and it matches the sum of mins of the slices', () => {
+    it('should return a null when there is only a min, and there are card-constrained slices, and min matches the sum of mins of all the slices', () => {
       const element = ProcessableElementDefinition.fromJSON(
         looseSDWithSlices.differential.element[0]
       );
@@ -89,7 +89,7 @@ describe('CardRuleExtractor', () => {
       expect(element.processedPaths).toEqual(['min']);
     });
 
-    it('should return a null when there is only a min and it matches the sum of mins of the slices and the slicing is only in snapshot', () => {
+    it('should return a null when there is only a min, and there are card-constrained slices, and min matches the sum of mins of the all the slices (and the slicing is only in snapshot)', () => {
       // Modify the fixture so the slicing is only in the snapshot (representing the case where it is inherited)
       const clonedSD = cloneDeep(looseSDWithSlices);
       delete clonedSD.differential.element[0].slicing;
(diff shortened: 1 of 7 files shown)


test/extractor/CardRuleExtractor.test.ts
```diff
@@ -43,9 +63,118 @@ describe('CardRuleExtractor', () => {
...
+      // Modify the fixture so the slicing is only in the snapshot (representing the case where it is inherited)
+      const clonedSD = cloneDeep(looseSDWithSlices);
+      delete clonedSD.differential.element[0].slicing;
+      const element = ProcessableElementDefinition.fromJSON(clonedSD.differential.element[0]);
+      const cardRule = CardRuleExtractor.process(element, clonedSD, defs);
+      expect(cardRule).toBeNull();
+      expect(element.processedPaths).toEqual(['min']);
+    });
+
+    it('should return a card rule with min even if it matches the sum of mins of the slices when there are no card-constrained slices in the snapshot', () => {
```
mint-thompson, line 102: If I understand this test case correctly, I think the description should be "when there are no card-constrained slices in the differential".

test/extractor/CardRuleExtractor.test.ts
```diff
@@ -43,9 +63,118 @@ describe('CardRuleExtractor', () => {
...
+      delete clonedSD.differential.element[2];
+      const element = ProcessableElementDefinition.fromJSON(clonedSD.differential.element[0]);
+      const cardRule = CardRuleExtractor.process(element, clonedSD, defs);
+      const expectedRule = new ExportableCardRule('component');
+      expectedRule.min = 3;
+      expect(cardRule).toEqual<ExportableCardRule>(expectedRule);
+      expect(element.processedPaths).toEqual(['min']);
+    });
+
+    it('should return a card rule with min even if it matches the sum of mins of the slices when there are no slices in the snapshot', () => {
```
mint-thompson, line 117: Same here as up on line 137.

mint-thompson (commented):
I imagine that at some point, it may be useful to move `isSliced` to a more general file, but that can happen when it is actually needed.

test/extractor/CardRuleExtractor.test.ts
cmoesel: You're right.  That was the result of changing what the test actually tested too many times!  Fixed in my last commit!

mint-thompson (approved):

jafeltra (approved):
These changes make sense to me and works as expected!

PR #13, opened 2020-10-08
Suppress ^date rules

jafeltra:
This PR addresses [CIMPL-531](https://standardhealthrecord.atlassian.net/browse/CIMPL-531). It suppresses `^date` rules if they appear to be set by the IG Publisher, rather than through a FSH rule. `^date` rules are considered to be set by the IG publisher if:

- all the dates at the root level of the resource are the same across every resource in the package
- the date has a time with hours, minutes, and seconds
- the date has GMT timezone specified (+00:00)

If all three of those conditions are met, the `^date` rule will be removed.

I do have one question about this approach. Should another condition included here be that every resource in the package has a date set? Right now, it will check that any resource with a date has the same date, but I'm thinking it might also be helpful to check that every resource has a date and that the date is the same across the package.

### DIFF
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -374,9 +374,13 @@ export class Package {
       const DEFAULT_DATE = new ExportableCaretValueRule('');
       DEFAULT_DATE.caretPath = 'date';
       DEFAULT_DATE.value = date;
-      [...this.profiles, ...this.extensions, ...this.valueSets, ...this.codeSystems].forEach(resource => {
-        (resource.rules as ExportableRule[]) = (resource.rules as ExportableRule[]).filter(rule => !isEqual(rule, DEFAULT_DATE));
-      });
+      [...this.profiles, ...this.extensions, ...this.valueSets, ...this.codeSystems].forEach(
+        resource => {
+          (resource.rules as ExportableRule[]) = (resource.rules as ExportableRule[]).filter(
+            rule => !isEqual(rule, DEFAULT_DATE)
+          );
+        }
+      );
     }
   }
 }
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -335,13 +335,17 @@ export class Package {
   private removeDateRules(): void {
     let allDatesMatch = true;
     let date: string;
-    [...this.profiles, ...this.extensions].forEach(sd => {
+    [...this.profiles, ...this.extensions, ...this.valueSets, ...this.codeSystems].forEach(sd => {
       for (const rule of sd.rules) {
         if (!allDatesMatch) break; // If any date hasn't matched the others, stop looking at other rules
-        if (rule instanceof ExportableCaretValueRule && rule.caretPath === 'date') {
+        if (
+          rule instanceof ExportableCaretValueRule &&
+          rule.caretPath === 'date' &&
+          rule.path === ''
+        ) {
           const dateValue = rule.value as string; // If the rule assigns a date, the value will be a string.
           if (!date) date = dateValue; // Set the date to match
-          if (date !== rule.value) {
+          if (date !== dateValue) {
             allDatesMatch = false;
           }
         }
@@ -354,14 +358,35 @@ export class Package {
     const dateTimeMatch = date?.match(dateTimeRegex);
     const usesGMT = dateTimeMatch?.[12] === '+00:00'; // dateTimeMatch[12] will hold timezone info if present
     if (allDatesMatch && usesGMT) {
+      const DEFAULT_DATE = new ExportableCaretValueRule('');
+      DEFAULT_DATE.caretPath = 'date';
+      DEFAULT_DATE.value = date;
       [...this.profiles, ...this.extensions].forEach(sd => {
         const rulesToRemove: number[] = [];
         sd.rules.forEach((rule, i) => {
-          if (rule instanceof ExportableCaretValueRule && rule.caretPath === 'date') {
+          if (isEqual(rule, DEFAULT_DATE)) {
             rulesToRemove.push(i);
           }
         });
         pullAt(sd.rules, rulesToRemove);
+      });
+      this.valueSets.forEach(vs => {
+        const rulesToRemove: number[] = [];
+        vs.rules.forEach((rule, i) => {
+          if (isEqual(rule, DEFAULT_DATE)) {
+            rulesToRemove.push(i);
+          }
+        });
+        pullAt(vs.rules, rulesToRemove);
+      });
+      this.codeSystems.forEach(cs => {
+        const rulesToRemove: number[] = [];
+        cs.rules.forEach((rule, i) => {
+          if (isEqual(rule, DEFAULT_DATE)) {
+            rulesToRemove.push(i);
+          }
+        });
+        pullAt(cs.rules, rulesToRemove);
       });
     }
   }
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -70,6 +70,7 @@ export class Package {
     this.removeImpliedZeroToZeroCardRules();
     this.suppressUrlAssignmentOnExtensions();
     this.removeDefaultSlicingRules();
+    this.removeDateRules();
   }
 
   private resolveProfileParents(processor: FHIRProcessor): void {
@@ -329,5 +330,39 @@ export class Package {
       );
       pullAt(sd.rules, rulesToRemove);
     });
+  }
+
+  private removeDateRules(): void {
+    let allDatesMatch = true;
+    let date: string;
+    [...this.profiles, ...this.extensions].forEach(sd => {
+      for (const rule of sd.rules) {
+        if (!allDatesMatch) break; // If any date hasn't matched the others, stop looking at other rules
+        if (rule instanceof ExportableCaretValueRule && rule.caretPath === 'date') {
+          const dateValue = rule.value as string; // If the rule assigns a date, the value will be a string.
+          if (!date) date = dateValue; // Set the date to match
+          if (date !== rule.value) {
+            allDatesMatch = false;
+          }
+        }
+      }
+    });
+
+    // If all dates are the same, are defined to the second, and are in GMT,
+    // we can assume they were set by the IG Publisher and can be safely removed.
+    const dateTimeRegex = /^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|1[0-2])(-(0[1-9]|[1-2][0-9]|3[0-1])(T([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00)))?)?)?$/;
+    const dateTimeMatch = date?.match(dateTimeRegex);
+    const usesGMT = dateTimeMatch?.[12] === '+00:00'; // dateTimeMatch[12] will hold timezone info if present
+    if (allDatesMatch && usesGMT) {
+      [...this.profiles, ...this.extensions].forEach(sd => {
+        const rulesToRemove: number[] = [];
+        sd.rules.forEach((rule, i) => {
+          if (rule instanceof ExportableCaretValueRule && rule.caretPath === 'date') {
+            rulesToRemove.push(i);
+          }
+        });
+        pullAt(sd.rules, rulesToRemove);
+      });
+    }
   }
 }
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -11,7 +11,8 @@ import {
   ExportableContainsRule,
   ExportableOnlyRule,
   ExportableCaretValueRule,
-  ExportableFixedValueRule
+  ExportableFixedValueRule,
+  ExportableRule
 } from '../exportable';
 import { FHIRProcessor } from './FHIRProcessor';
 import { logger } from '../utils';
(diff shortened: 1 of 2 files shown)


src/processor/Package.ts
```diff
@@ -10,7 +10,8 @@ import {
   ExportableCombinedCardFlagRule,
   ExportableContainsRule,
   ExportableOnlyRule,
-  ExportableCaretValueRule
+  ExportableCaretValueRule,
+  ExportableRule
```
ngfreiter, line 14: ⚠️ ⚠️

ngfreiter: Since it isn't clear here, I think the lint has a problem because `ExportableRule` is defined but not used.

jafeltra: :panic:

jafeltra: I didn't realize our lint check on GitHub wouldn't fail if there was a warning. I guess that's probably fair, but this is what I get for forgetting to run it myself. It should be fixed in this very non-suspicious commit 4a3a012.

src/processor/Package.ts
```diff
@@ -330,4 +331,63 @@ export class Package {
       pullAt(sd.rules, rulesToRemove);
     });
   }
+
+  private removeDateRules(): void {
+    let allDatesMatch = true;
+    let date: string;
+    [...this.profiles, ...this.extensions, ...this.valueSets, ...this.codeSystems].forEach(sd => {
```
ngfreiter, line 338: Maybe instead of `sd` we could call this `resource`? Since `valueSets` and `codeSystems` are not `sd`.

src/processor/Package.ts
```diff
@@ -330,4 +331,63 @@ export class Package {
...
+      });
+      this.codeSystems.forEach(cs => {
+        const rulesToRemove: number[] = [];
+        cs.rules.forEach((rule, i) => {
+          if (isEqual(rule, DEFAULT_DATE)) {
+            rulesToRemove.push(i);
+          }
+        });
+        pullAt(cs.rules, rulesToRemove);
+      });
```
ngfreiter, line 390: Are these blocks separated so that the variable can be named correctly (`sd`, `vs`, `cs`)? If that is the main reason, I'm wondering if we could just combine these into one block that loops over all artifacts and names the iterator something like `resource`?

src/processor/Package.ts
```diff
@@ -330,4 +331,63 @@ export class Package {
...
+      }
+    });
+
+    // If all dates are the same, are defined to the second, and are in GMT,
+    // we can assume they were set by the IG Publisher and can be safely removed.
+    const dateTimeRegex = /^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|1[0-2])(-(0[1-9]|[1-2][0-9]|3[0-1])(T([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00)))?)?)?$/;
+    const dateTimeMatch = date?.match(dateTimeRegex);
+    const usesGMT = dateTimeMatch?.[12] === '+00:00'; // dateTimeMatch[12] will hold timezone info if present
+    if (allDatesMatch && usesGMT) {
+      const DEFAULT_DATE = new ExportableCaretValueRule('');
```
ngfreiter, line 361: It freaks me out a little to all-caps this, because that makes me thing is is a global variable. But that's more personal preference so if you like the caps, keep the caps.

src/processor/Package.ts
```diff
@@ -330,4 +331,63 @@ export class Package {
...
+        }
+      }
+    });
+
+    // If all dates are the same, are defined to the second, and are in GMT,
+    // we can assume they were set by the IG Publisher and can be safely removed.
+    const dateTimeRegex = /^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|1[0-2])(-(0[1-9]|[1-2][0-9]|3[0-1])(T([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00)))?)?)?$/;
+    const dateTimeMatch = date?.match(dateTimeRegex);
+    const usesGMT = dateTimeMatch?.[12] === '+00:00'; // dateTimeMatch[12] will hold timezone info if present
+    if (allDatesMatch && usesGMT) {
```
ngfreiter, line 360: This is not a huge deal, but I'm wondering if maybe it would make sense to just simplify the regex so that the regex only matches dates that have seconds and hours and uses GMT. That way if we match the regex, we can remove rules, if not we are done. However that would require understanding that bear of a regex, so if you don't want to do that I'm fine with that.

ngfreiter (commented):
I have a few comments for your consideration.

src/processor/Package.ts
```diff
@@ -330,4 +331,63 @@ export class Package {
...
+        if (!allDatesMatch) break; // If any date hasn't matched the others, stop looking at other rules
+        if (
+          rule instanceof ExportableCaretValueRule &&
+          rule.caretPath === 'date' &&
+          rule.path === ''
+        ) {
+          const dateValue = rule.value as string; // If the rule assigns a date, the value will be a string.
+          if (!date) date = dateValue; // Set the date to match
+          if (date !== dateValue) {
+            allDatesMatch = false;
```
cmoesel, line 349: I was originally going to ask "why not just `break` here (instead of at line 340)?"

But then I realized that really, if you hit an non-matching date, we know right then that we don't want this optimization at all.  So rather than `break` here, you could just `return` out of the whole function here.  That will require converting the `forEach` on line 338 to a more traditional for loop though (otherwise you just return out of that single iteration).

src/processor/Package.ts
```diff
@@ -330,4 +331,63 @@ export class Package {
...
+          }
+        }
+      }
+    });
+
+    // If all dates are the same, are defined to the second, and are in GMT,
+    // we can assume they were set by the IG Publisher and can be safely removed.
+    const dateTimeRegex = /^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|1[0-2])(-(0[1-9]|[1-2][0-9]|3[0-1])(T([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00)))?)?)?$/;
```
cmoesel, line 357: Please tell me you found this regex somewhere.  If you did, can you put the link in the comment?  If you didn't, then God bless you for writing this regex.

src/processor/Package.ts
```diff
@@ -330,4 +331,63 @@ export class Package {
...
+    const dateTimeMatch = date?.match(dateTimeRegex);
```
cmoesel, line 358: We could avoid a few of the `?` operators (and a few executions) if we just put:
```ts
if (date == null) {
  return;
}
```
somewhere around line 355.

src/processor/Package.ts
```diff
@@ -330,4 +331,63 @@ export class Package {
...
+      DEFAULT_DATE.caretPath = 'date';
+      DEFAULT_DATE.value = date;
+      [...this.profiles, ...this.extensions].forEach(sd => {
+        const rulesToRemove: number[] = [];
+        sd.rules.forEach((rule, i) => {
+          if (isEqual(rule, DEFAULT_DATE)) {
+            rulesToRemove.push(i);
+          }
+        });
+        pullAt(sd.rules, rulesToRemove);
```
cmoesel, line 371: I know this is a pattern we use in a lot of the optimizers -- and I think there is a reason for it in the others (?) -- but here, wouldn't it be valid just to use the much simpler:
```ts
sd.rules = sd.rules.filter(rule => !isEqual(rule, DEFAULT_DATE));
```

cmoesel: And same for the two applications below.

test/processor/Package.test.ts
```diff
@@ -1092,5 +1092,84 @@ describe('Package', () => {
...
+      myPackage.add(valueSet);
+      myPackage.add(codeSystem);
+      myPackage.optimize(processor);
+      expect(profile.rules).toHaveLength(0); // date CaretValueRule removed
+      expect(extension.rules).toHaveLength(0); // date CaretValueRule removed
+      expect(valueSet.rules).toHaveLength(0); // date CaretValueRule removed
+      expect(codeSystem.rules).toHaveLength(0); // date CaretValueRule removed
+    });
+
+    it('should not remove date caret rules if date appears to be set by IG publisher (different dates)', () => {
```
cmoesel, line 1121: If there are different dates, then doesn't that mean the dates (or at least some of them) _don't_ appear to be set by IG Publisher?

test/processor/Package.test.ts
```diff
@@ -1092,5 +1092,84 @@ describe('Package', () => {
...
+      extension.rules = [dateCaretRule2];
+      const myPackage = new Package();
+      myPackage.add(profile);
+      myPackage.add(extension);
+      myPackage.optimize(processor);
+      expect(profile.rules).toHaveLength(1); // date CaretValueRule is not removed
+      expect(extension.rules).toHaveLength(1); // date CaretValueRule is not removed
+    });
+
+    it('should not remove date caret rules if date appears to be set by IG publisher (no time)', () => {
```
cmoesel, line 1141: If there is no time, then doesn't that mean the dates _don't_ appear to be set by IG Publisher?

test/processor/Package.test.ts
```diff
@@ -1092,5 +1092,84 @@ describe('Package', () => {
...
+      extension.rules = [dateCaretRule];
+      const myPackage = new Package();
+      myPackage.add(profile);
+      myPackage.add(extension);
+      myPackage.optimize(processor);
+      expect(profile.rules).toHaveLength(1); // date CaretValueRule is not removed
+      expect(extension.rules).toHaveLength(1); // date CaretValueRule is not removed
+    });
+
+    it('should not remove date caret rules if date appears to be set by IG publisher (different time zone)', () => {
```
cmoesel, line 1158: Again, if it's not GMT, then doesn't that mean the dates _don't_ appear to be set by IG Publisher?

cmoesel (changes requested):
I've made a few suggestions to consider in individual comments.

src/processor/Package.ts
cmoesel: I think of CAPS as being a true constant -- i.e., a value _defined_ at compile time (as opposed to a `constant` that is _assigned_ some value at runtime).

cmoesel: Oh.  But based on my definition, I guess that also means it should not be CAPS since we don't know `DEFAULT.value` until runtime. ;-)

src/processor/Package.ts
jafeltra: Oh yikes there's just no way in the entire world I could write this. It's the dateTime regular expression from FHIR. I'll add a link.

test/processor/Package.test.ts
jafeltra: This is all what I get for copying and pasting.

src/processor/Package.ts
jafeltra: Oh dear I'm sorry to freak you out. I can definitely change the caps, but I also now realize I did something similar in `removeDefaultSlicingRules`, so I'm not sure if it makes you more or less comfortable to have both of them caps or just one of the caps. Or neither of them caps.

src/processor/Package.ts
jafeltra: There's something that TypeScript is complaining about, which I think is due to the fact that the ValueSet and CodeSystem `rules` are not subsets of the `ExportableSdRule` type. I had separated them out because I wasn't really sure why it was upset, but I probably should take another look. So I'll look again in the morning!

src/processor/Package.ts
jafeltra: I simplified the check for looking if it uses GMT, but since this is the regular expression for `dateTime`s in FHIR, I think it might be safest to keep the full regex and make sure the value matches that. This gives another confirmation that the value is valid FHIR and likely set by the publisher, since the publisher wouldn't set an invalid `dateTime`.

jafeltra:
@cmoesel I believe the latest commit (dab4c3f) should address all your comments.

It also addresses the first of @ngfreiter's comments, but was going to confirm how he's feeling about the caps given the other variable that is caps and not global. I also still need to figure out the typescript errors I get when I combine the three `forEach`s when filtering out the rules, so that comment hasn't been addressed yet either.

src/processor/Package.ts
cmoesel: Oh yeah.  Well crap.  I forgot about path.  In that case, I've probably also violated my own definition of all CAPS constants.  Nick, why did you have to say anything?!  Now look what you've done!

src/processor/Package.ts
```diff
@@ -330,4 +331,57 @@ export class Package {
...
+      return;
+    }
+
+    // If all dates are the same, are defined to the second, and are in GMT,
+    // we can assume they were set by the IG Publisher and can be safely removed.
+    // Make sure the value matches one allowed by the dateTime type in FHIR.
+    // See: http://hl7.org/fhir/R4/datatypes.html#dateTime
+    const dateTimeRegex = /^([0-9]([0-9]([0-9][1-9]|[1-9]0)|[1-9]00)|[1-9]000)(-(0[1-9]|1[0-2])(-(0[1-9]|[1-2][0-9]|3[0-1])(T([01][0-9]|2[0-3]):[0-5][0-9]:([0-5][0-9]|60)(\.[0-9]+)?(Z|(\+|-)((0[0-9]|1[0-3]):[0-5][0-9]|14:00)))?)?)?$/;
+    const dateTimeMatch = date.match(dateTimeRegex);
+    const usesGMT = dateTimeMatch ? date.endsWith('+00:00') : false; // If it is a valid FHIR dateTime, check that it uses GMT time zone
```
cmoesel, line 371: I think this can be said a little more simply as:
```ts
const usesGMT = dateTimeMatch && date.endsWith('+00:00');
```

cmoesel (approved):
I made one small suggestion you can take or leave.

I'm personally fine w/ the ALL CAPS since you're just following a pattern we already established in previous optimizers.  But if Nick wants to fight it, that's OK too.

I think that the three separate `forEach` loops aren't such a big deal now that you've changed the logic within them to be so much more succinct.  I wouldn't bang my head against the typescript wall for too long trying to find a solution.

src/processor/Package.ts
ngfreiter: All caps are fine, I'm not too picky.

src/processor/Package.ts
ngfreiter: I think you can get away with this:
```
 [...this.profiles, ...this.extensions, ...this.valueSets, ...this.codeSystems].forEach(sd => {
        const rulesToRemove: number[] = [];
        sd.rules.forEach((rule: ExportableRule, i: number) => {
          if (isEqual(rule, DEFAULT_DATE)) {
            rulesToRemove.push(i);
          }
        });
        pullAt(sd.rules as ExportableRule[], rulesToRemove);
      });
```
since all rules inherit from `ExportableRule`.

src/processor/Package.ts
jafeltra: I will take it!

jafeltra:
@ngfreiter I just pushed the change to combine those three loops into one. I left the caps as they are for now, but maybe it's worth changing both places in the future. I think this covers all the outstanding comments on here, including the last one from Chris, but let me know if I missed anything or if you find anything new!

ngfreiter (approved):
Looks good!

PR #17, opened 2020-10-12
Write ObeysRule

mint-thompson:
Completes task [CIMPL-515](https://standardhealthrecord.atlassian.net/browse/CIMPL-515).

Instances of `ExportableObeysRule` are found in the `rules` of an `ExportableProfile` or `ExportableExtension` instance. So, not a whole lot of work is needed to make sure they will be exported when found. The only slight nuance is that a path on the root element is given as no path, instead of the `.` path used in other rules.


cmoesel (dismissed):
Yep.  Looks good!

ngfreiter (approved):
Looks good!

test/exportable/ExportableExtension.test.ts
```diff
@@ -98,16 +99,22 @@ describe('ExportableExtension', () => {
     input.rules.push(flagRule);
 
     const bindingRule = new ExportableBindingRule('value[x]');
-    bindingRule.valueSet = 'http://example.org/ValueSet/Foo';
+    bindingRule.
+      = 'http://example.org/ValueSet/Foo';
```
jafeltra, line 103:
```suggestion
    bindingRule.valueSet = 'http://example.org/ValueSet/Foo';
```

cmoesel: Thank you for fixing that nasty bad good for nothing merge commit.

cmoesel (approved):
Re-approving!

PR #18, opened 2020-10-13
Write Invariant

mint-thompson:
Completes task [CIMPL-518](https://standardhealthrecord.atlassian.net/browse/CIMPL-518).

An Invariant contains no rules, only metadata. However, it uses keywords that are different from most other FSH entities.


jafeltra (approved):
Looks good to me!

PR #19, opened 2020-10-14
Write Mapping and MappingRule

jafeltra:
This PR adds support for writing `Mapping`s and mapping rules ([CIMPL-517](https://standardhealthrecord.atlassian.net/browse/CIMPL-517))

I followed a similar pattern to invariants with keywords that are specific to Mappings. Since the processing side of things have not been implemented yet, I think this is safe to be merge into master.

### DIFF
--- a/src/exportable/ExportableMappingRule.ts
+++ b/src/exportable/ExportableMappingRule.ts
@@ -1,4 +1,5 @@
 import { fshrules } from 'fsh-sushi';
+import { fshifyString } from '../exportable/common';
 import { ExportableRule } from '.';
 
 export class ExportableMappingRule extends fshrules.MappingRule implements ExportableRule {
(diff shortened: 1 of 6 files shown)


src/exportable/ExportableMappingRule.ts
```diff
@@ -0,0 +1,15 @@
...
+export class ExportableMappingRule extends fshrules.MappingRule implements ExportableRule {
+  constructor(path: string) {
+    super(path);
+  }
+
+  toFSH(): string {
+    const path = this.path ? `${this.path} ` : '';
+    const comment = this.comment ? ` "${this.comment}"` : '';
+    const language = this.language ? ` ${this.language.toString()}` : '';
+    return `* ${path}-> "${this.map}"${comment}${language}`;
```
mint-thompson, line 13: Now might be the time when we need to properly decide how to use `fshifyString`, since currently it is sometimes used on the early side (AssignmentRuleExtractor, getFSHValue by way of CaretValueRuleExtractor) and sometimes on the late side (metadataToFSH). What you've written is correct if strings are always fshified as early as possible. Please let me know what you think about this.

mint-thompson (commented):
This looks good! But, it does raise a question about our use of `fshifyString`.

src/exportable/ExportableMappingRule.ts
jafeltra: Hmm that's a good point that I hadn't thought about. I don't think I have a super strong opinion beyond being consistent. If I had to choose, I guess ensuring that `fshifyString` gets called in the MappingRuleExtractor makes sense and follows the two other rule extractors you pointed out. That does seem to differ a bit from how metadataToFSH handles it, so if we wanted to go for consistency, we'd have to change those. Do you have a preference for which way we lean?

mint-thompson (approved):

PR #20, opened 2020-10-16
Extract configuration from non-IG resources

mint-thompson:
Completes task [CIMPL-541](https://standardhealthrecord.atlassian.net/browse/CIMPL-541).

The included set of configuration parameters is based on the task description. The way that `status` is pulled out feels a little shakier than the rest, since it's technically a code. The way that `id` and `name` are determined is based on the pattern I observed for MCODE.

One feature of this that probably merits some discussion is how to behave when the discovered values don't follow the pattern. Is it better to create the config with possibly nonsensical values, or to instead say "none of the discovered values appear more than once, so you don't get anything"?


src/extractor/ConfigurationExtractor.ts
```diff
@@ -0,0 +1,113 @@
...
+          potentialStrings.push(caretRule.value.toString());
+        }
+      }
+    });
+    return ConfigurationExtractor.getStringMode(potentialStrings);
+  }
+
+  private static getStringMode(potentials: string[]): string {
+    if (potentials.length > 0) {
+      return maxBy(toPairs(countBy(potentials)), pair => pair[1])[0];
```
cmoesel, line 108: This made my head hurt.  It's right.  But it made my head hurt.  I don't know if the following is necessary or if it would survive `prettier`'s weird code-formatting algorithms, but this helped me understand it:
```ts
      return maxBy(
        toPairs(
          countBy(
            potentials        // e.g., ['foo', 'bar', 'baz', 'foo', 'bar', 'bar']
          )                   // e.g., { 'foo': 2, 'bar': 3, 'baz': 1 }
        ),                    // e.g., [ ['foo', 2], ['bar': 3], ['baz': 1] ]
        pair => pair[1]       // e.g., ['foo', 2] => 2; ['bar': 3] => 3; ['baz': 1] => 1
      )[0];                   // e.g., ['bar', 3][0] -----> 'bar'
```

test/extractor/ConfigurationExtractor.test.ts
```diff
@@ -0,0 +1,128 @@
...
+    );
+
+    resources.add(typicalProfile);
+    resources.add(typicalExtension);
+    resources.add(typicalValueSet);
+    resources.add(typicalCodeSystem);
+  });
+
+  describe('#process', () => {
+    it('should return a configuration with a canonical and fhirVersion when there are no resources', () => {
```
cmoesel, line 67: Wait.  But how?!

cmoesel: OK.  I debugged it.  The `canonical` and `fhirVersion` are both set to `''`.  This doesn't really seem reasonable to me...  I mean, it's an edge case for sure, but what's the point?  I'd rather that we either don't return a `Configuration` at all or we just hard-code some values in this case (some example `canonical` and `4.0.1` for the `fhirVersion`).

test/extractor/ConfigurationExtractor.test.ts
```diff
@@ -0,0 +1,128 @@
...
+      noUrlResources.valueSets[0].rules[0] = cloneDeep(differentRule);
+      noUrlResources.codeSystems[0].rules[0] = cloneDeep(differentRule);
+      const result = ConfigurationExtractor.inferCanonical(noUrlResources);
+      expect(result).toBe('');
+    });
+
+    it('should use the common url base to determine the canonical when all resources have the same url base', () => {
+      const result = ConfigurationExtractor.inferCanonical(resources);
+      expect(result).toBe('http://example.org/our/ig');
+    });
```
cmoesel, line 113: I think it might be nice to have another test or two that demonstrates what happens w/ URLs w/ more or less path parts, for example, `http://example.org` and `http://example.org/this/is/our/ig`.  Hard to say for sure, but I think I'd expect `example` and `example.this.is.our.ig` respectively.

src/extractor/ConfigurationExtractor.ts
```diff
@@ -0,0 +1,113 @@
...
+    resources.codeSystems.forEach(codeSystem => {
+      const urlRule = codeSystem.rules.find(rule => {
+        return (
+          rule instanceof ExportableCaretValueRule && rule.path === '' && rule.caretPath === 'url'
+        );
+      }) as ExportableCaretValueRule;
+      if (urlRule) {
+        potentialCanonicals.push(urlRule.value.toString().replace(/\/CodeSystem\/[^/]+$/, ''));
+      }
+    });
```
cmoesel, line 74: The three blocks of code above are all exactly the same except for the variable they iterate over and the regular expression that they use.  I think you could just combine them all:
```ts
    [...resources.profiles, ...resources.extensions, ...resources.valueSets, ...resources.codeSystems].forEach(resource => {
      const urlRule = resource.rules.find(rule => {
        return (
          rule instanceof ExportableCaretValueRule && rule.path === '' && rule.caretPath === 'url'
        );
      }) as ExportableCaretValueRule;
      if (urlRule) {
        potentialCanonicals.push(
          urlRule.value.toString().replace(/\/(StructureDefinition|ValueSet|CodeSystem)\/[^/]+$/, '')
        );
      }
    });
```

src/extractor/ConfigurationExtractor.ts
```diff
@@ -0,0 +1,113 @@
+import { countBy, toPairs, maxBy, capitalize } from 'lodash';
+import { ImplementationGuideStatus } from 'fsh-sushi/dist/fhirtypes/ImplementationGuide';
+import { Package } from '../processor';
+import { ExportableConfiguration, ExportableCaretValueRule, ExportableRule } from '../exportable';
+import { fshtypes } from 'fsh-sushi';
+
+export class ConfigurationExtractor {
+  static canonicalRegex = /^https?:\/\/(\w+)\.[\w.]+\/(.*)$/;
```
cmoesel, line 8: I'm not sure we're agreed on how we name things like this, but should it be:
```ts
  static CANONICAL_REGEX = /^https?:\/\/(\w+)\.[\w.]+\/(.*)$/;
```
?

cmoesel: Actually, you only use it in one place.  I think it would be cleaner just to inline it on line 19 than to use a constant.  That way when people read the code they don't need to bounce from line 19 to line 8 just to see what's going on.

cmoesel: And... (sorry - my sort of stream-of-conscious review is made clear by my multiple comments)... A few things about the regex itself...

This would match `http://example.org/` but not `http://example.org`.  I think it should also match on the latter.

Because it only captures the first part of the domain, something like `http://www.example.org/our/ig` would result in `www.our.ig` -- but I would expect either `www.example.our.ig` or `example.our.ig` (if we wanted to put in special-case logic to ignore `www`).

cmoesel: And to answer a question I _think_ you asked -- I believe that the general pattern you saw in mCODE will hold for any official HL7 IGs, but I'm not sure it will for non-HL7 IGs, which is why I want to be more flexible w/ the regex.

cmoesel (changes requested):
This is really neat and employs some clever approaches!  I had a few questions and suggestions.

Also, I think we had talked about potentially operating on the formal definitions _before_ processing them (so the extractors would have the benefit of the config values).  In our discussion I think we said that would be ideal if it's similar effort, but also allowed for doing it _after_ if that was too complex to do it before.  Based on this PR, I'm assuming that you determined there was not enough ROI for operating on the formal definitions prior to processing?

src/extractor/ConfigurationExtractor.ts
```diff
@@ -0,0 +1,113 @@
+import { countBy, toPairs, maxBy, capitalize } from 'lodash';
+import { ImplementationGuideStatus } from 'fsh-sushi/dist/fhirtypes/ImplementationGuide';
```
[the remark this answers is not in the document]
mint-thompson: I can't do it quite like that, since `ImplementationGuideStatus` is a type, but I can clean up the import to not dig through `dist`.

src/extractor/ConfigurationExtractor.ts
mint-thompson: prettier makes it end up looking like this:
```typescript
    if (potentials.length > 0) {
      return maxBy(
        toPairs(
          countBy(
            // this is a little dense. here's an example:
            potentials // ['a', 'b', 'b', 'a', 'c', 'b']
          ) // {'a': 2, 'b': 3, 'c': 1}
        ), // [['a', 2], ['b', 3], ['c', 1]]
        pair => pair[1] // ['a', 2] => 2, ['b', 3] => 3, ['c', 1] => 1
      )[0]; // ['b', 3][0] ---> 'b'
    }
```
So if that seems helpful, I'll include it.

src/extractor/ConfigurationExtractor.ts
mint-thompson: Thanks for pointing out the issues with the regex. I had focused too much on the case of multipart top-level domains like `https://example.co.uk`, hence why was just taking the first part of the domain. Doing this well might require a list of TLDs to check against, but either way, I definitely need a lot more test cases for this function to make sure I'm giving something that is frequently correct and is always at least kind of close.

src/extractor/ConfigurationExtractor.ts
cmoesel: I don't know.  Not having control over the spacing makes it a little less helpful.  Let's get a 3rd opinion.  @ngfreiter , @jafeltra , @julianxcarter -- do you think it's better just to leave this line as-is (`return maxBy(toPairs(countBy(potentials)), pair => pair[1])[0];`) or to go with the annotated post-prettier format @mint-thompson posted above?

src/extractor/ConfigurationExtractor.ts
cmoesel: I hadn't thought about `co.uk` and domains like that... Good point. That said, I think we need to recognize that we can't get it right 100% of the time -- and that it's not all that difficult for someone to fix it.  So don't go crazy trying to cover every edge case.  We probably just need a "good enough".

src/extractor/ConfigurationExtractor.ts
ngfreiter: Can we leave it one line with the annotation above it? I think the annotations are nice, but post-prettier they make things more confusing than just no annotations at all

ngfreiter: Or even maybe just a comment indicating what this line of code does at a general level, if you don't want to include the full step by step process.

cmoesel: @ngfreiter -- do you mean:
```ts
      // return maxBy(
      //   toPairs(
      //     countBy(
      //       potentials    // e.g., ['foo', 'bar', 'baz', 'foo', 'bar', 'bar']
      //     )               // e.g., { 'foo': 2, 'bar': 3, 'baz': 1 }
      //   ),                // e.g., [ ['foo', 2], ['bar': 3], ['baz': 1] ]
      //   pair => pair[1]   // e.g., ['foo', 2] => 2; ['bar': 3] => 3; ['baz': 1] => 1
      // )[0];               // e.g., ['bar', 3][0] -----> 'bar'
      return maxBy(toPairs(countBy(potentials)), pair => pair[1])[0];
```

cmoesel: Or:
```ts
      // Example results after each function
      //   potentials: ['foo', 'bar', 'baz', 'foo', 'bar', 'bar']
      //   countBy:    { 'foo': 2, 'bar': 3, 'baz': 1 }
      //   toPairs:    [ ['foo', 2], ['bar': 3], ['baz': 1] ]
      //   maxBy:      ['bar', 3]
      //   maxyBy[0]:  'bar'
      return maxBy(toPairs(countBy(potentials)), pair => pair[1])[0];
```

ngfreiter: Yep that's pretty much what I meant. But I meant I'd also be fine with something like:
```
      // Find the most frequently occurring potential
      return maxBy(toPairs(countBy(potentials)), pair => pair[1])[0];
```

ngfreiter: Oh when I posted this comment I hadn't seen your "Or". Your "Or" is exactly what I meant.

cmoesel: I feel like `// Find the most frequently occurring potential` doesn't add much since the function is called `getStringMode` -- which implies finding the most frequent value already.  @mint-thompson - how do you feel about my suggestion in the "Or:" above?

mint-thompson:
@cmoesel I just gave some more thought to getting the configuration before processing resources, and I came up with an idea that I hadn't before. Since the `FHIRProcessor` is keeping all of the raw content for StructureDefinitions around anyway, it should be fine to do something like this:
- Iterate over all files, adding the raw content to the correct list on the `FHIRProcessor`
- If there's raw content for an ImplementationGuide, process it to make the config.
- If there isn't an ImplementationGuide available, use the other resources to extract the config.
- With our config, process all other resources.

For some reason, I kept imagining having to iterate and read the set of files twice, but that's not the case at all. I'll update this implementation accordingly.

src/extractor/ConfigurationExtractor.ts
mint-thompson: I like it! I'll add something in that format.

ngfreiter (changes requested):
This isn't working for me trying to roundtrip on US Core, because we don't set `FSHOnly`, so SUSHI tries to generate an IG, but in that case `copyrightYear` and `releaseLabel` are required by SUSHI. So should we just set `FSHOnly` on extracted configurations? That would make sense to me, since an IG is not coming into GoFSH, I don't think we need one out of SUSHI.

cmoesel:
I think that when we extract it from the IG JSON we also might not be setting `FSHOnly`.  We should check...

cmoesel:
... because unless GoFSH is going to reproduce the whole input IG as it was, I think it's better to always export FSHOnly so it's very clear we're _not_ importing _most_ of the IG-config stuff.

src/processor/FHIRProcessor.ts
```diff
@@ -7,52 +7,72 @@ import {
...
+      logger.debug(`Registered contents of ${inputPath} as CodeSystem.`);
     } else if (rawContent['resourceType'] === 'ValueSet') {
-      logger.debug(`Processing contents of ${inputPath} as ValueSet.`);
-      return [ValueSetProcessor.process(rawContent)];
+      this.valueSets.push(rawContent);
+      logger.debug(`Registered contents of ${inputPath} as ValueSet.`);
     } else if (rawContent['resourceType'] === 'ImplementationGuide') {
-      return [ConfigurationProcessor.process(rawContent)];
+      this.implementationGuides.push(rawContent);
+    }
```
cmoesel, line 41: Should we have an `else` that logs a warning -- "Skipping currently unsupported resource: /path/to/foo"?

src/processor/FHIRProcessor.ts
```diff
@@ -7,52 +7,72 @@ import {
...
+  }
+
+  process(resources: Package) {
+    let config: ExportableConfiguration;
+    if (this.implementationGuides.length > 0) {
+      config = ConfigurationProcessor.process(this.implementationGuides[0]);
```
cmoesel, line 47: Holy cow.  You can use the same exact code to extract the data from an IG as from the other conformance resources.  Amazing.

src/processor/FHIRProcessor.ts
```diff
@@ -7,52 +7,72 @@ import {
...
+    resources.add(config);
+    this.structureDefinitions.forEach(sd => {
+      try {
+        StructureDefinitionProcessor.process(sd, this.fhir, resources.invariants).forEach(
+          resource => {
+            resources.add(resource);
+          }
+        );
+      } catch (ex) {
+        logger.error(`Could not process StructureDefinition: ${ex.message}`);
```
cmoesel, line 60: Will `ex.message` contain the info needed to identify _which_ SD could not be processed?  I think it might not?  In which case we'd want to add some type of identifier to this error message.

src/processor/FHIRProcessor.ts
```diff
@@ -7,52 +7,72 @@ import {
...
+      }
+    });
+    this.codeSystems.forEach(cs => {
+      try {
+        resources.add(CodeSystemProcessor.process(cs));
+      } catch (ex) {
+        logger.error(`Could not process CodeSystem: ${ex.message}`);
```
cmoesel, line 67: Same as above -- we probably need to add an identifier so it's clear _which_ CS failed.

src/processor/FHIRProcessor.ts
```diff
@@ -7,52 +7,72 @@ import {
...
+      }
+    });
+    this.valueSets.forEach(vs => {
+      try {
+        resources.add(ValueSetProcessor.process(vs));
+      } catch (ex) {
+        logger.error(`Could not process ValueSet: ${ex.message}`);
```
cmoesel, line 74: Same as above -- we probably need to add an identifier so it's clear _which_ VS failed.

src/utils/Processing.ts
```diff
@@ -26,13 +26,12 @@ export function getResources(inDir: string, defs: fhirdefs.FHIRDefinitions) {
   logger.info(`Found ${files.length} JSON files.`);
   files.forEach(file => {
     try {
-      processor.process(file, resources.invariants).forEach(resource => {
-        resources.add(resource);
-      });
+      processor.register(file);
     } catch (ex) {
-      logger.error(`Could not process ${file}: ${ex.message}`);
+      logger.error(`Could not register ${file}: ${ex.message}`);
```
cmoesel, line 31: I'm not sure if the word "register" will mean much to our users.  Maybe "Could not _load_ foo.json"?

test/processor/FHIRProcessor.test.ts
```diff
@@ -15,47 +16,53 @@ describe('FHIRProcessor', () => {
...
+    processor = new FHIRProcessor(new fhirdefs.FHIRDefinitions());
     structureDefinitionSpy.mockClear();
     codeSystemSpy.mockClear();
     valueSetSpy.mockClear();
   });
 
   it('should try to process a Profile with the StructureDefinitionProcessor', () => {
-    processor.process(path.join(__dirname, 'fixtures', 'simple-profile.json'));
+    processor.register(path.join(__dirname, 'fixtures', 'simple-profile.json'));
+    processor.process(new Package());
```
cmoesel, line 33: I don't remember the big picture, but are there use cases where we definitely want to pass in the `Package` to be mutated (which is what I assume it is doing)?  Versus having `process` be a no-arg function and returning a `Package` it constructs itself?

cmoesel: Also, if we expect many more tests in this file, we may want to consider creating a helper function (e.g., `registerAndProcess`) so we don't need to make two very similar function calls in every test.

cmoesel (changes requested):
This is great.  I really dig it, @mint-thompson.  I still have a few small questions and/or suggestions.

ngfreiter (commented):
It looks like the `FSHOnly` flag is getting carried through to the `ExportableConfiguration`, but that doesn't get exported by `toFSH`, since that only exports specific fields, and `FSHOnly` is not one of those fields.

cmoesel (approved):
This is looking very good and works well.  I am approving it as-is but also leaving one potential area for improvement, which you can take or leave.

When I run this against US Core (after deleting its IG files), I get this config:
```yaml
canonical: http://hl7.org/fhir/us/core
fhirVersion: 4.0.1
FSHOnly: true
id: hl7.fhir.us.core
name: Hl7FhirUsCore
status: active
version: 3.1.0
```

Note the `name` is `Hl7FhirUsCore`.  The real name (in the real IG file) is actually `USCore`.  It strikes me that we can probably _always_ leave off the `Hl7Fhir` prefix in the name (if it exists) -- because IGs would never make that part of their real name (e.g., mCODE is `mCODE`, _not_ `HL7FHIRmCODE`).  If we do this, then US Core would end up being `UsCore`, which still isn't perfect, but is definitely _closer_.

ngfreiter (approved):
Looks good!

src/extractor/ConfigurationExtractor.ts
```diff
@@ -59,7 +59,15 @@ export class ConfigurationExtractor {
     const canonicalMatch = canonical.match(/^https?:\/\/([\w.]+)(\/.*)?$/);
     if (canonicalMatch) {
       const hostParts = canonicalMatch[1].split('.').slice(0, -1);
-      const pathParts = canonicalMatch[2]?.split('/') ?? [];
+      const pathParts = canonicalMatch[2]?.split('/').filter(part => part.length > 0) ?? [];
+      // Many IGs are hosted at a URL that starts with http://hl7.org/fhir/
+      // However, they generally do not include "hl7" and "fhir" at the start of their id and name.
```
cmoesel, line 64: I'm afraid you've taken this a bit too far!  While they don't include "hl7" or "fhir" in their `name`, they _do_ include them in their `id` (and, in fact, I think the IG Publisher requires them in the `id` if they are in the `canonical`).

So this change really should only apply to the `name`.

test/extractor/ConfigurationExtractor.test.ts
```diff
@@ -229,6 +229,14 @@ describe('ConfigurationExtractor', () => {
         id: 'example.orange.apple.banana'
       });
     });
+
+    it('should not include hl7 and fhir in the name and id when the canonical starts with http://hl7.org/fhir/', () => {
+      const canonical = 'http://hl7.org/fhir/us/core';
+      expect(ConfigurationExtractor.inferNameAndId(canonical)).toEqual({
+        name: 'UsCore',
+        id: 'us.core'
```
cmoesel, line 237: As noted in my prior comment, the `id` should still be `hl7.fhir.us.core`.

cmoesel (changes requested):
Your last commit works, but it works a little _too well_!  We still need `hl7.fhir.` to be in the `id` -- we just don't want it in the `name`.

ngfreiter (approved):
Looks good to me.

cmoesel (approved):
Thanks, @mint-thompson. Cool feature!

PR #21, opened 2020-10-19
Combine ContainsRules

ngfreiter:
This fixes CIMPL-536 by combining multiple contains rules into a single rule, if they share the same path. So for example:
```
* extension contains ombCategory 0..5 MS
* extension contains detailed 0..*
* extension contains text 1..1 MS
```
can be combined into:
```
* extension contains ombCategory 0..5 MS
    and detailed 0..* 
    and text 1..1 MS
```


test/processor/Package.test.ts
```diff
@@ -1178,5 +1178,79 @@ describe('Package', () => {
...
+      myPackage.optimize(processor);
+
+      const combinedContainsRule = new ExportableContainsRule('category');
+      combinedContainsRule.items.push({ name: 'foo' }, { name: 'bar' }, { name: 'baz' });
+      combinedContainsRule.cardRules.push(cardRule1, cardRule2, cardRule3);
+      combinedContainsRule.flagRules.push(flagRule3);
+      expect(profile.rules).toEqual([combinedContainsRule]);
+    });
+
+    it('should not combine multiple contains rules which have the differnt paths', () => {
```
cmoesel, line 1220:
```suggestion
    it('should not combine multiple contains rules which have the different paths', () => {
```

cmoesel (commented):
This looks good (minus a simple typo in the tests).  I ran this on US Core and found that it exports multiples like this:
```
* extension contains http://hl7.org/fhir/us/core/StructureDefinition/us-core-race named race 0..1 MS and
    http://hl7.org/fhir/us/core/StructureDefinition/us-core-ethnicity named ethnicity 0..1 MS and
    http://hl7.org/fhir/us/core/StructureDefinition/us-core-birthsex named birthsex 0..1 MS
```

I think the best practice style we want to encourage is:
```
* extension contains
    http://hl7.org/fhir/us/core/StructureDefinition/us-core-race named race 0..1 MS and
    http://hl7.org/fhir/us/core/StructureDefinition/us-core-ethnicity named ethnicity 0..1 MS and
    http://hl7.org/fhir/us/core/StructureDefinition/us-core-birthsex named birthsex 0..1 MS
```
This is how mCODE does it.

So... Ideally, if there is one contains, it puts it on the same line as the contains.  Otherwise it puts each contains on its own indented line.

I know that's outside the scope of this PR, but since this PR exposes it for the first time, I wonder if maybe we can/should fix it here?

ngfreiter:
Alright @cmoesel, improved the styling and made the other change in d52a25f.

cmoesel (approved):
Thanks, @ngfreiter!  Looks good and works well.  Tested on US Core and confirmed -- it is an improvement!

src/processor/Package.ts
```diff
@@ -388,4 +389,27 @@ export class Package {
...
+
+  private combineContainsRules(): void {
+    [...this.profiles, ...this.extensions].forEach(sd => {
+      const rulesToRemove: number[] = [];
+      sd.rules.forEach((rule, i) => {
+        if (rule instanceof ExportableContainsRule && !rulesToRemove.includes(i)) {
+          sd.rules.forEach((otherRule, otherRuleIdx) => {
+            if (
+              otherRule.path === rule.path &&
+              otherRule instanceof ExportableContainsRule &&
```
jafeltra, line 401: I'm not entirely sure if this matters, but I feel like I remember talking once about swapping the order of these checks, so that it checks the `instanceof` before looking at the path. I don't exactly remember the context of that conversation, so if you remember and don't think it's relevant, it's all good with me.

jafeltra (commented):
This looks good to me! I had one super minor comment below.

src/processor/Package.ts
ngfreiter: Hm yeah I do remember that conversation. I looked throughout the file, and there is variability in the order of us checking `path` and doing an `instanceof` check. I guess personally I must like doing `path` first because that is what I subconsciously did here, but I think @cmoesel had been a fan of doing `instanceof` (first)[https://github.com/FHIR/GoFSH/pull/10/files/890554e1a512a91249980dfdeec14f585ef691bb#diff-bf3b12f81fc71976ec2cec060d7dc5c9ecb4a186adbfd9bbb4d858c4e38dbd98]. So if @cmoesel will let this one slide, I'm ok with it.

cmoesel: Nick wins: https://jsben.ch/IpyO7

jafeltra: I ran that four times and you each won twice 🤷‍♀️

ngfreiter: Participation medals for all 🏅 !

cmoesel: I ran it twice and I lost both times. :-(

But I am using Firefox.  Is @jafeltra using Chrome?  If so, maybe the Chrome JS implementation isn't as definitive.

jafeltra: I was using Chrome. So maybe!

PR #22, opened 2020-10-19
Process Profiles and Extensions with StructureDefinitionProcessor

mint-thompson:
Completes task [CIMPL-560](https://standardhealthrecord.atlassian.net/browse/CIMPL-560).

Profiles and Extensions are similar enough that they can be processed by the same class. The ProfileProcessor and ExtensionProcessor classes are removed. Test cases are reorganized to more directly test functions of StructureDefinitionProcessor.


test/processor/StructureDefinitionProcessor.test.ts
```diff
@@ -0,0 +1,257 @@
...
+        fs.readFileSync(path.join(__dirname, 'fixtures', 'big-profile.json'), 'utf-8')
+      );
+      const [profile] = StructureDefinitionProcessor.process(input, defs);
+
+      expect(profile).toBeInstanceOf(ExportableProfile);
+      expect(profile.name).toBe('BigProfile');
+      expect(profile.parent).toBe('https://demo.org/StructureDefinition/SmallProfile');
+    });
+
+    it('should convert a profile with an invariant', () => {
```
jafeltra, line 62: Is there a reason you took off the comparison of the `invariants` list to the `ExportableInvariant` that was in the old ProfileProcessor tests?

jafeltra (commented):
This looks good to me! I think the reorganization of the tests make sense. I just had one question that I can across while looking at the tests.

cmoesel (approved):
This cleans things up nice!  I ran it against US Core and confirmed I get the same exact results as master!

test/processor/StructureDefinitionProcessor.test.ts
mint-thompson: The correctness of the invariant is part of testing `InvariantExtractor`. Here, we can assume that the extractor will produce correct results, and we just want to test that we got an invariant.

jafeltra: Ah! That makes sense! Thanks for the explanation!

jafeltra (approved):
Looks good to me!

PR #24, opened 2020-10-21
Read/Write support for ValueSet metadata

cmoesel:
This PR combines two tasks:
* CIMPL-475: FHIR to FSH ValueSet with simple metadata
* CIMPL-476: Write ValueSet with simple metadata

This also makes a slight tweak to `CodeSystemProcessor` to support reading in CodeSystems that don't have a `name` but do have an `id`.  Unlike `StructureDefinition`, which has `id 1..1`, `ValueSet` and `CodeSystem` have `id 0..1`, so technically `name` is not _required_.  If we get a VS or CS without a `name`, but it has an `id`, we'll try to create a reasonable `name` from the `id`.  We _must_ have a `name` because FHIR Shorthand _requires_ it.  _NOTE: This would not survive a round-trip since it introduces a `name` where there wasn't one before -- but I think that producing a `CodeSystem`/`ValueSet` w/ an extra name is preferred to producing none at all._


ngfreiter (dismissed):
All looks good to me!

jafeltra (approved):
This looks lovely.

PR #25, opened 2020-10-22
Detect and replace FHIR core resource urls

mint-thompson:
Completes task [CIMPL-488](https://standardhealthrecord.atlassian.net/browse/CIMPL-488).

When a FHIR core resource url is a profile parent or the type in an OnlyRule, it is not necessary to use the full url. The name of the resource is sufficient.

### DIFF
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -83,16 +83,16 @@ export class Package {
   }
 
   private resolveProfileParents(processor: FHIRProcessor): void {
-    for (const profile of this.profiles) {
-      if (profile.parent) {
+    for (const resource of [...this.profiles, ...this.extensions]) {
+      if (resource.parent) {
         // The parent might be another SD in the processor or a core FHIR resource
-        const parentSd = processor.structureDefinitions.find(sd => sd.url === profile.parent);
+        const parentSd = processor.structureDefinitions.find(sd => sd.url === resource.parent);
         if (parentSd?.name) {
-          profile.parent = parentSd.name;
+          resource.parent = parentSd.name;
         } else {
-          const fhirMatch = profile.parent.match(FHIR_BASE_URL);
+          const fhirMatch = resource.parent.match(FHIR_BASE_URL);
           if (fhirMatch?.[1]) {
-            profile.parent = fhirMatch[1];
+            resource.parent = fhirMatch[1];
           }
         }
       }
--- a/src/processor/Package.ts
+++ b/src/processor/Package.ts
@@ -21,6 +21,8 @@ import { pullAt, groupBy, values, flatten, isEqual } from 'lodash';
 import { fshtypes } from 'fsh-sushi';
 const { FshCode } = fshtypes;
 
+const FHIR_BASE_URL = /http:\/\/hl7\.org\/fhir\/StructureDefinition\/(.+)/;
+
 export class Package {
   public readonly profiles: ExportableProfile[] = [];
   public readonly extensions: ExportableExtension[] = [];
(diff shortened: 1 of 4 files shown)


src/processor/Package.ts
```diff
@@ -78,14 +79,21 @@ export class Package {
     this.removeDefaultSlicingRules();
     this.removeDateRules();
     this.combineContainsRules();
+    this.simplifyFHIROnlyRules();
   }
 
   private resolveProfileParents(processor: FHIRProcessor): void {
     for (const profile of this.profiles) {
```
cmoesel, line 86: I know this is pre-existing code, but technically and extension *can* have a parent -- so we really ought to be doing this for `this.extensions` too.

mint-thompson: Does the existing logic for finding parents in `processor.structureDefinitions` also apply to Extensions? If that's the case, then the whole function can apply to both Profiles and Extensions.

cmoesel: Yes, I think you can just change this line to be something like:
```ts
for (const def of [...this.profiles, ...this.extensions]) {
```

cmoesel (commented):
This works well.  I have two questions for your consideration:

1. IIRC, in SUSHI, we had specially handle cases where an IG defined a profile w/ _the same name_ as a FHIR resource (e.g., `Patient`).  In this case, I think SUSHI would resolve `Patient` to the profile defined in the IG rather than the FHIR resource (and you needed to use the FHIR resource URL if you wished to refer to the FHIR resource).  As this PR is written here, GoFSH would always turn the FHIR URL into the resource name, even if there exists a profile w/ the same name in the IG -- and therefore would fail a roundtrip.  This is admittedly a rare thing (I hope) so, do we (a) fix it now or (b) log a bug and fix it later?

2. When I ran this against US Core, I noticed that FHIR URLs are converted to their names, but URLs for local profiles are not.  I guess we missed this.  We should probably create a story for that unless you think it fits into the spirit of this PR.  Totally up to you.

jafeltra:
I reviewed this and everything looks good to me! I didn't find anything new, so I'm good to review once Chris's questions are resolved!

src/processor/Package.ts
```diff
@@ -424,14 +424,23 @@ export class Package {
...
-              onlyRuleType.type = fhirMatch[1];
+            // The type might be another SD in the processor or a core FHIR resource
+            const typeSd = processor.structureDefinitions.find(sd => sd.url === onlyRuleType.type);
+            if (typeSd?.name) {
+              onlyRuleType.type = typeSd.name;
+            } else {
+              const fhirMatch = onlyRuleType.type.match(FHIR_BASE_URL);
+              if (fhirMatch?.[1]) {
+                // Only change the FHIR url into a name if it won't collide with a local SD
+                if (!processor.structureDefinitions.some(sd => sd.name === fhirMatch[1])) {
```
cmoesel, line 440: I'll sing [the same song you sang](https://github.com/FHIR/sushi/pull/645#discussion_r508817143) to @julianxcarter (and I missed in my own review of his).  The above two `if` statements could be combined as a single `if` statement.  See, I learned from you, @mint-thompson!  ;-)

cmoesel (approved):
Thanks for sneaking in that extra functionality, @mint-thompson!  I left you one comment just to tease you a little.  It's actually up to you if you want to change it or not.  It's not egregious.

src/processor/Package.ts
mint-thompson: hoist by my own petard!!! updating now

cmoesel (approved):
I just ran this against US Core and I must say: it is beautiful.

PR #26, opened 2020-10-22
Extract Mappings and MappingRules

jafeltra:
This PR supports extracting Mappings with MappingRules from StructureDefinitions ([CIMPL-514](https://standardhealthrecord.atlassian.net/browse/CIMPL-514)).

To extract mappings, I look through the `mapping` array on the StructureDefinition to create the `Mapping`. Then I look through each element's `mapping` and apply `MappingRules` to the appropriate `Mapping`. Any `Mapping`s that are unchanged from the parent are not included. However, if there are new rules on the inherited `Mapping`, the `Mapping` will be exported with only the changed rules.


test/extractor/MappingExtractor.test.ts
```diff
@@ -0,0 +1,123 @@
...
+  it('should not extract mappings that do not exist on the top level structure definition', () => {
+    const element = elements[3]; // Observation.status
+    const mappings = MappingExtractor.process(looseSD, elements, defs);
+    const unexpectedMapping = new ExportableMapping('RogueMapping');
+    unexpectedMapping.source = 'MyObservation';
+    const mappingRule = new ExportableMappingRule('status');
+    mappingRule.map = 'Observation.notReal';
+    unexpectedMapping.rules.push(mappingRule);
+
+    expect(mappings).toHaveLength(4); // Only the three mappings in the SD mapping list are created
```
mint-thompson, line 90: Four mappings are created, but the comment says "three mappings".

src/extractor/MappingExtractor.ts
```diff
@@ -0,0 +1,101 @@
...
+    return [...newMappings, ...changedInheritedMappings];
+  }
+
+  static extractRules(element: ProcessableElementDefinition, mappings: ExportableMapping[]) {
+    element.mapping?.forEach((mapping, i) => {
+      // Mappings are created at SD, so should always find a match at this point
+      const matchingMapping = mappings.find(m => m.name === mapping.identity);
+      if (!matchingMapping) return;
+      const mappingRule = this.processMappingRule(element, mapping, i);
+      matchingMapping.rules.push(mappingRule);
```
mint-thompson, line 74: Minor style request: instead of an early exit with a one-line if, this could be written as:
```suggestion
      if (matchingMapping) {
        const mappingRule = this.processMappingRule(element, mapping, i);
        matchingMapping.rules.push(mappingRule);
      }
```
I feel like this is a bit clearer at expressing "if there is a matching mapping, we make a `MappingRule` for it."

mint-thompson (changes requested):

src/extractor/MappingExtractor.ts
cmoesel, line 74: Instead of the early return, I'd suggest something just like this:
```ts
if (matchingMapping) {
  const mappingRule = this.processMappingRule(element, mapping, i);
  matchingMapping.rules.push(mappingRule);
}
```
or even
```ts
if (matchingMapping) {
  matchingMapping.rules.push(this.processMappingRule(element, mapping, i));
}
```
or I guess this would actually get rid of the early return _and_ the `if` block:
```ts
matchingMapping?.rules.push(this.processMappingRule(element, mapping, i));
```

cmoesel: I left all three of those blocks in my comment because that's how my mind works, BTW.  Just want to show I didn't immediately jump to that last suggestion -- it took some processing to get there!

src/processor/StructureDefinitionProcessor.ts
```diff
@@ -42,8 +46,9 @@ export class StructureDefinitionProcessor {
         elements,
         existingInvariants
       );
+      const mappings = StructureDefinitionProcessor.extractMappings(elements, input, fhir);
       StructureDefinitionProcessor.extractRules(input, elements, sd, fhir);
-      return [sd, ...invariants];
+      return [sd, ...[...invariants, ...mappings]];
```
cmoesel, line 51:
```suggestion
      return [sd, ...invariants, ...mappings];
```
I think that should work, shouldn't it?

test/extractor/CaretValueRuleExtractor.test.ts
```diff
@@ -110,36 +110,18 @@ describe('CaretValueRuleExtractor', () => {
 
     it('should not extract caret rules when an array is exactly the same as the inherited array', () => {
       const sd = cloneDeep(looseSD);
-      sd.mapping = [
+      sd.context = [
```
cmoesel, line 113: Oops.  Sorry I didn't have the foresight to realize these tests would break as soon as we added mapping support!

cmoesel (changes requested):
Thank you, @jafeltra!  I think maybe I was feeling in a refactor-y mood when I reviewed this.  I apologize for all the comments!

I tested this against the US Core in my fhir cache and it actually caused an error.  Here's the command I ran if you want to reproduce this:
```
$ ts-node src/app.ts ~/.fhir/packages/hl7.fhir.us.core\#3.1.0/package/ -o build-uscore-pr-26
```

And here is the error I recieved:

> **error** Could not process /Users/cmoesel/.fhir/packages/hl7.fhir.us.core#3.1.0/package/StructureDefinition-us-core-organization.json: Cannot read property 'mapping' of undefined

src/extractor/MappingExtractor.ts
```diff
@@ -0,0 +1,101 @@
...
+  ): ExportableMapping[] {
+    const mappings =
+      input.mapping?.map(m => {
+        const mapping = new ExportableMapping(m.identity);
+        mapping.source = input.name;
+        if (m.name) mapping.title = m.name;
+        if (m.uri) mapping.target = m.uri;
+        if (m.comment) mapping.description = m.comment;
+        return mapping;
+      }) ?? [];
```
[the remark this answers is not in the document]
jafeltra: I would never have thought to do that, but I like it!

src/export/FSHExporter.ts
```diff
@@ -47,6 +47,14 @@ export class FSHExporter {
...
       }.`
     );
+    for (const mapping of this.fshPackage.mappings) {
+      results.push(mapping.toFSH());
+    }
+    logger.info(
+      `Exported ${this.fshPackage.mappings.length} Mapping${
+        this.fshPackage.mappings.length == 1 ? '' : 's'
+      }.`
+    );
```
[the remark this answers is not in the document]
jafeltra: Oh oops was that not part of this task? If not, oops and bonus!

test/extractor/MappingExtractor.test.ts
jafeltra: Well oops. I should probably read my own comments when I update things.

src/extractor/MappingExtractor.ts
```diff
@@ -0,0 +1,101 @@
...
+    const inheritedMappings = mappings.filter(mapping =>
+      parent?.mapping.some((i: fhirtypes.StructureDefinitionMapping) => i.identity === mapping.name)
+    );
+
+    // For each inherited mapping, look to see if there are any new rules
+    const changedInheritedMappings = inheritedMappings.filter(mapping => {
+      const changedIndexes: number[] = [];
+      let isNew = false;
+      mapping.rules.forEach((rule, i) => {
+        const element = fhirtypes.StructureDefinition.fromJSON(parent).findElementByPath(
```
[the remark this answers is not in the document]
jafeltra: Hm I think at one point I just returned all the `Mapping`s if there was no parent and didn't bother doing any of this, but I guess I took that out. I'll look into that or an `if (parent != null)` block.

src/extractor/MappingExtractor.ts
jafeltra: Ah I like the thought process! And I like the final conclusion too!

src/processor/StructureDefinitionProcessor.ts
jafeltra: I've been struggling over this line. If I leave the type that we return for this function as it is, I don't think your suggested change works, although I really don't understand why. I believe that the type we are declaring is designed so that we know the first element should be either an `ExportableProfile` or `ExportableExtension`. If that is the case and we do want to ensure that the first element is a profile or extension, I'm not sure I know what the right way to have the rest of the elements be either `ExportableInvariants` or `ExportableMappings`. If we aren't intending to have the type of the first element be specifically a Profile or Extension, then I think we should update the return type of the function, which would let me use the simpler suggestion you have here.

But again, I'm getting fairly confused by TypeScript here, so I'm definitely willing to believe I'm missing something.

src/extractor/MappingExtractor.ts
```diff
@@ -0,0 +1,101 @@
...
+        if (!elementMappingRules.find(r => isEqual(r, rule))) {
+          isNew = true;
+        } else {
+          // Remove inherited rules so we don't include them if the mapping is exported
+          changedIndexes.push(i);
+        }
+      });
+      pullAt(mapping.rules, changedIndexes);
+      return isNew;
+    });
```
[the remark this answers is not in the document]
jafeltra: Hm this is an intriguing refactor idea. When I was iterating through the rules of the existing mappings, I was trying to limit how many additional elements I was looking at and creating rules for. But maybe in the end I didn't actually save myself that many iterations, so maybe this would simplify the code enough to make it worth it. I'll take a look (it possibly (probably) will extend to Monday morning though).

jafeltra:
Thanks for the review @cmoesel and @mint-thompson! The latest commit (c7783b3) addresses both of @mint-thompson's comments and most of @cmoesel's. The two pieces it doesn't address is the `[sd, ...invariants, ...mappings]` vs [sd, ...[...invariants, ...mappings]]` and corresponding types since I haven't quite figured out the best way forward and the overarching refactor that Chris suggested. I'm going to look more into the refactor for a bit, but that will probably extend to Monday. I may also need to talk to someone about the TypeScript question on Monday.

src/processor/StructureDefinitionProcessor.ts
cmoesel: OK.  If my suggestion doesn't work as-is then I am also confused by typescript, and in that case I don't really have any suggestions.  If what you did works, then it is fine.  I just thought it was weird -- but I guess it is _necessarily_ weird.

jafeltra: a4901fe adds a TODO to use this much clearer version of this line when we use TypeScript 4.0

jafeltra:
Alrighty, I think this PR is ready for re-review. On Friday I had left open the question about the TypeScript mystery return type, which is now resolved in a4901fe, and the refactor that Chris mentioned, which is implemented in 16868b1. All the other comments should be resolved in c7783b3 or were taken out as part of the refactor. I haven't resolved the comments because I thought they would be helpful as you re-reviewed, but if it is easier to have them marked as resolved, I can do that!

src/extractor/MappingExtractor.ts
```diff
@@ -10,68 +14,75 @@ export class MappingExtractor {
...
-    const newMappings = mappings.filter(mapping => newItems.some(i => i.identity === mapping.name));
-    const inheritedMappings = mappings.filter(mapping =>
-      parent?.mapping.some((i: fhirtypes.StructureDefinitionMapping) => i.identity === mapping.name)
-    );
 
-    // For each inherited mapping, look to see if there are any new rules
-    const changedInheritedMappings = inheritedMappings.filter(mapping => {
+    // If there is no parent found, all the mappings should be exported.
+    if (parent == null || !StructureDefinitionProcessor.isProcessableStructureDefinition(parent))
+      return mappings;
```
cmoesel, line 32: We don't actually have a linter rule for this, but I feel like we always use `{` `}` around the body of `if` statements (except, possibly, for one-liners).  Is that just me?  @mint-thompson?

src/extractor/MappingExtractor.ts
```diff
@@ -10,68 +14,75 @@ export class MappingExtractor {
...
+    const parentSDElements =
+      parent.snapshot?.element?.map(rawElement => {
+        return ProcessableElementDefinition.fromJSON(rawElement, false);
+      }) ?? [];
+
+    const parentMappings = this.extractMappings(parent, parentSDElements);
+
+    // Filter out any mappings that come directly from the parent SD.
+    // Only return mappings that are new to the profile or inherited mappings that include new MappingRules
+    const newMappings = mappings.filter(mapping => {
```
cmoesel, line 43: This feels a little weird to me because inside the filter, you are _mutating_ the input `mappings`.  This isn't something we usually expect from a filter operation...  It's valid, but... I don't know... it just feels a little surprising to me.

You _could_ refactor so it has:
```ts
const unmodifiedMappingIndexes: number[] = [];
```
then iterate `mappings.forEach` and push to the `unmodifiedMappingIndexes` any mappings that are 100% inherited from the parent.  Finally `pullAt` those indexes from `mappings`.  You are still mutating `mappings`, but since it is happening in `forEach` and since there is no additional `newMappings` array, it seems fine to mutate in this context.

Or ignore me if I'm just hung up on mutating in a filter function and I need to get over it.

cmoesel (commented):
Thanks, @jafeltra!  This is looking good!  I left a few comments for your consideration -- let me know what you think.  I don't know if they are fair comments or not. ;-)

BTW -- this also seems to have fixed whatever issue I had run into w/ US Core before, so that's a bonus!

src/extractor/MappingExtractor.ts
jafeltra: Ah I'm not sure if we have a rule, but that's something I totally support. This was a new line until prettier changed it on me and I didn't notice that it didn't add the `{ } `.

jafeltra:
Thanks @cmoesel. The US Core issue was because something was trying to get accessed from an `undefined` value. I had made that change in c7783b3 where I grouped a bunch of changes all together!

src/extractor/MappingExtractor.ts
jafeltra: Fixed in e0b28d5!

src/extractor/MappingExtractor.ts
jafeltra: I tried to refactor this in a branch called [extract-mapping-filter-refactor](https://github.com/FHIR/GoFSH/tree/extract-mapping-filter-refactor). The last commit is the only one different from this branch. However, I'm a little confused by how you were suggesting to refactor this, so I'm not sure if that commit is getting at the spirit of what you were suggesting. If you already have your suggestion written out, feel free to commit it to that branch. Or I'm happy to chat on the phone if it is easier at this point.

cmoesel: That wasn't exactly what I had in mind, but when I implemented what I had in mind, I didn't like it (bad idea "Yesterday Chris").  But instead, I came up with this -- yet another approach (on branch [extract-mapping-filter-refactor-cmm](https://github.com/FHIR/GoFSH/tree/extract-mapping-filter-refactor-cmm)): https://github.com/FHIR/GoFSH/commit/021330fdd01059d08138591c80fa53e3e6fba7c5

cmoesel: BTW -- my commit above is not the most efficient (it loops through rules even when there is no parent) -- but I'm not too concerned about efficiency and it looked much prettier that way than my more efficient attempt. ;-)

jafeltra: Thanks for looking at it again. I guess I'm still struggling to understand why this approach is better than the current approach since the original `mappings` array is modified in both cases. But I guess it's just the difference between mutating `mappings` within a `forEach` vs within a `filter`. Is that right? If so, I'm comfortable using your implementation on this branch.

cmoesel: Yeah.  It's basically that you don't expect a _filter_ operation to have side-effects.  It should have one job and one job only -- to narrow a set down to a (potentially) smaller set.  A _forEach_ however, is much more general in nature and using a _forEach_ to mutate objects is a common use case -- so no one should be "surprised" by it.

jafeltra: Okay! Then I shall push your implementation. Thanks for looking at it and coming up with a solution!

cmoesel: I feel kind of like a GitHub bully pushing my solution on you.  I'm sorry!

jafeltra: Here it is! 23a679e

And you are not a GitHub bully!! You have helped me learn something and you simplified something that I've stared at for too long and couldn't simplify!

cmoesel (approved):
Thanks, @jafeltra!  Looks good!

jafeltra:
@cmoesel and @mint-thompson I just resolved conflicts with master, so now I think this is set again!

cmoesel (approved):

mint-thompson (approved):

PR #29, opened 2020-10-30
0.3.0

ngfreiter:
GoFSH 0.3.0 release. I know @jafeltra perhaps has some incoming changes to how `Mapping`s work. My understanding is those should not change the user experience at all, so I think we can do a release. If we want to wait though that's fine too, just hold off on merging this.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "gofsh",
-  "version": "0.2.0",
+  "version": "0.3.0",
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
Looks good to me.  I gave it a test run against US Core and then ran SUSHI 1.0.0-beta.3 and SUSHI 0.16.0 against it.  One error on each, but it's an expected one due to a bug in US Core.

PR #30, opened 2020-10-30
Extract caret rules on ValueSets and CodeSystems

ngfreiter:
This adds `^` rules to ValueSets and CodeSystems. Anything that will not be handled by a keyword or a rule gets put into a `^` rule, since unlike StructureDefinitions there is no parent to compare to.

This also updates the `.gitignore` file to include `gofsh/` since that is the default build directory.

This addresses [CIMPL-581](https://standardhealthrecord.atlassian.net/browse/CIMPL-581) and [CIMPL-582](https://standardhealthrecord.atlassian.net/browse/CIMPL-582).


src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -112,38 +112,65 @@ export class CaretValueRuleExtractor {
...
+    'resourceType',
+    'id',
+    'url',
+    'name',
+    'title',
+    'description',
+    'compose.include',
+    'compose.exclude'
+  ],
+  CodeSystem: ['resourceType', 'id', 'url', 'name', 'title', 'description', 'concept'],
```
mint-thompson, line 155: Are we still planning for CodeSystems that have extra elements on a `concept` that aren't handled by other rules to be exported as FSH Instances? If so, this is fine, but if not, there might need to be some additional logic for getting those extra elements captured as CaretValueRules. And, similar question, but for ValueSet.

cmoesel: @mint-thompson -- yes, that is the plan.  If we detect any unsupported properties on `concept` then we should generate an `Instance` representation instead.  I have a few stories assigned to me that start to get this underway, but I don't think I'll get to them this sprint.

mint-thompson (dismissed):

mint-thompson (dismissed):

src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -112,38 +112,65 @@ export class CaretValueRuleExtractor {
...
+  StructureDefinition: [
+    'resourceType',
+    'id',
+    'url', // NOTE: For now, we assume URL matches canonical and is generated
```
cmoesel, line 159: Now that we have a config in all cases, we probably shouldn't assume this anymore.  But that's beyond the scope of this PR.

src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -112,38 +112,65 @@ export class CaretValueRuleExtractor {
...
+  static processResource(
+    input: any,
+    fisher: utils.Fishable,
+    resourceType: 'ValueSet' | 'CodeSystem'
+  ): ExportableCaretValueRule[] {
+    const caretValueRules: ExportableCaretValueRule[] = [];
+    const flatVS = getPathValuePairs(input);
+    Object.keys(flatVS)
+      .filter(
+        key => !RESOURCE_IGNORED_PROPERTIES[resourceType].some(property => key.startsWith(property))
```
cmoesel, line 132: I don't think that `key.startsWith(property)` is strict enough.  For example, wouldn't the key `identifier` match on the property `id` (and then get unintentionally ignored)?

I think we ought to be checking ``key === property || key.startsWith(`${property}.`)``, right?

test/extractor/CaretValueRuleExtractor.test.ts
```diff
@@ -406,6 +416,58 @@ describe('CaretValueRuleExtractor', () => {
...
+        looseVS.resourceType
+      );
+      expect(caretRules).toEqual<ExportableCaretValueRule[]>([]);
+    });
+
+    it('should extract ValueSet caret rules when non keyword-based properties have changed', () => {
+      const vs = cloneDeep(looseVS);
+      vs.status = 'active';
+      vs.compose.inactive = true;
+      const caretRules = CaretValueRuleExtractor.processResource(vs, defs, vs.resourceType);
```
cmoesel, line 433: To ensure that _only_ the expected values are received, we should assert that the length is 2:
```ts
expect(caretRules).toHaveLength(2);
```

Or we should just do a full compare on the array instead of the separate assertions on lines 440 and 441:
```ts
expect(caretRules).toEqual([expectedRule1, expectedRule2]);
```

cmoesel: BTW -- I modified this test to also set `identifier` to a value and confirmed that with the current code, it does _not_ export a caret rule for `identifier` -- so we do need to fix that comparison in `processResource` and probably go ahead and use `identifier` in the tests as well.

cmoesel (changes requested):
I noted one bug that will need to be addressed.  Otherwise looks good!

P.S.  When running this on US Core, I noticed lots of caret rules for `^text.*` where `^text.status = #generated` or `^text.status = #extensions`.  I'm wondering if maybe we should have an optimizer that strips those out.  What do you think?

test/extractor/CaretValueRuleExtractor.test.ts
ngfreiter: Fixed in https://github.com/FHIR/GoFSH/pull/30/commits/27fd64dd2479cd55d7eb17f02657e9d1e00f40fd

src/extractor/CaretValueRuleExtractor.ts
ngfreiter: Yeah good catch. It was actually a little more complex, since in the second case, we would also want to ignore keys like `compose.include[0].system`, but I added a regex that I think should do this correctly.

ngfreiter:
> I noted one bug that will need to be addressed. Otherwise looks good!
> 
> P.S. When running this on US Core, I noticed lots of caret rules for `^text.*` where `^text.status = #generated` or `^text.status = #extensions`. I'm wondering if maybe we should have an optimizer that strips those out. What do you think?

Yeah I agree, and I think I should just do that on this PR, since the resulting FSH is pretty unreadable in parts because of those huge `text.div` strings. So let's not merge this until I do that. As I'm understanding it, the optimization would be to strip out `^text.div` and `^text.status` rules, if the `^text.status` is `#generated` or `#extension`. Would this optimization apply to any resource, or just VS/CS?

src/extractor/CaretValueRuleExtractor.ts
cmoesel: Oooh, yeah.  Good thinking.

cmoesel:
> Would this optimization apply to any resource, or just VS/CS?

I think it applies to any resource.

ngfreiter:
Added the discussed optimizer in https://github.com/FHIR/GoFSH/pull/30/commits/7f24d47d5a9183ea7bf2769e03b35a0d9a36f939, should be ready for re-review @cmoesel and @mint-thompson.

src/optimizer/plugins/RemoveGeneratedTextRulesOptimizer.ts
```diff
@@ -0,0 +1,36 @@
...
+          rule.caretPath === 'text.status' &&
+          rule.value instanceof fshtypes.FshCode &&
+          ['generated', 'extensions'].includes(rule.value.code)
+      );
+      if (hasGeneratedText) {
+        resource.rules.forEach((rule: ExportableRule, i: number) => {
+          if (
+            rule instanceof ExportableCaretValueRule &&
+            rule.path === '' &&
+            rule.caretPath.match(/text\./)
```
mint-thompson, line 27: Should this match rules that have `text.` anywhere in their `caretPath`, or only those where the path starts with `text.`?

mint-thompson (commented):
I have one small question about path matching for text rules, but everything else looks good.

src/optimizer/plugins/RemoveGeneratedTextRulesOptimizer.ts
cmoesel: Good question, @mint-thompson.  I feel like we probably really want `/^text\./`  I can't think of any cases where this would be embedded in a deeper path -- and if it were, I'm not sure we'd want to clear it anyway.

cmoesel (commented):
Aside from the one thing @mint-thompson mentioned, this all looks good to me.

I'll be out Thursday, but consider this my approval as long as you address Mint's concern.

src/optimizer/plugins/RemoveGeneratedTextRulesOptimizer.ts
ngfreiter: Ah yikes, thanks for finding this. It's especially bad, since there are several elements on StructureDefinitions that would contain "text." in the path. Fixed and tested in https://github.com/FHIR/GoFSH/pull/30/commits/395fd59b072369e2623c603280a6e0c8a41aae32.

mint-thompson (approved):

cmoesel (approved):

PR #31, opened 2020-10-30
Update Mapping optimizer approach

jafeltra:
This PR adds a few additional updates to Mappings based on discussion yesterday:

1. Since the Mapping optimizer wasn't technically an optimization but was fixing incorrect Mappings, I took a different approach. The approach I took (after discussing a few different ones at stand up this morning) was to make all the Mapping `name`s the verbose version with `source` information by default. The optimization step now does true optimization and simplifies the `name` to not include the `source` information if it does not need to.
2. I added a test to check for escaped characters in the `target` since @cmoesel kindly added the `fshifyString` in the FSH for that property in a separate PR.

These were the only two additional Mappings fixes I am aware of, but if there are other that we're aware of, let me know. I also don't think this PR is high priority since these changes don't fix any existing bugs; they're just improvements to stick to our definition of "optimization".


src/extractor/MappingExtractor.ts
```diff
@@ -76,7 +77,7 @@ export class MappingExtractor {
   static extractRules(element: ProcessableElementDefinition, mappings: ExportableMapping[]) {
     element.mapping?.forEach((mapping, i) => {
       // Mappings are created at SD, so should always find a match at this point
-      const matchingMapping = mappings.find(m => m.name === mapping.identity);
+      const matchingMapping = mappings.find(m => m.id === mapping.identity);
```
ngfreiter, line 80: Why do we use the `id` here instead of the `name`? I thought the `name` was the one that would be unique, but the `id` may not be?

ngfreiter (commented):
I ran this with goFSH and SUSHI and it worked exactly the same as on `master`, and the optimization makes sense. I just have one question that I think must just be me getting confused, since this does look like it is working well.

src/extractor/MappingExtractor.ts
cmoesel: I think it is because the `name` is a FSH-only construct.  It does not have any representation in FHIR, so we can't compare it against a value from the FHIR mapping.  So the correct thing to compare the FHIR `mapping.identity` to is the FSH Mapping `id`.

While `name` is unique across the whole FSH Tank, `id` _should_ be unique across the `StructureDefinition`, and that's all that matters in this case since FHIR `mapping`s are scoped within the SD only.

ngfreiter (dismissed):
Chris answered my question so I approve!

src/extractor/MappingExtractor.ts
jafeltra: What Chris said!

mint-thompson (approved):

jafeltra:
Oops I rebased to resolve the conflicts and dismissed @mint-thompson and @cmoesel's reviews.

mint-thompson (approved):

ngfreiter (approved):

PR #32, opened 2020-11-03
LakeOfFHIR, WildFHIR, Fishers, and Ignoring Non-FHIR resources

cmoesel:
Instead of passing around `FHIRDefinitions` objects, we mostly pass around `Fishable` fishers now -- which is a more flexible approach and aligns more closely w/ SUSHI.  This PR also introduces a separately managed collection of the local resources we are processing, called `LakeOfFHIR` (analogous to SUSHI's `FSHTank`), which contains a collection of `WildFHIR` (analogous to SUSHI's `RawFSH`). Like SUSHI's `FSHTank`, `LakeOfFHIR` is `Fishable`.

When I started this branch, I was trying to solve errors I saw when running GoFSH against mCODE output built by the IG Publisher.  For example:
> Cannot reliably export top-level caret rules for IntersectionUsCorePediatricHeadOccipitalFrontalCircumferencePercentileProfileAndTNMClinicalPrimaryTumorCategory because GoFSH cannot find a definition for its parent: undefined. If its parent is from another IG, run GoFSH again declaring that IG as a dependency.

I *thought* there was an issue w/ finding and using the local resources being processed, but it turns out that the file being reported (`IntersectionUsCorePediatricHeadOccipitalFrontalCircumferencePercentileProfileAndTNMClinicalPrimaryTumorCategory`) is an intermediate temporary `StructureDefinition` created by the IG Publisher that we should *not* process at all.  So this PR also fixes this by detecting these types of files and skipping them.  It also more elegantly skips (and reports) other non-FHIR JSON files.

To see the difference in mCODE, first build [mCODE](https://github.com/HL7/fhir-mCODE-ig/) using SUSHI and run the IG Publisher on it.  Then run `GoFSH` against it, for example:
```
ts-node src/app.ts ~/dev/HL7/fhir-mCODE-ig/output/ -d hl7.fhir.us.core@3.1.0
```
You'll see a bunch of those warnings go by...  Now do the same thing with this PR branch and they'll be gone.  If you compare outputs, you'll also notice that the `master` branch generated FSH for all of those bogus intermediate temporary files.


src/app.ts
```diff
@@ -73,7 +73,6 @@ async function app() {
   }
 
   await Promise.all(dependencyDefs);
-  fhirdefs.loadFromPath(inDir, '', defs);
```
cmoesel, line 76: Note that we don't need to load the `inDir` into the `defs` anymore because we now manage if in the `LakeOfFHIR`.

src/processor/LakeOfFHIR.ts
```diff
@@ -0,0 +1,73 @@
+import { fhirdefs, utils } from 'fsh-sushi';
+import { WildFHIR } from './WildFHIR';
+
+// Like FSHTank in SUSHI, but it doesn't contain FSH, it contains FHIR.  And who ever heard of a tank of FHIR?  But a lake of FHIR...
+export class LakeOfFHIR implements utils.Fishable {
+  constructor(public readonly docs: WildFHIR[]) {}
```
ngfreiter, line 6: Why the consistent use of `docs` when describing the contents of `LakeOfFhir`? Would something like either `wilds` make more sense to match `WildFHIR`, or something like `resources` make sense? Unless I'm missing something all these docs should be `resources` right?

ngfreiter (commented):
This looks good. I could only muster up one small question. I also noticed there were no tests for `MasterFisher`, so those might be good to add. In investigating the coverage report to make that discovery, I also noticed this:
image

Which is out of scope of this PR, but if you knew how to fix it off the top of your head and the fix was simple, I thought I'd let you know in case you wanted to fix it. Otherwise let me know and I can add a task to address that.

src/processor/LakeOfFHIR.ts
cmoesel: It was kind of inspired by `FSHTank`, which has `docs` -- but `docs` probably makes more sense there than here.  I think I shied away from `resources` because in other places you would expect that to be the FHIR JSON, but this is the `WildFHIR` wrapper (so we can keep a reference to the path).  What do you think?  Do you have a preference for `resources`, `wilds`, or something else?

ngfreiter: I guess I wasn't thinking about the fact that the resources are wrapped. In that case I think docs is fine, I'm not sure `wilds` is better. That may just be because the word `wilds` sounds weird to me, but although `docs` is somewhat confusing here, it does align well with what we think of as a `doc` in SUSHI, so I'm fine with `docs`.

ngfreiter (approved):
This all is looking good to me. Thanks for the test coverage table fix.

cmoesel:
@ngfreiter -- I just pushed up a unit test for `MasterFisher` -- which, by the way, dismissed your approval.

mint-thompson (approved):

PR #34, opened 2020-11-05
Dependencies are now extracted from IGs

julianxcarter:
This addresses [CIMPL-577](https://standardhealthrecord.atlassian.net/browse/CIMPL-577), exporting dependencies from an IG file and including them in a generated config.yaml (using the abbreviated syntax with just the packageID and version number). One more test case is still needed within the `ExportableConfiguration.test.ts` file, that will be added shortly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,6 +37,7 @@
     "@types/jest": "^25.1.4",
     "@types/lodash": "^4.14.159",
     "@types/node": "^12.12.34",
+    "@types/semver": "^7.3.4",
     "@types/temp": "^0.8.34",
     "@types/toposort": "^2.0.3",
     "@typescript-eslint/eslint-plugin": "^2.26.0",
@@ -52,7 +53,6 @@
     "typescript": "^3.8.3"
   },
   "dependencies": {
-    "@types/semver": "^7.3.4",
     "chalk": "^4.1.0",
     "commander": "^6.0.0",
     "flat": "^5.0.2",
@@ -60,6 +60,7 @@
     "fsh-sushi": "^1.0.0",
     "ini": "^1.3.5",
     "lodash": "^4.17.19",
+    "semver": "^7.3.2",
     "temp": "^0.9.1",
     "toposort": "^2.0.2",
     "winston": "^3.3.3",
--- a/package.json
+++ b/package.json
@@ -52,6 +52,7 @@
     "typescript": "^3.8.3"
   },
   "dependencies": {
+    "@types/semver": "^7.3.4",
     "chalk": "^4.1.0",
     "commander": "^6.0.0",
     "flat": "^5.0.2",
--- a/src/app.ts
+++ b/src/app.ts
@@ -65,7 +65,7 @@ async function app() {
   const defs = new fhirdefs.FHIRDefinitions();
 
   // Trim empty spaces from command line dependencies
-  const dependencies = program.dependencies?.map((dep: string) => dep.trim());
+  const dependencies = program.dependency?.map((dep: string) => dep.trim());
 
   // Load FhirProcessor and config object
   const processor = getFhirProcessor(inDir, defs);
--- a/src/app.ts
+++ b/src/app.ts
@@ -7,13 +7,12 @@ import { fhirdefs, utils } from 'fsh-sushi';
 
 import {
   ensureOutputDir,
-  getIGDependencies,
   getInputDir,
   getResources,
   loadExternalDependencies,
   writeFSH
 } from './utils/Processing';
-import { logger, stats } from './utils';
+import { getConfig, logger, stats } from './utils';
 import { Package } from './processor';
 
 const FSH_VERSION = '0.13.x';
@@ -65,11 +64,27 @@ async function app() {
   const defs = new fhirdefs.FHIRDefinitions();
   const dependencies = program.dependency;
 
-  // Fetch dependencies from included IG(s)
-  const dependencyDefs = loadExternalDependencies(defs, [
-    ...(dependencies ?? []),
-    ...getIGDependencies(inDir)
-  ]);
+  // Trim empty spaces from commad line dependencies
+  if (dependencies && dependencies.length > 0) {
+    dependencies.forEach((dep: string, i: number) => {
+      dependencies[i] = dep.trim();
+    });
+  }
+
+  const config = await getConfig(inDir, defs, dependencies);
+  let allDependencies: string[];
+
+  if (config.config.dependencies && config.config.dependencies.length > 0) {
+    allDependencies = [];
+
+    config.config.dependencies.forEach(dep => {
+      const newDep = `${dep.packageId}@${dep.version}`;
+      allDependencies.push(newDep);
+    });
+  }
+
+  // Load dependencies from config for GoFSH processing
+  const dependencyDefs = loadExternalDependencies(defs, allDependencies);
 
   let outDir: string;
   try {
@@ -84,6 +99,7 @@ async function app() {
   let resources: Package;
   try {
     resources = await getResources(inDir, defs);
+    resources.add(config);
   } catch (err) {
     logger.error(`Could not use input directory: ${err.message}`);
     process.exit(1);
--- a/src/app.ts
+++ b/src/app.ts
@@ -3,11 +3,12 @@
 import path from 'path';
 import fs from 'fs-extra';
 import program from 'commander';
-import { fhirdefs, utils } from 'fsh-sushi';
+import { fhirdefs, fhirtypes, utils } from 'fsh-sushi';
 
 import {
   ensureOutputDir,
   getInputDir,
+  getFhirProcessor,
   getResources,
   loadExternalDependencies,
   writeFSH
@@ -62,26 +63,17 @@ async function app() {
 
   // Load dependencies
   const defs = new fhirdefs.FHIRDefinitions();
-  const dependencies = program.dependency;
+  let dependencies = program.dependency;
 
   // Trim empty spaces from commad line dependencies
-  if (dependencies && dependencies.length > 0) {
-    dependencies.forEach((dep: string, i: number) => {
-      dependencies[i] = dep.trim();
-    });
-  }
+  dependencies = dependencies?.map((dep: string) => dep.trim());
 
-  const config = await getConfig(inDir, defs, dependencies);
-  let allDependencies: string[];
+  const processor = await getFhirProcessor(inDir, defs);
+  const config = await getConfig(processor, dependencies);
 
-  if (config.config.dependencies && config.config.dependencies.length > 0) {
-    allDependencies = [];
-
-    config.config.dependencies.forEach(dep => {
-      const newDep = `${dep.packageId}@${dep.version}`;
-      allDependencies.push(newDep);
-    });
-  }
+  const allDependencies = config.config.dependencies?.map(
+    (dep: fhirtypes.ImplementationGuideDependsOn) => `${dep.packageId}@${dep.version}`
+  );
 
   // Load dependencies from config for GoFSH processing
   const dependencyDefs = loadExternalDependencies(defs, allDependencies);
@@ -98,7 +90,7 @@ async function app() {
 
   let resources: Package;
   try {
-    resources = await getResources(inDir, defs);
+    resources = await getResources(processor);
     resources.add(config);
   } catch (err) {
     logger.error(`Could not use input directory: ${err.message}`);
--- a/src/app.ts
+++ b/src/app.ts
@@ -7,10 +7,10 @@ import { fhirdefs, utils } from 'fsh-sushi';
 
 import {
   ensureOutputDir,
+  getIGDependencies,
   getInputDir,
   getResources,
   loadExternalDependencies,
-  getIGDependencies,
   writeFSH
 } from './utils/Processing';
 import { logger, stats } from './utils';
@@ -64,12 +64,12 @@ async function app() {
   // Load dependencies
   const defs = new fhirdefs.FHIRDefinitions();
   const dependencies = program.dependency;
-  
+
   // Fetch dependencies from included IG
   const igDependencies = getIGDependencies(inDir);
   const allDependencies = dependencies.concat(igDependencies);
   const dependencyDefs = loadExternalDependencies(defs, allDependencies);
-  
+
   let outDir: string;
   try {
     outDir = ensureOutputDir(program.out);
--- a/src/app.ts
+++ b/src/app.ts
@@ -89,8 +89,7 @@ async function app() {
 
   let resources: Package;
   try {
-    resources = await getResources(processor);
-    resources.add(config);
+    resources = await getResources(processor, config);
   } catch (err) {
     logger.error(`Could not use input directory: ${err.message}`);
     process.exit(1);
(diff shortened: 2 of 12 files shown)


src/exportable/ExportableConfiguration.ts
```diff
@@ -1,6 +1,7 @@
 import YAML from 'yaml';
 import { fshtypes } from 'fsh-sushi';
 import { Exportable } from '.';
+import { ImplementationGuideDependsOn } from 'fsh-sushi/dist/fhirtypes';
```
cmoesel, line 4: We've been trying to avoid importing from `fsh-sushi/dist/**` and preferring imports directly from `fsh-sushi`.  So in this case you would do:
```ts
import fhirtypes from 'fsh-sushi';

// ... then to use ImplementationGuideDependsOn later:
this.config.dependencies.forEach((dependency: fhirtypes.ImplementationGuideDependsOn) => {
```

It's a trade-off for sure...  One could argue that one approach is better than the other, but... for now, this is the way we're doing it!

src/exportable/ExportableConfiguration.ts
```diff
@@ -26,6 +27,16 @@ export class ExportableConfiguration implements Exportable {
     if (this.config.version) {
       yaml.contents.add({ key: 'version', value: this.config.version });
     }
+    if (this.config.dependencies) {
+      const fshDependencies: any[] = [];
+      this.config.dependencies.forEach((dependency: ImplementationGuideDependsOn) => {
+        fshDependencies.push({
+          [dependency.packageId]: dependency.version
```
cmoesel, line 34: The config file supports _two_ ways of indicating dependencies.  You implemented this approach:
```yaml
dependencies:
  hl7.fhir.us.core: 3.1.0
```

The other approach is:
```yaml
dependencies:
  hl7.fhir.us.core:
    id: uscore
    uri: http://hl7.org/fhir/us/core/ImplementationGuide/hl7.fhir.us.core
    version: 3.1.0
```

Usually we can derive the `uri` from the `packageId` (which is the `key` in the yaml), so the more advanced approach is usually only needed if the config had explicitly set an `id` for the dependency.

For GoFSH, I think that if `dependency.id` does not exist you can use `[dependency.packageId]: dependency.version` , but if `dependency.id` _does_ exist, you should probably use the more advanced format with `id` and `version` sub-keys (and maybe while you're at it go ahead and throw in the `uri`).

See: https://fshschool.org/docs/sushi/configuration/#recommended-configuration

cmoesel: The other thing is that the value of `dependencies` should not be an array; it should be an object.  Currently you export this:
```yaml
dependencies:
  - hl7.fhir.us.core: 3.1.0
```
but it needs to be this:
```yaml
dependencies:
  hl7.fhir.us.core: 3.1.0
```

cmoesel (changes requested):
Looks like you found all the right places in the code!  Cool!  I found a few things I think need to be updated:

1. `dependencies` is being written as a yaml sequence (array) but it should be an object
2. we should probably support the "advanced" dependency representation too (sorry, this was not documented as a requirement, so there was no way for you to know!)

src/exportable/ExportableConfiguration.ts
ngfreiter, line 34: I'm not sure we want to push the `dependency.packageId` as an array. That seems to lead to output that looks like this:
```
dependencies:
  - hl7.fhir.us.core: 3.1.0
```
Which is not what SUSHI is expecting. SUSHI expected the output to look like this:
```
dependencies:
  hl7.fhir.us.core: 3.1.0
```
In this case `dependencies` is an object, not an array. So instead of making the `fshDependencies` array and pushing onto it, you could make `fshDependencies` an object, and add each `packageId` and `version` as a key and value pair on that object.

ngfreiter: Oh jeez

ngfreiter: @cmoesel just said this! So please ignore this thread.

julianxcarter: Gotcha, I'll go ahead and implement these changes!

ngfreiter (commented):
I think we should use the extracted dependency not only to generate a configuration, but also for GoFSH. In GoFSH, you can specify a dependency using the `--dependency` flag. This tells GoFSH to load the definitions from that dependency, which are required by certain parts of GoFSH. For example, the `CaretValueRuleExtractor` uses those definitions to compare a StructureDefinition to its parent. 

When I run this branch against mcode, I get a lot of warnings about GoFSH being unable to find certain USCore definitions. These warnings go away if I add ` --dependency hl7.fhir.us.core@3.1.0` as a flag. But the configuration GoFSH creates correctly captures that dependency, so it feels like we should be able to automatically tell GoFSH to load the files from US Core.

test/exportable/ExportableConfiguration.test.ts
```diff
@@ -64,5 +76,11 @@ describe('ExportableConfiguration', () => {
     it('should include the version when present', () => {
       expect(result).toMatch(/^version: 0\.13.0/m);
     });
+
+    it('should include the dependencies when present', () => {
+      expect(result).toMatch(/^dependencies:/m);
+      expect(result).toMatch(/hl7\.fhir\.us\.core: 3\.1\.0/m);
```
cmoesel, line 82: As it is, I think this would match `foohl7.fhir.us.core: 3.1.0bar`.  I think you can be more specific by adding start/end anchors:
```
      expect(result).toMatch(/^\s+hl7\.fhir\.us\.core: 3\.1\.0$/m);
```
I didn't test it, but I *think* it should work?

cmoesel (commented):
Awesome.  I tested it and it works well.  I tried it using both the complex representation (w/ `id`) and the simple representation.  I had one minor comment about how to make your test a little stricter.  I also agree w/ Nick that it would be great if we took this dependency info and then actually _used_ it in the run of GoFSH!

ngfreiter (changes requested):
I noticed a few things that needed fixes.

This is a tricky problem, because I think the ideal approach would be to process the ImplementationGuides first, get a configuration, and then just use the dependencies specified on that configuration when processing the remaining resources. However, this is easier said than done since I think we would need to refactor some of the processing flow so that ImplementationGuides are treated differently from other resources. The approach you used here makes sense, I think it will just be a bit less efficient, since we will have to load the JSON files once to get dependencies, and then once again to actually stock the `LakeOfFHIR`. I know @cmoesel had mentioned that efficiency isn't as much of a concern in GoFSH, so I'll let him weigh in on this as well.

src/utils/Processing.ts
```diff
@@ -84,6 +84,23 @@ export function loadExternalDependencies(
   return dependencyDefs;
 }
 
+export function getIGDependencies(inDir: string): string[] {
+  const igFiles = getFilesRecursive(inDir).filter(file =>
+    file.startsWith(inDir + '/ImplementationGuide')
```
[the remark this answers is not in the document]
ngfreiter: Yeah, `path.join` is generally a pretty good way to please the demands of Windows.

src/utils/Processing.ts
```diff
@@ -85,17 +85,21 @@ export function loadExternalDependencies(
 }
 
 export function getIGDependencies(inDir: string): string[] {
-  const igFiles = getFilesRecursive(inDir).filter(file =>
-    file.startsWith(inDir + '/ImplementationGuide')
-  );
+  const inputFiles = getFilesRecursive(inDir);
```
ngfreiter, line 88: I think it would be best here to filter out non-JSON. So something like:
```suggestion
  const inputFiles = getFilesRecursive(inDir).filter(file => file.endsWith('.json'));
```

ngfreiter (commented):
One small change, but other than that, looking good.

cmoesel (commented):
I just tested this on mCODE and it definitely made a difference (for the better!).  So that's awesome!

One thing I noticed, however, is that it loops through *all* IG files and uses *all* their dependencies -- but the actual `config.yaml` is written with the dependencies of only a single selected IG file.  I'm a little concerned that this could cause some weird results if dependencies from a non-selected IG are used in processing but then not present in the generated `config.yaml`.  I also wonder what would happen if one IG file had a dependency on `foo@1.0.0` and another IG file had a dependency on `foo@2.0.0`.  It looks like this would load them both and then... who knows?

Last, it seems that if someone passes in a dependency on the command-line, that should also be included in the generated `config.yaml`, right?

src/processor/FHIRProcessor.ts
```diff
@@ -41,7 +40,67 @@ export class FHIRProcessor {
...
+          found = true;
+          break;
+        } else if (internalVersionArr[i] > externalVersionArr[i]) {
+          resolvedVersion = internalVersionArr.join('.');
+          break;
+        } else i++;
+      }
+    }
+    return resolvedVersion;
+  }
```
cmoesel, line 97: This is cool, but... instead of rolling our own semantic versioning comparison algorithm, I'd recommend we just use the insanely popular [semver](https://www.npmjs.com/package/semver) library from NPM. It will also better handle some of the edge cases that might come up.

src/utils/Processing.ts
```diff
@@ -40,6 +41,19 @@ export async function getResources(
   return resources;
 }
 
+export async function getConfig(
+  inDir: string,
+  defs: fhirdefs.FHIRDefinitions,
+  externalDeps: string[]
+): Promise<ExportableConfiguration> {
+  const lake = getLakeOfFHIR(inDir);
```
cmoesel, line 49: It seems that now we're calling `getLakeOfFHIR` twice; once here and once in `getResources`.  Since that is a fairly expensive operation, and since it can also cause log messages (which will now be duplicated), I think we should refactor the code so that we only do it once and `getConfig` and `getResources` can both use the same lake of FHIR.  In the end, this is all to just build up the `FHIRProcessor`, so maybe that is the part that can be refactored out and shared.

cmoesel (changes requested):
Cool!  This is definitely headed in the right direction.  I really like the idea of building up the config first and then just using that to load the dependencies.  Right on!

I found a few opportunities to simplify the code, one opportunity to use an NPM library, and one place that I think we need to streamline for efficiency (and to avoid redundant logs).  Thanks for pushing this forward!

src/processor/FHIRProcessor.ts
julianxcarter: This is revolutionary to me! Haha I had no idea this existed, I'm implementing this and the other comments into the next commit. 👍🏽

package.json
```diff
@@ -52,6 +52,7 @@
     "typescript": "^3.8.3"
   },
   "dependencies": {
+    "@types/semver": "^7.3.4",
```
cmoesel, line 55: I'm a little surprised that `semver` is working because the line above is only supposed to add the TypeScript definitions of semver.  I think you might be getting lucky because some other library pulled in `semver`.  But here is what it should be:
* `@types/semver` in the _dev_ dependencies
* `semver` in the dependencies

To fix this:
```
npm remove @types/semver --save
npm add semver --save
npm add @types/semver --save-dev
```

src/app.ts
```diff
@@ -58,12 +59,25 @@ async function app() {
...
+  inDir = getInputDir(inDir);
+
   // Load dependencies
   const defs = new fhirdefs.FHIRDefinitions();
-  const dependencies = program.dependency;
-  const dependencyDefs = loadExternalDependencies(defs, dependencies);
+  let dependencies = program.dependency;
+
+  // Trim empty spaces from commad line dependencies
+  dependencies = dependencies?.map((dep: string) => dep.trim());
```
cmoesel, line 69: Now that I'm seeing this together, I think you might as well simplify even further to:
```suggestion
  // Trim empty spaces from commad line dependencies
  const dependencies = program.dependency?.map((dep: string) => dep.trim());
```

src/utils/Processing.ts
```diff
@@ -21,15 +22,16 @@ export function ensureOutputDir(output = path.join('.', 'gofsh')): string {
   logger.info(`Using output directory: ${output}`);
   return output;
 }
-
-export async function getResources(
-  inDir: string,
-  defs: fhirdefs.FHIRDefinitions
-): Promise<Package> {
+export async function getFhirProcessor(inDir: string, defs: fhirdefs.FHIRDefinitions) {
```
cmoesel, line 25: I don't think this function is really `async`, is it?  As far as I can tell, it doesn't `await` anything and it doesn't return a `Promise` (or at least it doesn't need to).

src/utils/Processing.ts
```diff
@@ -21,15 +22,16 @@ export function ensureOutputDir(output = path.join('.', 'gofsh')): string {
...
   const lake = getLakeOfFHIR(inDir);
   const igIniIgPath = getIgPathFromIgIni(inDir);
   const fisher = new MasterFisher(lake, defs);
   const processor = new FHIRProcessor(lake, fisher, igIniIgPath);
+  return processor;
```
cmoesel, line 30:
```suggestion
  return new FHIRProcessor(lake, fisher, igIniIgPath);
```

src/utils/Processing.ts
```diff
@@ -40,6 +42,14 @@ export async function getResources(
...
 }
 
+export async function getConfig(
+  processor: FHIRProcessor,
+  externalDeps: string[]
+): Promise<ExportableConfiguration> {
+  const config = processor.processConfig(externalDeps);
+  return config;
+}
+
```
cmoesel, line 52: I don't think this function is `async` anymore, but honestly, this function is so simple now that I don't think it's necessary anymore.  It's just as easy to invoke `processor.processConfig(externalDeps)` as it is to invoke `getConfig(processor, externalDeps)`.  So I'd just remove this extra layer of indirection.

cmoesel (changes requested):
I really like what you've done w/ the refactor.  I think that cleans it up nicely.

I have a few more suggestions for your consideration... see individual comments for details.

package.json
julianxcarter: Ahhh yes, I actually think this still works because I initially did import `server` and ran an `npm install`, then got confused and thought I'd need `@types/semver` haha. I'll revert this back in the next commit!

src/app.ts
```diff
@@ -76,7 +89,8 @@ async function app() {
 
   let resources: Package;
   try {
-    resources = await getResources(inDir, defs);
+    resources = await getResources(processor);
+    resources.add(config);
```
cmoesel, line 93: So... when I ran this against US Core I got some unexpected results.  After some investigation, I realized it is because the processor may need the config when it's processing.  In the `master` branch, the config is added to the `Package` (represented by the `resources` variable) _before_ it does any processing.  In this PR branch, it is added _after_ it does the processing (so the processing does not benefit from it).

You'll either need to pass the `config` around in a few function calls in order to get it into the `Package` prior to processing, or you need to construct the package here and pass _it_ around instead.  Either way, we're going to need to pass some stuff around.

If you're trying this yourself, here's what tipped me off to the problem.  If you run the `master` branch against US Core and then this branch against US Core, there will only be a few differences in the output -- but the differences will be that this branch does not remove the extension URL assignment rules that _should_ be removed by the `RemoveExtensionURLAssignmentRules` optimizer.

image

cmoesel (changes requested):
Found one more issue, detailed in my individual comment.

cmoesel (approved):
It's good!  Cleans up mCODE processing real nice!

ngfreiter (approved):
Good stuff, thanks for addressing everything.

PR #35, opened 2020-11-05
Extract ValueSetConceptComponentRule

mint-thompson:
Completes task [CIMPL-483](https://standardhealthrecord.atlassian.net/browse/CIMPL-483).

A ValueSet may contain included or excluded components that can be defined by as a FSH ValueSet concept component. These components must contain at least one concept definition, and therefore must also have a system specified. Other components are defined by a FSH ValueSet filter component, which are outside the scope of this extractor.


cmoesel:
Ah.  You beat me to it!  I also implemented `ProcessableValueSet` and some of this stuff in my support for extracting `ValueSetFilterComponentRule`.  But, you got the PR in first, so I'll wait until it's reviewed/merged and then reconcile my branch with it!

src/extractor/ValueSetConceptComponentRuleExtractor.ts
```diff
@@ -0,0 +1,39 @@
...
+
+export class ValueSetConceptComponentRuleExtractor {
+  static process(
+    vsComponent: ProcessableValueSetComponent,
+    include: boolean
+  ): ExportableValueSetConceptComponentRule {
+    if (vsComponent.concept?.length > 0) {
+      const rule = new ExportableValueSetConceptComponentRule(include);
+      rule.from.system = vsComponent.system;
+      rule.from.valueSets = cloneDeep(vsComponent.valueSet);
```
cmoesel, line 14: This will usually explicitly set at least one key to `undefined`, resulting in an object like:
```ts
{
  "system": undefined,
  "valueSets": ["http://foo.org/VS"]
}
``` 
or
```ts
{
  "system": http://foo.org/CS,
  "valueSets": undefined
}
```

Sometimes we care about that when we might check objects for existence or keys, but it's probably fine here, right?  I just like to double-check these sorts of things.

src/extractor/ValueSetConceptComponentRuleExtractor.ts
```diff
@@ -0,0 +1,39 @@
...
+      vsComponent.concept.forEach(concept => {
+        if (ValueSetConceptComponentRuleExtractor.isConceptComponent(concept)) {
```
cmoesel, line 16: Technically it's not valid FHIR if `code` isn't there.  I wouldn't be opposed to just skipping this check, because I actually think it's ok (and maybe even better) if the result reflects the fact that there was invalid FHIR source.  As it is now, we'd just swallow that.

src/processor/ValueSetProcessor.ts
```diff
@@ -14,14 +15,64 @@ export class ValueSetProcessor {
...
+
+  static extractValueSetComponentRules(
+    vsComponent: any,
+    target: ExportableValueSet,
+    include: boolean
+  ) {
+    if (ValueSetProcessor.isValueSetComponent(vsComponent)) {
+      target.rules.push(ValueSetConceptComponentRuleExtractor.process(vsComponent, include));
+    }
+    target.rules = compact(target.rules);
```
cmoesel, line 37: If we make this function `private` then I think we can safely remove the `compact` here and call it at the end of `extractRules` instead.  Otherwise we're calling it in multiple places once we merge in the care rule extractor.

src/processor/ValueSetProcessor.ts
```diff
@@ -14,14 +15,64 @@ export class ValueSetProcessor {
...
+  // by FHIR spec, if the include list exists, it must contain at least one element
+  // but we can still do some processing without that.
+  // see http://hl7.org/fhir/r4/valueset-definitions.html#ValueSet.compose.include
+  static isProcessableValueSet(input: any): input is ProcessableValueSet {
+    return input.name != null || input.id != null;
+  }
+
+  static isValueSetComponent(input: any): input is ProcessableValueSetComponent {
+    return typeof input.system === 'string' || Array.isArray(input.valueSet);
+  }
```
cmoesel, line 60: As noted above, I'm not sure it's such a bad thing if we process it even if there is no `input.system` or `input.valueSet`.  That should be very rare (since at least one is required by FHIR) and in the case that it does pop up, I think it's more appropriate to generate FSH that reflects the issue rather than pretend it didn't exist.  If skipping it really is the best option (because it really mucks with things in other code) then we should log an error.  But I think it's easier just to let the null value pass through as long as it doesn't wreak havoc elsewhere in GoFSH.

test/extractor/ValueSetComponentRuleExtractor.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    const input: ProcessableValueSetComponent = {
+      system: 'http://example.org/zoo',
+      concept: []
+    };
+    const result = ValueSetConceptComponentRuleExtractor.process(input, true);
+
+    expect(result).toBeNull();
+  });
+
+  it('should return null when the component has concepts, but none of them have a code', () => {
```
cmoesel, line 90: This is the type of thing I actually think we want to surface.

test/extractor/ValueSetComponentRuleExtractor.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    expect(result.inclusion).toBe(true);
+    expect(result.from).toEqual({
+      system: 'http://example.org/zoo'
+    });
+    expect(result.concepts).toHaveLength(2);
+    expect(result.concepts).toContainEqual(new FshCode('BEAR', 'http://example.org/zoo'));
+    expect(result.concepts).toContainEqual(new FshCode('LION', 'http://example.org/zoo', 'Lion'));
+  });
+
+  it('should extract a ValueSetConceptComponentRule from a component with a system, at least one value set, and at least one concept', () => {
```
cmoesel, line 32: Since FHIR allows for `valueSet` _without_ `system` should we also have one test to demonstrate that scenario?

cmoesel (commented):
This is looking good.  I left a few comments that I'd like your thoughts on.  They may result in no changes or perhaps they will require changes.  The most significant point of discussion is how we should hand source that is technically invalid FHIR input (e.g., when a required property is missing).

src/extractor/ValueSetConceptComponentRuleExtractor.ts
mint-thompson: I'm reasonably confident about this, based on my test behavior and the current implementation of `fromString` in `ExportableValueSetComponentRule.ts`.

src/processor/ValueSetProcessor.ts
mint-thompson: That's a good point. No reason to have this happen in the individual extractors when it can safely happen at the end.

src/extractor/ValueSetConceptComponentRuleExtractor.ts
mint-thompson: Agreed. I'll go put another quarter in the "goFSH is not a FHIR validator" jar, because I seem to have forgotten that important fact yet again.

test/extractor/ValueSetComponentRuleExtractor.test.ts
mint-thompson: This is something I considered, but intentionally left out. A concept component must have at least one concept, and when there is at least one concept, `system` is required. So, this scenario would be represented as a filter component. So, either `ValueSetProcessor.process` needs some logic to determine which rule extractor to use for a given component, or the extractors need some logic to determine when they've given a component that is valid, but not their responsibility.

cmoesel: Spot on.  I forgot that presence of a `concept` requires `system`.  Good call.

src/processor/ValueSetProcessor.ts
```diff
@@ -14,14 +15,64 @@ export class ValueSetProcessor {
     }
   }
 
+  static extractRules(input: ProcessableValueSet, target: ExportableValueSet): void {
```
ngfreiter, line 18: I noticed `extractRules` is not called in the `process` function. Is that intentional, because that line is added in #30? If we plan to merge #30 first, then that makes sense, but if this PR were to end up going in first, that should be added to `process`.

mint-thompson: #30 will probably be merged first, but might as well have it in both.

src/processor/ValueSetProcessor.ts
cmoesel: Based on your 👍 , I take it that you _intended_ to address this comment?

cmoesel (commented):
I think there was one other thing you intended to address but missed...

I ran this against US Core and got some long rules like this:
```
* include urn:oid:2.16.840.1.113883.6.238#1002-5 "American Indian or Alaska Native" and urn:oid:2.16.840.1.113883.6.238#2028-9 "Asian" and urn:oid:2.16.840.1.113883.6.238#2054-5 "Black or African American" and urn:oid:2.16.840.1.113883.6.238#2076-8 "Native Hawaiian or Other Pacific Islander" and urn:oid:2.16.840.1.113883.6.238#2106-3 "White" from system urn:oid:2.16.840.1.113883.6.238
```

That's actually not the longest, but anyway... I assume we'll want to change that to either:
```
* include
    #1002-5 "American Indian or Alaska Native" and
    #2028-9 "Asian" and
    #2054-5 "Black or African American" and
    #2076-8 "Native Hawaiian or Other Pacific Islander" and
    #2106-3 "White"
  from system urn:oid:2.16.840.1.113883.6.238
```
or
```
* urn:oid:2.16.840.1.113883.6.238#1002-5 "American Indian or Alaska Native" 
* urn:oid:2.16.840.1.113883.6.238#2028-9 "Asian"
* urn:oid:2.16.840.1.113883.6.238#2054-5 "Black or African American"
* urn:oid:2.16.840.1.113883.6.238#2076-8 "Native Hawaiian or Other Pacific Islander"
* urn:oid:2.16.840.1.113883.6.238#2106-3 "White"
```
But that's probably a job for an optimizer, right?

ngfreiter (approved):
Looks good to me!

src/processor/ValueSetProcessor.ts
```diff
@@ -71,9 +65,3 @@ interface ProcessableValueSet {
     exclude?: any;
   };
 }
-
-export interface ProcessableValueSetComponent {
```
cmoesel: I don't think that having a `ProcessableValueSetComponent` was such a bad thing, as it still helps devs avoid typos in property names and such.  But... I'm also OK with its omission.

cmoesel (approved):
Me too!

PR #36, opened 2020-11-09
Use IG pointed to by ig.ini for configuration

jafeltra:
This PR addresses [CIMPL-584](https://standardhealthrecord.atlassian.net/browse/CIMPL-584).

If there is an `ig.ini` file with an `ig` property, it will be used to identify the correct ImplementationGuide to be used for the Configuration and Instance processors. The approach for finding and parsing the `ig.ini` is similar to the one used by SUSHI.


src/processor/FHIRProcessor.ts
```diff
@@ -23,8 +27,11 @@ export class FHIRProcessor {
   process(): Package {
     const resources = new Package();
     let config: ExportableConfiguration;
+    const igForConfig =
+      this.lake.getAllImplementationGuides().find(doc => doc.path.endsWith(this.igPath)) ??
```
ngfreiter, line 31: I don't think `endsWith` works completely here. To test this out, I had a directory with an `ig.ini` contained in a folder, `test/ig.ini`, and then had that `ig.ini` point at an IG file at the same level as the `test` folder. So the path in my `ig.ini` was something like: `..\\ImplementationGuide-hl7.fhir.us.core.json`. But the path in goFSH to the ImplementationGuide file was just `ImplementationGuide-hl7.fhir.us.core.json`, which does not end with `..\\ImplementationGuide-hl7.fhir.us.core.json`, so the ImplemetationGuide is not found.

ngfreiter (changes requested):
I noticed a bit of a bug under certain circumstances.

package.json
```diff
@@ -51,11 +51,13 @@
     "typescript": "^3.8.3"
   },
   "dependencies": {
+    "@types/ini": "^1.3.30",
```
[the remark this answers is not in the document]
jafeltra: Oops yes!

src/processor/FHIRProcessor.ts
jafeltra: Hm I actually don't think I knew this was allowed, which is probably why I didn't test it. I'm on it!

jafeltra:
@ngfreiter these should be resolved in 8790f38 and afdb5e5!

ngfreiter (approved):
This looks good to me now!

julianxcarter (approved):
This looks good to me!

PR #37, opened 2020-11-16
Extract concept definition from CodeSystem

mint-thompson:
Completes task [CIMPL-453](https://standardhealthrecord.atlassian.net/browse/CIMPL-453).

Each concept in the CodeSystem is represented using an `ExportableConceptRule`. These rules have a pretty straightforward set of attributes, and there's no special logic that needs to be applied, so the code changes are not too complex.


src/processor/CodeSystemProcessor.ts
```diff
@@ -18,6 +18,9 @@ export class CodeSystemProcessor {
 
   static extractRules(input: any, target: ExportableCodeSystem, fisher: utils.Fishable): void {
     const newRules: ExportableCodeSystem['rules'] = [];
+    input.concept?.forEach((concept: any) => {
+      newRules.push(new ExportableConceptRule(concept.code, concept.display, concept.definition));
+    });
     newRules.push(...CaretValueRuleExtractor.processResource(input, fisher, input.resourceType));
```
ngfreiter, line 24: We might want to reverse the order of these, so that `CaretValueRule`s are written first, before the concepts, since I think that is the convention we generally follow.

ngfreiter (commented):
This looked good to me, outside of one suggestion.

jafeltra (dismissed):
I didn't see anything else. Assuming Nick's comment is addressed, this looks good to me, so I'm approving!

ngfreiter (approved):
Looks good!

jafeltra (approved):
I still approve!

PR #38, opened 2020-11-18
Combine CaretValueRules and AssignmentRules for Coding and Quantity types

mint-thompson:
Completes task [CIMPL-590](https://standardhealthrecord.atlassian.net/browse/CIMPL-590)

Coding and Quantity elements have convenient FSH syntax that let the author define more than one sub-element in a single FSH statement. CaretValueRules are combined on Profiles and Extensions, and AssignmentRules are combined on Instances. The presence of a rule defining a `code` leads to checks for potentially useful rules on sibling elements.

It is probably possible to parameterize and isolate the processing for Profiles/Extensions with Instances, but the small differences in how to handle CaretValueRules and AssignmentRules made it feel like it would be a better plan to keep it verbose.


cmoesel (changes requested):
I didn't end up having time for a full review, but here is one thing I noticed that may need further consideration:

You only consider caret rules in profiles and extensions, but assignment rules are totally possible in profiles and extensions too.  In fact, it's quite common to do something like assign an `Observation` profile's `code` element to a specific LOINC code.  It's a lot less common for quantities, but it still _could_ happen.

julianxcarter (approved):
I struggled with finding a real-world profile that's a good candidate to test this on, but the code looks exceedingly solid to me and the test cases seem to indicate that the optimizer is combining these rules as expected. I'm gonna combine these changes with PR #39 to test it with instances next since I think those cases are a bit more common, but pending that Im gonna go ahead and approve!

mint-thompson:
> I didn't end up having time for a full review, but here is one thing I noticed that may need further consideration:
> 
> You only consider caret rules in profiles and extensions, but assignment rules are totally possible in profiles and extensions too. In fact, it's quite common to do something like assign an `Observation` profile's `code` element to a specific LOINC code. It's a lot less common for quantities, but it still _could_ happen.

This is an important point. There are test cases in `AssignmentRuleExtractor.test.ts` that demonstrate these scenarios. The tests in the `#complex-values` section starting on line 94 deal with a variety of complex types, including Coding and Quantity. Therefore, I do not think that this optimizer needs to address those scenarios.

cmoesel:
@mint-thompson -- ah, ok.  At first that seemed a little inconsistent to me -- that we handle Quantities and Codings in an extractor in some cases and an optimizer in others.  But on second thought, I think I get it.  In profiles and extensions, Quantities and Codings are already represented as a whole (via `patternQuantity`, `fixedQuantity`, etc) -- so it make sense for the _extractor_ to handle them as a whole (and, in fact, it would be _wrong_ to split them up over multiple rules).

I think it would be helpful (for me) if you put a comment in the optimizer indicating that it does not need to handle assignment rules in extractors because quantities and codings are extracted as complete types via the `fixed[x]` and `pattern[x]` properties.  That way if I (or anyone else) ever goes into the optimizer looking for that code, we know why it's not there.

cmoesel (commented):
This looks good.  There is one thing I would like to discuss from a best practice point of view.  If a quantity value uses the `code`, `system`, `unit`, _and_ `value`, the optimizer produces rules like this:
```
* ^extension[0].valueQuantity = http://unitsofmeasure.org#GW "Gigawatts"
* ^extension[0].valueQuantity.value = 1.21
```

Personally, I would prefer the following, as it is more succinct and avoids the kind of weird _setting-a-quantity-to-a-code_ thing:
```
* ^extension[0].valueQuantity = 1.21 'GW'
* ^extension[0].valueQuantity.unit = "Gigawatts"
```

Did you intentionally favor one over the other or is this just a case you did not consider?  If the latter, how do you feel about my proposed change?

mint-thompson:
@cmoesel The behavior is intentional, but I hadn't really considered the alternative case. I mostly just had that first example style on my mind due to recent work to correctly handle that case. The second style you show is a bit cleaner, although importantly it can only be used when the system is `http://unitsofmeasure.org`. If the system is anything else (uncommon for quantities, but permitted), only the first style can be used.

Given that this is an optimizer, I think it's worthwhile to take advantage of the second style in situations where it is possible to do so. I'll update the optimizer accordingly.

cmoesel:
@mint-thompson and I discussed this a few days ago and agreed it would be good to also support `CodeableConcept` in this PR so that this:
```
* ^jurisdiction[0].coding[0].system = "urn:iso:std:iso:3166"
* ^jurisdiction[0].coding[0].code = #US
* ^jurisdiction[0].coding[0].display = "United States of America"
```
gets converted to this:
```
* ^jurisdiction[0] = urn:iso:std:iso:3166#US "United States of America"
```
instead of this:
```
* ^jurisdiction[0].coding[0] = urn:iso:std:iso:3166#US "United States of America"
```

src/optimizer/plugins/CombineCodingAndQuantityValuesOptimizer.ts
```diff
@@ -137,3 +146,46 @@ export default {
...
+function moveUpCaretValueRule(
+  rule: ExportableCaretValueRule,
+  sd: ExportableProfile | ExportableExtension,
+  knownSiblings: ExportableCaretValueRule[]
+): void {
+  // if the caretPath ends with coding[0], and there are no sibling rules that use indices on coding,
+  // we can move the rule up a level to take advantage of how SUSHI handles rules on CodeableConcepts.
+  if (rule.caretPath.endsWith('coding[0]')) {
+    const basePath = rule.caretPath.replace(/\.coding\[0]$/, '');
+    const siblings = sd.rules.filter(
```
cmoesel, line 159: I'd change this to:
```ts
const hasOtherSiblings = sd.rules.some(
```
Then in line 167 you'd have:
```ts
if (!hasOtherSiblings) {
```
To me, that's just a little more intentional -- and the use of `some` is more efficient than `filter` anyway (although efficiency isn't really a strong goal for this tool).

src/optimizer/plugins/CombineCodingAndQuantityValuesOptimizer.ts
```diff
@@ -137,3 +146,46 @@ export default {
...
+      otherRule =>
+        otherRule instanceof ExportableCaretValueRule &&
+        !knownSiblings.includes(otherRule) &&
+        rule !== otherRule &&
```
cmoesel, line 163: It's kind of silly, but I'd swap 162 and 163 -- as I think "don't count yourself" feels like something you'd naturally check first.  I can't explain why.  I just feel it.

I wasn't actually going to even suggest this until I got to lines 183-184 below and saw that _there_ you did put the self-check first.

src/optimizer/plugins/CombineCodingAndQuantityValuesOptimizer.ts
```diff
@@ -137,3 +146,46 @@ export default {
...
+        rule.path === otherRule.path &&
+        otherRule.caretPath.startsWith(`${basePath}.coding[`)
+    );
+    if (siblings.length === 0) {
+      rule.caretPath = basePath;
+    }
+  }
+}
+function moveUpAssignmentRule(
```
cmoesel, line 172: Can you put a blank line between these function definitions?

src/optimizer/plugins/CombineCodingAndQuantityValuesOptimizer.ts
```diff
@@ -137,3 +146,46 @@ export default {
...
+  instance: ExportableInstance,
+  knownSiblings: ExportableRule[]
+): void {
+  // if the path ends with coding[0], and there are no sibling rules that use indices on coding,
+  // we can move the rule up a level to take advantage of how SUSHI handles rules on CodeableConcepts.
+  if (rule.path.endsWith('coding[0]')) {
+    const basePath = rule.path.replace(/\.coding\[0]$/, '');
+    const siblings = instance.rules.filter(
+      otherRule =>
+        rule !== otherRule &&
```
cmoesel, line 183: You're not checking if it is an `AssignmentRule`.  Is that because this will only happen in `Instance`s -- and Instances can only have `AssignmentRule` and `InsertRule` (and InsertRule never has a coded value)?

I'm a little torn on this one, but part of me wants the `AssignmentRule` check in case we ever add support for other type rules in instances.  But the other part of me says we'll probably never do that so it's fine.

I'll leave it to you to decide.

src/optimizer/plugins/CombineCodingAndQuantityValuesOptimizer.ts
cmoesel: BTW, same for 181/187 below.

cmoesel (changes requested):
I've tested this w/ mCODE and US Core and it's looking good.  I had a few minor suggestions, mostly around code style.

cmoesel (approved):
Looks good.  Tested again against mCODE and US Core -- and wow does this clean up the output nicely.  It reduced mCODE by almost 1000 lines!

julianxcarter (approved):
This looks good to me, great job @mint-thompson!!

PR #39, opened 2020-11-18
Extract Assignment Rules on Instances

jafeltra:
This PR adds support for extracting Assignment Rules on Instances ([CIMPL-481](https://standardhealthrecord.atlassian.net/browse/CIMPL-481)).

Right now, all values on an Instances JSON are converted to Assignment Rules. This does not yet include the optimization for combining codings. I followed a similar pattern to how we extract Caret Value Rules on StructureDefinitions.


src/processor/InstanceProcessor.ts
```diff
@@ -24,9 +26,48 @@ export class InstanceProcessor {
...
+      utils.Type.Extension,
+      utils.Type.Type
+    );
+
+    if (instanceOfJSON == null) {
+      logger.error(
+        `InstanceOf definition not found for ${input.id} and cannot export any Assignment Rules.`
+      );
+      return;
+    }
```
mint-thompson, line 46: I think that emitting an error here is correct, but it might be fine to proceed and add the assignment rules anyway. The value of having `instanceOfJSON` is it lets you know what values on the instance don't need rules. If `instanceOfJSON` isn't found, then all of the non-ignored property values on the instance will get rules.

mint-thompson (changes requested):
Thank you! I have one small suggestion.

src/processor/InstanceProcessor.ts
jafeltra: I thought about this and was a little unsure of the way to go on this, but I decided to not process the rules because we also use the `instanceOfJSON` to determine the types of the elements when adding the assignment rules. So without the `instanceOfJSON` we won't know if a string value in the FHIR JSON should really be a FSHCode or if it should stay a string. GoFSH could potentially make those rules, but then if we ran SUSHI on that FSH, SUSHI will log errors that it cannot fix values to mismatched types. I thought it was better to not add assignment rules that could be incorrect rather and avoid those SUSHI errors, but if you think it's better to add the rules regardless, I can make those changes.

mint-thompson: I keep going back and forth on "don't write things when we can't apply the proper logic" vs "write as much as you can, and emit errors when we can't apply the proper logic." I think in this case, since it's specifically _not_ a FHIR validation issue, I'm back on your side of not writing any rules. The error is not FHIR validity, it's GoFSH validity.

mint-thompson (approved):
I have been convinced, through logic.

src/processor/InstanceProcessor.ts
```diff
@@ -24,9 +26,48 @@ export class InstanceProcessor {
...
+      delete inputJSON[prop];
+    });
+    if (inputJSON.text?.status === 'generated') {
+      delete inputJSON.text;
+    }
+
+    const flatInstance = getPathValuePairs(inputJSON);
+    const flatInstanceOf = getPathValuePairs(instanceOfJSON);
+    Object.keys(flatInstance).forEach(key => {
+      if (flatInstanceOf[key] == null || !isEqual(flatInstance[key], flatInstanceOf[key])) {
```
ngfreiter, line 58: I don't think this check is having its intended effect. `flatInstance` is an Instance, and `flatInstanceOf` is a StructureDefinition, so `flatInstanceOf` wouldn't have a field at `key`. `flatInstanceOf` would have a snapshot which describes the elements in `flatInstance`, but I think this comparison is mixing the worlds of Instances and StructureDefinitions.

I don't think it is necessary to have a check like this at all. If something is set on the instance, and it isn't captured by a keyword, then I think we want to make a rule for setting that value. Certain elements on the StructureDefinition may be `1..1` and have a `fixed` or `pattern` value set, but the rules around which values SUSHI can infer are complicated, and I don't like the idea of trying to recreate those inference rules on GoFSH.

It's possible we could just take the logic directly from SUSHI, but if we do that, we should do it in an optimizer anyway, and I don't think it is within scope of this task.

src/processor/InstanceProcessor.ts
```diff
@@ -49,3 +90,6 @@ const CONFORMANCE_AND_TERMINOLOGY_RESOURCES = new Set([
   'NamingSystem',
   'TerminologyCapabilities'
 ]);
+
+// The properties that are handled via keywords
+const IGNORED_PROPERTIES = ['resourceType', 'id', 'meta'];
```
ngfreiter, line 95: Not sure we can totally ignore `meta`. There are several fields on `meta` not handled by any keywords.

ngfreiter (changes requested):
I ran this on US Core and noticed two issues with some edge cases:
1. When an instance has values that are children of primitives, they are represented using a prepended `_` in the JSON. In FSH, we do not add any `_`, we just specify the path, and SUSHI takes care of detecting that an element is a child of a primitive and adding `_` in the JSON. Currently GoFSH is adding the `_` to the output FSH, which causes an error for any child of a primitive.
2. Maybe out of scope of this PR, but when writing a FSH Code that has spaces in it, we are not using the quotation marks to contain the code, as is required by the FSH syntax (and expected by SUSHI).

I also had a few questions below.

src/processor/InstanceProcessor.ts
cmoesel: Hmmm... I had the same thought at first (that you can't output an instance at all if you can't find the `InstanceOf` -- because you need to know the types for the paths).  BUT... I think we actually _can_ find the types for the paths.  _Every_ instance should have a `resourceType` field, and the `resourceType` should _always_ be a valid FHIR resource.  Profiles cannot _add_ new types or _change_ existing types; they can only _narrow_ types.  Therefore, no matter what the profile is, we should be able to resolve the types for every element by using the resource from the `resourceType`.

So, it seems to me that if something is an instance of a profile, we should try to find the profile first.  But if that fails, we should fall back to the `resourceType` -- which we should _never_ fail.  If we fall back to the `resourceType`, however, we should issue a warning that we couldn't find the profile, so the generated instance may not be the optimal representation.

jafeltra: Ah that's a good point! I've implemented that in 7528e9b

src/processor/InstanceProcessor.ts
jafeltra: I think you're right. I think I had changed what `flatInstanceOf` was based off of a few times and hadn't reconsidered where it ended up. I took out the check in 76b4903.

src/processor/InstanceProcessor.ts
jafeltra: You're definitely right. I removed `meta` from the list of ignored properties. If the first profile is used as the `InstanceOf` keyword, I do remove it since we won't need to set an Assignment Rule for that. If that then leaves `profile` or `meta` empty, I also remove those properties. Fixed in 8736f5b.

jafeltra:
@ngfreiter I believe I've addressed `1.` in 6123202. It now gets rid of the `_` in paths of children of primitives. I have a separate story for `2.` so that will be handled separately from this PR. I do still get a lot of errors when running a round trip on US Core, but I'm going to look into if those are things I should address here.

jafeltra:
Alright, so I investigated the other errors I was getting from a round trip on US Core and addressed them. The main source of the errors was that GoFSH was not able to find the ElementDefinition for elements on `contained` resources and resources on `entry` on Bundles. GoFSH would not be able to find elements that are on the definition of a particular resource because most elements are not on the base `Resource` definition. If it couldn't find the ElementDefinition, it would just fix the values as strings. However, then SUSHI would appropriately check the types for the ResourceType of the contained resource and log an error that the fixed string value was not the correct type. I resolved this issue in de985a4 by recursively calling the `getFSHValue` function to look for the ElementDefinition of a contained resource on the contained resource's type.

I did look for similar functionality in SUSHI and couldn't quite see something that I should reuse in this case, but I'm willing to believe I either thought too much about this or missed a function I could reuse from SUSHI. There's a decent amount of path manipulation in this approach, which always makes me a little nervous.

src/exportable/common.ts
```diff
@@ -98,3 +98,14 @@ export function fshifyString(input: string): string {
     .replace(/\r/g, '\\r')
     .replace(/\t/g, '\\t');
 }
+
+export function splitOnPathPeriods(path: string): string[] {
+  return path.split(/\.(?![^\[]*\])/g); // match a period that isn't within square brackets
+}
```
ngfreiter, line 104: Is this function necessary? I think the paths we are using this function on are coming from FHIR definitions right. On SUSHI this function is needed when we are splitting a FSH path, so the path may have something like a URL within the parentheses. I don't think that could happen here. Let me know if I'm missing something here.

jafeltra: No I think you're right. I had been using a regular `split('.')` and then at the last minute decided to switch to use that function to be safe. But I think you're right that these will only be FHIR definition paths, so the regular `split` should be sufficient. I'll switch back!

jafeltra: Fixed in 7183f39

ngfreiter (approved):
Looks good to me!

mint-thompson (approved):

PR #40, opened 2020-11-18
ExportableInstance test for assignment rules

julianxcarter:
@ngfreiter had actually included code implementing the actual writing of FixedValueRules on ExportableInstances in PR #33, so this PR only includes a unit test which tests that writing assignment rules is supported.

### DIFF
--- a/test/exportable/ExportableInstance.test.ts
+++ b/test/exportable/ExportableInstance.test.ts
@@ -1,3 +1,4 @@
+import { FshCode } from 'fsh-sushi/dist/fshtypes';
 import { EOL } from 'os';
 import { ExportableInstance, ExportableAssignmentRule } from '../../src/exportable';
 
@@ -33,7 +34,8 @@ describe('ExportableInstance', () => {
     exInstance.instanceOf = 'Observation';
 
     const statusRule = new ExportableAssignmentRule('status');
-    statusRule.value = 'final';
+    const statusCode = new FshCode('final');
+    statusRule.value = statusCode;
     exInstance.rules.push(statusRule);
 
     const dateRule = new ExportableAssignmentRule('effectiveDateTime');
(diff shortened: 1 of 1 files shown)


jafeltra (commented):
I had one minor suggestion below, but besides that, this looks good! I think it tests the one additional piece for expecting assignment rules to be written!

test/exportable/ExportableInstance.test.ts
```diff
@@ -27,4 +27,26 @@ describe('ExportableInstance', () => {
     ].join(EOL);
     expect(i.toFSH()).toBe(expectedResult);
   });
+
+  it('should export an instance with assignment rules', () => {
+    const exInstance = new ExportableInstance('MyInstance');
+    exInstance.instanceOf = 'Observation';
+
+    const statusRule = new ExportableAssignmentRule('status');
+    statusRule.value = 'final';
```
jafeltra, line 36: I think that `final` is actually a `code` on Observation. I think this would be better represented using a `FSHCode` here. That would change the `expectedResult` below slightly.

jafeltra (commented):
Oops it would help if I saved the comment...

test/exportable/ExportableInstance.test.ts
julianxcarter: You're absolutely right! I'll change that to make it an actual code and change the example FSH as well. Nice catch!

test/exportable/ExportableInstance.test.ts
```diff
@@ -1,3 +1,4 @@
+import { FshCode } from 'fsh-sushi/dist/fshtypes';
```
jafeltra, line 1: I think the usual way we import this is `import { fshtypes } from 'fsh-sushi` and then access `FshCode` by doing `fshTypes.FshCode` or pulling `FshCode` off of `fshtypes` separately like `const  const { FshCode } = fshtypes;`.

jafeltra (commented):
This change looks good! I had a very minor comment related to the way we import `fshtypes`.

test/exportable/ExportableInstance.test.ts
julianxcarter: Ahhh yes, this was done as an auto import. The most recent commit updates this test to the usual motif that we use.

jafeltra (approved):
This looks good to me! Thanks for making those small fixes!

ngfreiter (approved):
Looks good to me!

PR #41, opened 2020-11-19
Extract Value Set Filters

cmoesel:
This PR extracts all sorts of filters from ValueSet definitions.  As a bonus, It also adds an optimizer that will resolve system and valueset URLs in the ValueSet definitions.

The optimizer has one potentially controversial feature that I cannot decide if it is good or bad.  It will resolve URLs to names from FHIR (or IG dependencies) as well.  Some optimizers already did that, but it's a little more obvious now. For example, the rule:
```
* include http://snomed.info/sct#449868002 "Current every day smoker"
```
gets optimized to:
```
* include SNOMED_CT "Current every day smoker"
```

This is because FHIR core defines the SNOMED-CT code system and has `SNOMED_CT` as its `name`, so we resolve to it.  On one hand, it's nice to have these built-in aliases like that.  On the other hand, it's not really an approach we've been encouraging.  We've been suggesting that external entities should be referenced using aliases to their URLs.  So, should we put in logic to _avoid_ resolving external URLs to names?


cmoesel:
BTW.  I expect that we'll soon implement an optimizer that introduces aliases.  So either way, future versions won't have as many URIs hanging around.

src/optimizer/plugins/ResolveValueSetComponentRuleURLsOptimizer.ts
```diff
@@ -0,0 +1,41 @@
...
+  ExportableValueSetFilterComponentRule
+} from '../../exportable';
+
+export default {
+  name: 'resolve_value_set_component_rule_urls',
+  description:
+    'Replace URLs in value set rules with their names (for local and FHIR Core URLs only)',
+
+  optimize(pkg: Package, fisher: utils.Fishable): void {
+    [...pkg.valueSets].forEach(vs => {
```
mint-thompson, line 16: This can just be `pkg.valueSets.forEach`

test/optimizer/plugins/ResolveValueSetComponentRuleURLsOptimizer.test.ts
```diff
@@ -0,0 +1,207 @@
...
+      const myPackage = new Package();
+      myPackage.add(valueset);
+      optimizer.optimize(myPackage, fisher);
+
+      const expectedRule = new ExportableValueSetFilterComponentRule(true);
+      expectedRule.from = { valueSets: ['ObservationStatus'] };
+      expect(valueset.rules).toContainEqual(expectedRule);
+    });
+
+    it('should not replace filter rule system url with the name of a core FHIR ValueSet when it is same as local code system name', () => {
```
mint-thompson, line 140: It took me a second to figure out why this is necessary, but you're right: SUSHI wouldn't be happy being told `from valueset ABC` when `ABC` is the name of a local CodeSystem.

test/optimizer/utils.test.ts
```diff
@@ -0,0 +1,73 @@
...
+      const result = resolveURL(
+        'http://hl7.org/fhir/StructureDefinition/Group',
+        [utils.Type.Resource, utils.Type.Profile, utils.Type.Extension],
+        fisher
+      );
+      expect(result).toBe('Group');
+    });
+
+    it('should not resolve a URL to a name that would not be supported by FSH', () => {
+      // FHIR Core's definition for http://terminology.hl7.org/CodeSystem/v2-0487 has name v2.0487! Totally outrageous!
```
mint-thompson, line 37: baffling but true

mint-thompson (approved):
Hooray for value set filter components!!

src/optimizer/plugins/ResolveValueSetComponentRuleURLsOptimizer.ts
cmoesel: Ha.  Of course.  Silly me.

cmoesel: Fixed in c6aede1

test/extractor/ValueSetFilterComponentRuleExtractor.test.ts
```diff
@@ -0,0 +1,200 @@
...
+    expect(filter).toEqual(expectedFilter);
+  });
+
+  it('should log an error for an unknown operator and treat it as a string-based rule', () => {
+    const input: fhirtypes.ValueSetComposeIncludeOrExclude = { system: 'http://mycodesystem.org' };
+    input.filter = [{ property: 'a', op: 'begets', value: 'b' }];
+    loggerSpy.reset();
+    const filter = ValueSetFilterComponentRuleExtractor.process(input, VALUESET, false);
+    expect(loggerSpy.getLastMessage('error')).toBe(
+      'Unsuported filter operator in ValueSet test-vs: begets'
```
jafeltra, line 182: Equivalent typo here

```suggestion
      'Unsupported filter operator in ValueSet test-vs: begets'
```

jafeltra (commented):
This PR looks really good to me! I just had one minor typo that I noticed that I figured was worth fixing since it was simple and could be seen by a user.

As for the optimizer resolving URLs to names from FHIR, it feels a bit odd to me to have GoFSH output an optimization that we don't encourage, since previously any optimizations are actually bringing the FSH closer to what we would encourage. However, it doesn't feel like it should be the responsibility of this PR to address that since it was an existing "feature" already. Although now that I wrote this out, I think Nick has a task to address this, so I guess it really isn't something for this PR and the question has been answered!

test/helpers/testdefs/StructureDefiniton-geolocation.json
```diff
@@ -0,0 +1,763 @@
+{
```
jafeltra, line 1: I think this makes literally zero difference, but now that I've seen it, I can't un-see it. I think there's a typo in the file name, it should be `StructureDefinition-geolocation.json`.

cmoesel: Oh man.

cmoesel: It's like typo city today.  I'm embarrassed.  Fixed in ae365f3.

jafeltra (approved):
This looks excellent to me! Thanks for making those small fixes!

cmoesel:
@mint-thompson -- when you get a chance, can you review my changes since your last review?

mint-thompson (approved):

PR #43, opened 2020-11-20
Update fsh-sushi dependency to ^1.0.0

mint-thompson:
Completes task INT-893.

Because the dependency was previously on a beta version of 1.0.0, no further code changes were necessary to support this dependency.


julianxcarter (approved):
Well, I don't see how this could go wrong!

ngfreiter (approved):

PR #44, opened 2020-11-20
Updated output style

julianxcarter:
Addresses [CIMPL-594](https://standardhealthrecord.atlassian.net/browse/CIMPL-594). 

Updates GoFSH to output in a Sushi 1.0.0 project structure.

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -53,7 +53,7 @@ export function writeFSH(resources: Package, outDir: string): void {
       logger.info(`Wrote config to ${configPath}.`);
     }
   } catch (error) {
-    logger.error(`Could not load output direcotry: ${error.message}`);
+    logger.error(`Could not write to output directory: ${error.message}`);
     process.exit(1);
   }
 }
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -43,12 +43,13 @@ export async function getResources(
 export function writeFSH(resources: Package, outDir: string): void {
   const exporter = new FSHExporter(resources);
   try {
-    const resourceDir = ensureOutputDir(path.join(outDir, 'input', 'fsh'));
+    const resourceDir = path.join(outDir, 'input', 'fsh');
+    fs.ensureDirSync(resourceDir);
     const outputPath = path.join(resourceDir, 'resources.fsh');
     fs.writeFileSync(outputPath, exporter.export());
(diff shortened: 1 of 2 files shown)


src/utils/Processing.ts
```diff
@@ -42,13 +42,19 @@ export async function getResources(
...
   const exporter = new FSHExporter(resources);
-  const outputPath = path.join(outDir, 'resources.fsh');
-  fs.writeFileSync(outputPath, exporter.export());
-  logger.info(`Wrote fsh to ${outputPath}.`);
-  if (resources.configuration) {
-    const configPath = path.join(outDir, 'config.yaml');
-    fs.writeFileSync(configPath, resources.configuration.toFSH());
-    logger.info(`Wrote config to ${configPath}.`);
+  try {
+    const resourceDir = ensureOutputDir(path.join(outDir, 'input', 'fsh'));
```
ngfreiter, line 46: I think instead of using `ensureOutputDir` here, we might want to just use `ensureDirSync` directly. The problem with `ensureOutputDir` is that it logs a message indicating the output directory, so since we call `ensureOutputDir` once in `app.ts`, and now once here, we get a somewhat confusing combination of log messages:
image

src/utils/Processing.ts
```diff
@@ -42,13 +42,19 @@ export async function getResources(
...
+    const outputPath = path.join(resourceDir, 'resources.fsh');
+    fs.writeFileSync(outputPath, exporter.export());
+    logger.info(`Wrote fsh to ${outputPath}.`);
+    if (resources.configuration) {
+      const configPath = path.join(outDir, 'config.yaml');
```
ngfreiter, line 51: SUSHI 1.0.0 will work with `config.yaml`, but prefers that the file is named `sushi-config.yaml` (so SUSHI logs a warning at `config.yaml`).

ngfreiter (changes requested):
Good work! I noticed two things that I think need to be changed.

mint-thompson (changes requested):
It looks like `Processing.test.ts` never got a test case for checking that a config file is created. So, since you will be updating the name of that file, could you also add a test case to check for that file?

ngfreiter (dismissed):
This looks good to me!

mint-thompson (dismissed):

ngfreiter (approved):

PR #45, opened 2020-11-21
Construct Inline Instances

ngfreiter:
This PR addresses [CIMPL-479](https://standardhealthrecord.atlassian.net/browse/CIMPL-479) by adding an optimizer that extracts inline instances from groups of rules on an instance.

To find groups of rules to extract, the optimizer looks for rules at one of the two paths:
* `contained.resourceType`
* `entry.resource.resourceType`

If such a path is found, all rules at the same level as that path are extracted to an inline instance. For example, if the path found is `contained[2].resourceType`, any rule starting with `contained[2]` is extracted to the inline instance, and removed from the original instance. Rules that set `resourceType`, `id`, or `meta.profile[0]` are used to set keywords, and are not added to the inline instance, but are still removed from the original instance. Finally, a rule pointing towards the newly created inline instance is added to the original instance.

I've written tests that I think should test this, but it's hard to be sure I'm not making any bad assumptions until we can test this on real examples, so I think we should wait to merge this until it can incorporate the changes from https://github.com/FHIR/GoFSH/pull/39.


src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
```diff
@@ -0,0 +1,69 @@
...
+import { ExportableAssignmentRule, ExportableInstance } from '../../exportable';
+
+export default {
+  name: 'construct_inline_instance',
+  description:
+    'Construct inline instances from groups of rules in a contained resource or a Bundle',
+
+  optimize(pkg: Package): void {
+    const inlineInstances: ExportableInstance[] = [];
+    [...pkg.instances].forEach(instance => {
```
mint-thompson, line 13: I think your test cases provide good coverage of instances (including `Bundle`) that contain inline instances. However, any resource that extends `DomainResource` may contain inline instances in its `contained` property. So, that would include `StructureDefinition`, `CodeSystem`, and `ValueSet`.

mint-thompson (changes requested):

src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
ngfreiter: Oh oops! Good point. I've tested this out in SUSHI, and it looks like we don't do the necessary magic on ValueSets and CodeSystems to allow for inline instances being set on `contained`. So something like
```
ValueSet: MyVS
* ^contained = MyInlineObservation
```
doesn't appear to work in SUSHI (unless I'm missing something). But it does work on a `Profile`, so I will add support (and a test) for those.

ngfreiter: Added in https://github.com/FHIR/GoFSH/pull/45/commits/6b430f958bd9f32991dd01c625e5258b38f20567

test/optimizer/plugins/ConstructInlineInstanceOptimizer.test.ts
```diff
@@ -0,0 +1,550 @@
...
+        const inlineInstanceRule1 = new ExportableAssignmentRule('contained[0]');
+        inlineInstanceRule1.value = 'Bar';
+        inlineInstanceRule1.isInstance = true;
+        const inlineInstanceRule2 = new ExportableAssignmentRule('entry[0].resource');
+        inlineInstanceRule2.value = 'Bam';
+        inlineInstanceRule2.isInstance = true;
+        expect(instance.rules).toEqual([inlineInstanceRule1, inlineInstanceRule2]);
+      });
+
+      it('should create an inline instance from a contained resource with non-numeric paths', () => {
```
cmoesel, line 125: Again, if you want to be super defensive against any possible changes to our code in the future, then... OK.  But otherwise I don't think we currently generate paths like this.

cmoesel (changes requested):
I agree that it would be good to wait for #39 so we can try this against some real IGs.  In the meantime, I've given you a few more things to think about...  The only suggestion that is probably _required_ is that I imagine if SUSHI supports `^contained` for profiles then it must support it for extensions too?

src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
```diff
@@ -0,0 +1,95 @@
...
+} from '../../exportable';
+
+export default {
+  name: 'construct_inline_instance',
+  description:
+    'Construct inline instances from groups of rules in a contained resource or a Bundle',
+
+  optimize(pkg: Package): void {
+    const inlineInstances: ExportableInstance[] = [];
+    [...pkg.instances, ...pkg.profiles].forEach(resource => {
```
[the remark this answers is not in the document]
ngfreiter: Wait, can extensions have a `^contained` element? I didn't think they inherited that.

src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
```diff
@@ -0,0 +1,95 @@
...
+        resource instanceof ExportableInstance
+          ? ExportableAssignmentRule
+          : ExportableCaretValueRule;
+      // First get all possible paths for which child elements should be extracted onto an inline instance
+      const basePaths = resource.rules
+        .filter(
+          rule =>
+            rule instanceof ruleType &&
+            (/^contained(\[\d+\])?\.resourceType$/.test(getRulePath(rule)) ||
+              /^entry(\[\d+\])?\.resource\.resourceType$/.test(getRulePath(rule)))
```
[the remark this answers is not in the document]
ngfreiter: It seems possible it's something we would add, and I'm a bit worried we would forget to specify that it runs after this one, since inline instances are such a specific case that we might not notice it things like this.

src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
```diff
@@ -0,0 +1,95 @@
...
+          ) {
+            profile = rule.value as string;
+          } else {
+            newInstance.rules.push(rule);
+          }
+        });
+
+        newInstance.id = id ?? `Inline-Instance-for-${resource.id}-${(generatedIdCount += 1)}`;
+        newInstance.instanceOf = profile ?? resourceType;
+        newInstance.usage = 'Inline';
```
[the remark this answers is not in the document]
ngfreiter: Ok, issue logged here https://standardhealthrecord.atlassian.net/browse/CIMPL-601

src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
```diff
@@ -0,0 +1,95 @@
...
+          resource.rules.push(inlineInstanceRule);
+        }
+      });
+    });
+    pkg.instances.push(...inlineInstances);
+  }
+} as OptimizerPlugin;
+
+function getRulePath(rule: ExportableSdRule): string {
+  return rule instanceof ExportableCaretValueRule ? rule.caretPath : rule.path;
```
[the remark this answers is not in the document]
ngfreiter: I don't think `caretPath` can be null, but I don't mind being explicit here, even if it's not as clean. I wasn't able to figure out how to get inlining to work, because TypeScript is not sure that the `rule` is a care or assignment, so `rule.caretPath` is not allowed by TS.

src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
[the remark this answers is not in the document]
cmoesel: Extension definitions are `StructureDefinition`s -- and `StructureDefinition` has `contained` -- so you should be able to use `^contained` in an Extension definition.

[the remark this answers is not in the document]
ngfreiter: Oh `^` rules, you strike once again!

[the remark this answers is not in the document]
cmoesel: Yeah -- last night I almost fooled myself into thinking it wasn't relevant to extensions -- mainly because they never stand on their own.  Right after I submitted my review, I almost retracted it, and then I remembered my reasoning for it, breathed a sigh of relief, cracked open a bottle of wine, lit up a fine Cuban cigar, and watched reruns of MASH all night.

Well, all of that is true except the last three bits.

cmoesel:
Changes look good.  Will wait for #39 before declaring victory though.

mint-thompson (approved):

ngfreiter:
@cmoesel, https://github.com/FHIR/GoFSH/pull/39 has been merged, so if you still like the look of this, I think we can merge.

cmoesel (changes requested):
Unfortunately, this doesn't seem to be working as well IRL as it does in the unit tests.  I ran it against generated mCODE definitions and ended up w/ replacements like this:

```
Instance: scenario1-mcode-patient-bundle
InstanceOf: http://hl7.org/fhir/us/mcode/StructureDefinition/mcode-patient-bundle
Title: "scenario1-mcode-patient-bundle"
Description: "Extended example 1 as a mCODE Patient Bundle"
Usage: #example
* type = #collection
* entry[0].resource = "scenario1-mcode-cancer-patient"
* entry[0].fullUrl = "http://example.org/fhir/Patient/scenario1-mcode-cancer-patient"
* entry[1].resource = "scenario1-mcode-primary-cancer-condition"
* entry[1].fullUrl = "http://example.org/fhir/Condition/scenario1-mcode-primary-cancer-condition"
```
I think the references to the instances are supposed to be bare (not wrapped in `"` like strings).

In addition, it's generating a bunch of instances with _no name_ and with paths that match their original use in `Bundle` rather than the thing they're supposed to be an instance of.  For example:
```
Instance: 
InstanceOf: http://hl7.org/fhir/us/mcode/StructureDefinition/mcode-cancer-disease-status
Usage: #inline
* entry[2].resource.code.coding[0].code = #88040-1
* entry[2].resource.code.coding[0].system = "http://loinc.org"
* entry[2].resource.code.coding[0].display = "Response to cancer treatment"
* entry[2].resource.extension[0].url = "http://hl7.org/fhir/us/mcode/StructureDefinition/mcode-evidence-type"
* entry[2].resource.extension[0].valueCodeableConcept.coding[0].code = #363679005
* entry[2].resource.extension[0].valueCodeableConcept.coding[0].system = "http://snomed.info/sct"
* entry[2].resource.extension[0].valueCodeableConcept.coding[0].display = "Imaging (procedure)"
* entry[2].resource.status = #final
* entry[2].resource.category[0].coding[0].code = #imaging
* entry[2].resource.category[0].coding[0].system = "http://terminology.hl7.org/CodeSystem/observation-category"
* entry[2].resource.subject.reference = "Patient/scenario1-mcode-cancer-patient"
* entry[2].resource.effectiveDateTime = "2020-03-01"
* entry[2].resource.performer[0].reference = "Practitioner/scenario1-us-core-practitioner"
* entry[2].resource.focus[0].reference = "Condition/scenario1-mcode-primary-cancer-condition"
* entry[2].resource.valueCodeableConcept.coding[0].code = #268910001
* entry[2].resource.valueCodeableConcept.coding[0].system = "http://snomed.info/sct"
* entry[2].resource.valueCodeableConcept.coding[0].display = "Patient's condition improved (finding)"
```

ngfreiter:
Ok @cmoesel, those bugs should be fixed in https://github.com/FHIR/GoFSH/pull/45/commits/747d3f24d44a1411f9aeb9de7c96b0d493b44309, along with a bug that comes from repeating instance names, and having purely numeric instance names.

cmoesel (changes requested):
This is looking much better, but it also exposes the real root of some of the issues we were running into w/ duplicate `id`s and such.  mCODE uses a lot of `#example` instances _inline_ too.  As this optimizer is right now, we end up with many duplicate instances (e.g., `Instance: Inline-Instance-for-scenario1-mcode-patient-bundle-1` w/ `* id = "scenario1-mcode-cancer-patient"` _and_ `Instance: scenario1-mcode-cancer-patient`).

I think we need to avoid this, as this is something we wouldn't ever want to see in a real FSH project.  It seems to me that it might not be _too_ hard to do, as you'd basically just need to see if another instance already exists w/ the same `id` and then compare the generated rules from the inline instance to the existing instance.  This would also help w/ inline instances that are used in more than one place (which I think would currently suffer from the same problem).

src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
```diff
@@ -53,20 +53,31 @@ export default {
...
           } else if (
-            new RegExp(`${escapeRegExp(basePath)}\\.meta\\.profile(\\[0\\])?`).test(rule.path)
+            new RegExp(`^${escapeRegExp(basePath)}\\.meta\\.profile(\\[0\\])?$`).test(rule.path)
           ) {
             profile = rule.value as string;
           } else {
-            newInstance.rules.push(rule);
+            const inlineInstanceRule = cloneDeep(rule);
+            inlineInstanceRule.path = rule.path.replace(`${basePath}.`, '');
+            inlineInstanceRules.push(inlineInstanceRule);
```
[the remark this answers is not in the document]
ngfreiter: @cmoesel discussed this offline, but yes, we decided the SUSHI fishers will still find the Instance.

src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
```diff
@@ -53,20 +53,31 @@ export default {
             rule = newRule;
           }
           // id and resourceType and meta.profile should be used for keywords, all other rules are added
-          if (rule.path === `${basePath}.id`) {
+          if (
+            rule.path === `${basePath}.id` &&
+            isNaN(parseInt(rule.value as string)) &&
+            ![...pkg.instances, ...inlineInstances].find(
+              instance => instance.name === ((rule as ExportableAssignmentRule).value as string)
```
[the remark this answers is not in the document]
ngfreiter: Yes, for some reason TypeScript makes me do it, I'm not really sure why, but had to please the TS lords.

cmoesel:
I don't remember clicking the merge button.  Oops.  Sorry, @ngfreiter!

ngfreiter:
> This is looking much better, but it also exposes the real root of some of the issues we were running into w/ duplicate `id`s and such. mCODE uses a lot of `#example` instances _inline_ too. As this optimizer is right now, we end up with many duplicate instances (e.g., `Instance: Inline-Instance-for-scenario1-mcode-patient-bundle-1` w/ `* id = "scenario1-mcode-cancer-patient"` _and_ `Instance: scenario1-mcode-cancer-patient`).
> 
> I think we need to avoid this, as this is something we wouldn't ever want to see in a real FSH project. It seems to me that it might not be _too_ hard to do, as you'd basically just need to see if another instance already exists w/ the same `id` and then compare the generated rules from the inline instance to the existing instance. This would also help w/ inline instances that are used in more than one place (which I think would currently suffer from the same problem).

This turned out to be a little harder than I thought it would be, but it is added in https://github.com/FHIR/GoFSH/pull/45/commits/857bf04e1df152a66b48c28aab44911a78697e4b. As a bonus task that commit also adds instances to the optimizer that removes `text.status` and `text.div` when that `text` is generated. This was relevant because `text.div` will be generated differently if the instance is being used inside of another instance, and so it was messing with rule comparisons. I've tested it on mCODE and US Core, and to my eye it is working well on both of those, with the only differences I'm seeing being due to `text.status` or `text.div`.

src/optimizer/plugins/RemoveGeneratedTextRulesOptimizer.ts
```diff
@@ -32,5 +33,45 @@ export default {
...
+    [...pkg.instances].forEach(instance => {
+      const rulesToRemove: number[] = [];
+      if (hasGeneratedText(instance)) {
+        instance.rules.forEach((rule: ExportableRule, i: number) => {
+          if (rule instanceof ExportableAssignmentRule && rule.path.match(/^text\./)) {
+            rulesToRemove.push(i);
+          }
+        });
+      }
+      pullAt(instance.rules as ExportableRule[], rulesToRemove);
```
cmoesel, line 46: I think maybe we're using `pullAt` more than necessary in our optimizers now.  I think it makes sense to use when we might have multiple loops or other reasons for not messing w/ indexes while we process rules, but in cases like above, can't we just get away with:
```ts
if (hasGeneratedText(instance)) {
  instance.rules = instance.rules.filter(rule => {
    return !(rule instanceof ExportableAssignmentRule && rule.path.match(/^text\./));
  });
}
```
or using lodash [remove](https://lodash.com/docs/4.17.15#remove):
```ts
if (hasGeneratedText(instance)) {
  remove(instance.rules, rule => rule instanceof ExportableAssignmentRule && rule.path.match(/^text\./));
}
```

cmoesel: Also, I think the same would go for lines 22 - 34 above.

cmoesel (dismissed):
Wow -- de-duplicating definitely made a _huge_ difference.

I have a recommendation to tighten up some of the optimizer code.  I'd suggest doing it, but won't hold up the PR for it -- so it's up to you, @ngfreiter.

mint-thompson (dismissed):

src/optimizer/plugins/RemoveGeneratedTextRulesOptimizer.ts
ngfreiter: Cleaned up in https://github.com/FHIR/GoFSH/pull/45/commits/de97ccec2574c3fdcd07d9478992e1fecb91ff11

src/optimizer/plugins/RemoveGeneratedTextRulesOptimizer.ts
```diff
@@ -19,31 +19,24 @@ export default {
...
-        resource.rules.forEach((rule: ExportableRule, i: number) => {
-          if (
-            rule instanceof ExportableCaretValueRule &&
-            rule.path === '' &&
-            rule.caretPath.match(/^text\./)
-          ) {
-            rulesToRemove.push(i);
-          }
-        });
+        resource.rules = (resource.rules as ExportableSdRule[]).filter(
```
cmoesel, line 23: I guess TS is letting you say `resource.rules as ExportableSdRule[]`, but aren't some of the rules in `ExportableValueSet` and `ExportableCodeSystem` _not_ `ExportableSdRules`?  Wouldn't it be more correct to do something more generic like `resource.rules as fshrules.Rule[]`?

cmoesel (commented):
There's one thing that looked a little suspicious to me...

src/optimizer/plugins/RemoveGeneratedTextRulesOptimizer.ts
ngfreiter: Addressed in https://github.com/FHIR/GoFSH/pull/45/commits/a3cd779db79245595b46aabe2bd67c36e835a9fa, and also switched to using `remove`.

src/optimizer/plugins/RemoveGeneratedTextRulesOptimizer.ts
```diff
@@ -1,36 +1,69 @@
...
-        });
+        );
+      }
+    });
+
+    [...pkg.instances].forEach(instance => {
+      if (hasGeneratedText(instance)) {
+        instance.rules = instance.rules.filter(
+          rule => !(rule instanceof ExportableAssignmentRule && rule.path.match(/^text\./))
+        );
```
cmoesel, line 37: Maybe this is pedantic, but for consistency, it seems like since you used `remove` above, you should use `remove` here, since it's kind of a parallel situation.

cmoesel (changes requested):
Something in that last merge broke this feature -- because now when I run it on mCODE, all of those extra `Inline-instance-for-*` instances _are back_.  If I revert the last merge, it works as expected again.  I didn't fully investigate, but maybe it has to do w/ the order of optimizers?

Also, would it be possible to add a test that _might_ break in these types of situations?  I know we can't predict all the possibilities of things that might break it -- but a unit test of a fairly robust `#example` instance that is used inline might do the trick?

src/optimizer/plugins/RemoveGeneratedTextRulesOptimizer.ts
ngfreiter: https://github.com/FHIR/GoFSH/pull/45/commits/7ca90a984381e68d80c737b4acebd0347172a358

ngfreiter:
> Something in that last merge broke this feature -- because now when I run it on mCODE, all of those extra `Inline-instance-for-*` instances _are back_. If I revert the last merge, it works as expected again. I didn't fully investigate, but maybe it has to do w/ the order of optimizers?
> 
> Also, would it be possible to add a test that _might_ break in these types of situations? I know we can't predict all the possibilities of things that might break it -- but a unit test of a fairly robust `#example` instance that is used inline might do the trick?

Alright I've re-fixed this in https://github.com/FHIR/GoFSH/pull/45/commits/7ca90a984381e68d80c737b4acebd0347172a358. It was a result of an optimizer that resolves `InstanceOf` values from a URL to a name. This would happen on the non-inline copy of an instance, but not the inline copy, so the inline copy wouldn't appear to be a duplicate of the non-inline copy, and the extra instances would get created. So now the `ConstructInlineInstancesOptimizer` runs before that other optimizer.

I added a test in https://github.com/FHIR/GoFSH/pull/45/commits/cc1f5b646611537f8b46df1e83427fd951a44f97 that tests that the de-duplication of instances is working correctly.

cmoesel (approved):
Just ran it on mCODE, and it's so nice... Also confirmed that the new test failed before the fix and passes after it, so we're good on that.  Thanks!

mint-thompson (approved):

PR #46, opened 2020-11-22
ValueSetComponentRules may export to multiple lines

mint-thompson:
Completes task [CIMPL-576](https://standardhealthrecord.atlassian.net/browse/CIMPL-576).

When the FSH export from an ExportableValueSetComponentRule is more than 100 characters long, break it up into multiple lines. Each line in a filter component will begin with "and" or "where". Each line in a concept component will begin with "and" or "from".


src/exportable/ExportableValueSetComponentRule.ts
```diff
@@ -15,6 +16,10 @@ export class ExportableValueSetConceptComponentRule extends fshrules.ValueSetCon
     let fsh = `* ${this.inclusion ? 'include' : 'exclude'} `;
     fsh += this.concepts.map(concept => concept.toString()).join(' and ');
     fsh += fromString(this.from);
+    // if the result is more than 100 characters long, break it up
+    if (fsh.length > 100) {
+      fsh = fsh.replace(/ (and|from) /g, `${EOL}  $1 `);
```
ngfreiter, line 21: I think this may be an overly aggressive replacement. Codes may have displays, and those displays could have the words `and` and `from` in them, resulting in output that looks like this:
image

Which doesn't cause errors in SUSHI, but does change the `display` produced by SUSHI.

Once [CIMPL-599](https://standardhealthrecord.atlassian.net/browse/CIMPL-599) is done, we could run into codes that have spaces in them, and those codes could have `and`, `from`, or `where` in them. In that case, if GoFSH splits the quoted code across multiple lines, it will cause errors in SUSHI.

src/exportable/ExportableValueSetComponentRule.ts
```diff
@@ -39,13 +44,16 @@ export class ExportableValueSetFilterComponentRule extends fshrules.ValueSetFilt
           }`
       )
       .join(' and ');
+    // if the result is more than 100 characters long, break it up
+    if (fsh.length > 100) {
+      fsh = fsh.replace(/ (and|where) /g, `${EOL}  $1 `);
```
ngfreiter, line 49: Same concerns here as on line 21.

ngfreiter (changes requested):
I think this could run into some issues with certain `display` strings, or with certain cruel and unusual codes.

mint-thompson:
I had a feeling such a small change was too good to be true. I'll revise this to separate ahead of time.

cmoesel (commented):
Technically this seems OK to me, but stylistically I don't think it represents common practice.  

From what I've seen in the [documentation](https://hl7.org/fhir/uv/shorthand/reference.html#filters) and in [mCODE](https://github.com/HL7/fhir-mCODE-ig/blob/master/input/fsh/VS_Other.fsh#L33-L42) (which mainly counts because it was written by a FSH designer), it is preferable to list a rule-per-code.  So instead of:
```
* include #foo and #bar and #baz from https://foobar.org/codes
```
this:
```
* https://foobar.org/codes#foo
* https://foobar.org/codes#bar
* https://foobar.org/codes#baz
```
Personally, I prefer that as well (especially if/when we replace the URLs w/ aliases).  For large value sets, I'd rather mentally parse dozens of separate rules than one very large compound rule (for which I won't even know the system until I get to the very end of it).  For consistency I'd even suggest we do it that way for small value sets (even if there is only _one_ code from the system).  If we did that, then the only case we'd still use the `from system` syntax is if there are other qualifiers (like `from valueset`).

The other stylistic thing, which is even more a matter of stylistic preference, is how spaces and indenting are used. I actually prefer that if a rule gets split over multiple lines that the following lines have a larger hanging indent.  The FSH documentation also seems to prefer this, as its [section on whitespace](https://hl7.org/fhir/uv/shorthand/reference.html#whitespace) suggests reformatting:
```
* component contains appearanceScore 0..3 and pulseScore 0..3 and grimaceScore 0..3 and activityScore 0..3 and respirationScore 0..3
```
to:
```
* component contains
    appearanceScore 0..3 and
    pulseScore 0..3 and
    grimaceScore 0..3 and
    activityScore 0..3 and
    respirationScore 0..3
```
Also note that it prefers _trailing_ `and` rather than _leading_ `and` (likely to keep focus on the meat of the rule rather than prepositions).  mCODE does not have any VS rules complex enough to require multi-line, but we see the same wrapping style in its [slice declarations](https://github.com/HL7/fhir-mCODE-ig/blob/master/input/fsh/SD_Genomics.fsh#L34-L43) (although we shouldn't be surprised since the same person wrote the doc as authored mCODE).  I think I personally agree with this as well, but I realize that others may have different opinions.

Now, honestly, if I were writing it out, I might even be more sophisticated w/ indenting.  For example, I might have written one of your test cases as:
```
* include codes
    from
      system http://fhir.example.org/myImplementationGuide/CodeSystem/AppleCS and
      valueset http://fhir.example.org/myImplementationGuide/ValueSet/BananaVS and
        http://fhir.example.org/myImplementationGuide/ValueSet/CupcakeVS
    where display = "this and that" and
      version exists true and
      concept descendent-of #comestible
```
... but even I will admit that's probably going a little too far.

cmoesel (approved):
This looks great and I'm approving as-is, but I noted that I don't think there are any tests for `exclude`.  In my manual tests, I see it is working, but it might be nice to have a unit test for excluded codes and/or filters.

ngfreiter (approved):
Looks good to me!

cmoesel (approved):
Super sweet!

PR #47, opened 2020-11-26
Process unsupported ValueSets as Instances

cmoesel:
If a ValueSet uses paths that cannot easily be represented using FSH VS syntax, process it as an Instance instead.

The tests demonstrate it, but if you want to manually test it, find an existing project of FHIR definitions and modify one of the ValueSets so that one of its `compose.include` components has an id or extension or concept designation somewhere inside it.  With master, it will generate it as a `ValueSet`.  With this PR, it will generate it as an `Instance`.

NOTE: Instance processing is not complete in `master` yet.  Currently it only exports Instance keywords.

### DIFF
--- a/src/processor/FHIRProcessor.ts
+++ b/src/processor/FHIRProcessor.ts
@@ -62,14 +62,14 @@ export class FHIRProcessor {
         logger.error(`Could not process CodeSystem at ${wild.path}: ${ex.message}`);
       }
     });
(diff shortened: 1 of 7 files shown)


mint-thompson (approved):
Looks good to me.

ngfreiter (approved):
Looks good to me as well!

PR #48, opened 2020-11-30
Process unsupported CodeSystems as Instances

cmoesel:
Like #47 but for `CodeSystem`s.

If a CodeSystem uses paths that cannot easily be represented using FSH CS syntax, process it as an Instance instead.

The tests demonstrate it, but you can also test with US Core because US Core uses nested `concept`s and `concept` properties (neither of which are supported w/ FSH CodeSystem syntax).  With master, GoFSH will generate `CodeSystem` definitions for `RaceAndEthnicityCDC` and `USCoreConditionCategoryExtensionCodes`. With this PR, those code systems will be generated as `Instance`s instead and you'll notice the extra information that was previously being dropped.

P.S. This also includes a few very minor changes to the ValueSet processing (for consistency).


src/processor/CodeSystemProcessor.ts
```diff
@@ -26,8 +33,8 @@ export class CodeSystemProcessor {
   }
 
   static process(input: any, fisher: utils.Fishable): ExportableCodeSystem {
-    // We need something to call the CodeSystem, so it must have a name or id
-    if (input.name != null || input.id != null) {
+    // It must be representable using the FSH CodeSystem syntax
+    if (this.isProcessableCodeSystem(input)) {
```
ngfreiter, line 37: This check is not strictly necessary anymore right? Because we only sent `ProcessableCodeSystems` into here from the `FHIRProcessor`? I don't really mind having the double checks, since I think if we wanted to get rid of them we'd have to do some refactoring to make the types work which may not be worth the benefit, mostly this is just a question for me to make sure I'm not missing something.

ngfreiter (commented):
I tested this on US Core, and those CodeSystems you mentioned are now represented as Instances, but they are still referenced in other places by name as if they are a CodeSystem. For example, the rule:
```
* include USCoreConditionCategoryExtensionCodes#health-concern "Health Concern" from system USCoreConditionCategoryExtensionCodes
```
causes an error, since `USCoreConditionCategoryExtensionCodes` is being represented as an instance, and apparently you cannot reference an instance by name. I had not realized this limitation of SUSHI.

Instead, I think we want to reference the Instance by URL directly, which I think would be a case where we need to stop the optimizer from making this optimization.

src/processor/CodeSystemProcessor.ts
cmoesel: Technically speaking, that is true, since we only call this from code that first filters out the unprocessable CodeSystems.  That said, since this is a public function, that could be called (in theory) from anywhere by anyone, I'd lean toward keeping the check.  As you note, it also helps w/ the typing.

cmoesel:
> For example, the rule _\[omitted]_ causes an error, since `USCoreConditionCategoryExtensionCodes` is being represented as an instance, and apparently you cannot reference an instance by name. I had not realized this limitation of SUSHI.
> 
> Instead, I think we want to reference the Instance by URL directly, which I think would be a case where we need to stop the optimizer from making this optimization.

I don't think that _how_ you create a `CodeSystem` should affect the ways in which it can be referenced.  For instance (no pun intended), I think that SUSHI _does_ allow you to reference dependency IG `CodeSystem`s (which are JSON) by name -- and probably BYOJ `CodeSystem`s too (although I'm only 90% confident about that).  So I think it would be weird to say that `CodeSystem`s authored as an `Instance` can only be referenced by URL even though they're in the same IG.  So my vote is to actually fix this in SUSHI.

In fact, looking at the SUSHI `FSHTank.fish` function implementation, I think it _still_ would not work even if we referenced `CodeSystem`s by URL -- because the real problem is that when you tell the fisher to fish for `CodeSystem`s, it doesn't look at `Instance`s at all!  BUT... now we potentially have a bigger problem if code that fishes for `CodeSystem` expects `FSHCodeSystem`s in return, so we'd need to watch out for that...

cmoesel:
Actually, I guess if it was a URL that SUSHI wouldn't fish at all for it, would it?  So nevermind that part about it not working in SUSHI even if it's a URL. ;-)

ngfreiter:
> I don't think that _how_ you create a `CodeSystem` should affect the ways in which it can be referenced. For instance (no pun intended), I think that SUSHI _does_ allow you to reference dependency IG `CodeSystem`s (which are JSON) by name -- and probably BYOJ `CodeSystem`s too (although I'm only 90% confident about that). So I think it would be weird to say that `CodeSystem`s authored as an `Instance` can only be referenced by URL even though they're in the same IG. So my vote is to actually fix this in SUSHI.

I agree this should be fixed in SUSHI. Seems like a limitation of SUSHI, not something that should be a feature, and one that could very easily confuse people I think. But until this is fixed in SUSHI, I'm not sure what we want to do with this PR. I guess we have already merged the sibling PR, so maybe it makes sense to just ignore this and let it be fixed in SUSHI. But it makes me a little uncomfortable, since in US Core for example, this PR fixes some incomplete CodeSystems, but at the cost of creating some invalid ValueSets.

jafeltra (commented):
I haven't noticed anything else on this PR besides the one issue with referencing CodeSystems by name that @ngfreiter pointed out. So once there's a decision on that, I'm good to approve!

cmoesel:
@ngfreiter and @jafeltra -- Nick's comments are addressed in my last commit.  Also, oops -- I didn't mean to rebase, but... I did  -- so had to force push the branch.  Sorry!

jafeltra (commented):
I had one minor comment inline below. Also, I ran this on US Core and when I run SUSHI on the generated FSH, I get a lot of errors that seem to be related to the Assignment Rules that are being added to the CodeSystem instances. I didn't have these when I ran this the other day and I'm wondering if my PR for Assignment Rules on Instances hadn't been included in this branch at that time. I'm happy to try to help resolve some of these issues.

src/optimizer/utils.ts
```diff
@@ -10,15 +12,29 @@ import { utils } from 'fsh-sushi';
...
+    fisher.lakeOfFHIR.fishForFHIR(url, ...types) != null
+  ) {
+    return url;
+  }
+
   // NOTE: Testing against a regex from FHIR because some FHIR core definitions have names that are
   // invalid against the spec!  Good heavens!
   if (
+    // !isUnsupportedCodeSystem &&
+    // !isUnsupportedValueSet &&
```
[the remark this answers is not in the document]
cmoesel: How embarrassing!  I just forgot to remove those lines!  That's from an earlier implementation!

cmoesel:
@jafeltra said:
> Also, I ran this on US Core and when I run SUSHI on the generated FSH, I get a lot of errors that seem to be related to the Assignment Rules that are being added to the CodeSystem instances.

Ugh.  I was in a hurry to update this PR before heading out and I missed running SUSHI on the generated output.  You are right (no surprise)!  I haven't looked at all the errors, but it seems that a lot of them are actually related to the `CombineCodingAndQuantityValueOptimizer` being too aggressive and converting `CodeSystem.concept` values to use the FSH concept syntax even though `CodeSystem.concept` is not actually a `CodeableConcept`, `Coding`, or `code`.

We should discuss this with the team.  Either we need to make it so the optimizer is not so aggressive (likely requiring the optimizer to be aware of the actual types associated with paths) or we need to update SUSHI to treat `CodeSystem.concept` as a special case.  I think the latter actually kind of makes sense, _but_ it might actually not be technically correct against the spec, so I'm not sure if we can do it. (Probably not).

jafeltra:
I looked more closely at the errors I'm getting and I think you're right that most, if not all, are related to combining `concept` into FSH concepts. There are two main error types that I see - one about the type we are assigning to `concept`s and others about `concept.code` having a minimum cardinality of 1 but occuring 0 times. So I think that addressing the issue you pointed out should resolve all the errors. I'm happy to discuss an approach with the team!

For what it is worth, I no longer get the original errors from referencing CodeSystems by name!

cmoesel:
@jafeltra and @ngfreiter -- I've addressed the issue Julia noticed in e960da6 and cadbabb.  The optimizer that combines code rules now attempts to check the type at the path it is optimizing, and will not optimize if it's not a supported type for the combination.

I'm embarrassed by how long it took me to come up w/ this solution.  There were a few iterations...

jafeltra (approved):
This looks good to me! The implementation makes sense, and I compared the SUSHI round trip on US Core on master and on this branch and everything looks the same.

ngfreiter (approved):
Looks good to me, but looks like there are some conflicts to resolve.

cmoesel:
> Looks good to me, but looks like there are some conflicts to resolve.

Nooooooooooooooooooooooooooooo!
image

cmoesel:
@ngfreiter @jafeltra - merge is done.  TBH, it was _not_ trivial, but all the tests pass now and I manually tested against mCODE and USCore and confirmed it still works as expected (as best I can tell).

jafeltra (approved):
I compared the output I got from GoFSH and SUSHI from before and after the merge, and the only differences I saw were from Aliases, so I think still looks good!

ngfreiter (approved):
Looks good!

PR #49, opened 2020-12-02
Replace URLs with Aliases

ngfreiter:
This addresses [CIMPL-596](https://standardhealthrecord.atlassian.net/browse/CIMPL-596).

Aliases are now tracked in the `Package` and exported at the top of the `resource.fsh` file. Aliases are created in optimizers which resolve URLs. If an optimizer attempts to resolve a URL to a FHIR definition and fails, the optimizer will then try to use an existing alias, or construct a new alias, for that URL. Several optimizers (`ResolveOnlyRuleURLsOptimizer`, `ResolveValueSetComponentRuleURLsOptimizer`, and `ResolveParentURLsOptimizer`) already existed to resolve URLs to FHIR definitions, so these optimizers have been extended to use aliases as well. This PR adds several other optimizers (`ResolveBindingRuleURLsOptimizer`, `ResolveInstanceOfURLsOptimizer`, `ResolveValueRuleURLsOptimizer`) to resolve URLs and aliases in places where they weren't yet being resolved, but should be.

Aliases are created from a URL using the last part of the URL path (`bar` in `http://example.com/foo/bar`) if it exists, or the first part of the domain (excluding `www`) otherwise (`example` in `http://www.example.com`). This is a bit arbitrary, but seemed to work pretty well in my testing on US Core and mCODE.

I'd suggest testing on both US Core and mCODE, since they cover a pretty broad variety of cases. You should see a significant difference in the `resources.fsh` generated by this branch and the `resources.fsh` generated by `master`, since this branch will replace many URLs with aliases (or FHIR resource names in the case of the new optimizers). But when you run SUSHI on the results, you should get the same output on this branch and `master`.


mint-thompson (approved):
This looks good! Hooray for aliases!

julianxcarter (approved):
This looks good and works well for me!

PR #50, opened 2020-12-08
Configurable Layout

ngfreiter:
This addresses [CIMPL-597](https://standardhealthrecord.atlassian.net/browse/CIMPL-597) by adding a command line option that can be used to configure the the style of the generated output files. The command line option is specified using `-s` or `--style`, and there are four options:
* `by-category`: Definitions are written to files based on what type of FSH entity they are (`Alias`, `Profile`, `Extension`, and so on). Note that this is now the default behavior.
* `by-type`: The same as `by-category` for non-Profiles and non-Instances, but now each Profile is written to a different file, and any Instances which are an example of that profile are written to that same file. So you may have a file like `CancerRelatedMedicationRequest.fsh` which contains the Profile `CancerRelatedMedicationRequest` along with several example Instances of that Profile.
* `single-file`: All definitions are written to one file. Note that this is no longer the default behavior.
* `file-per-definition`: Each standalone definition is written to an individual file. Since Invariants, Mappings, and Aliases don't really standalone, they are still all grouped into one single file (like `aliases.fsh`).

Feel free to suggest changes to how these options should work, or what the names of the options should be. Those aren't set in stone.


cmoesel:
I haven't tried this, but I've read your description and have one comment/question...

I'm wondering if `Invariant`s and `#inline` instances should be grouped w/ the things they're used by?  I feel like most authors would want their invariants in the same file as the profile/extension that uses them.  And the same for inline instances.  The only tricky part is that _technically_ those things can be used by more than one profile/extension (although, often, they are not).

ngfreiter:
> I haven't tried this, but I've read your description and have one comment/question...
> 
> I'm wondering if `Invariant`s and `#inline` instances should be grouped w/ the things they're used by? I feel like most authors would want their invariants in the same file as the profile/extension that uses them. And the same for inline instances. The only tricky part is that _technically_ those things can be used by more than one profile/extension (although, often, they are not).

In which mode? I think in `by-category` mode, Invariants and Instances should stay in their respective files, `invariants.fsh` and `instances.fsh`, since otherwise you aren't really grouping by category anymore. In `file-per-definition` mode or `by-type` mode this could make sense. To me this sounds like a different mode altogether thought. It would not be `file-per-definition` since a file could have an instance and then a bunch of inline ones defined as well, and it wouldn't be `by-type`, since a file could contain a profile, an example of that profile, and then several inline instances of different profiles. It seems more like a `by-use` type of mode, where things are grouped by where they are used.

src/export/FSHExporter.ts
```diff
@@ -5,68 +5,211 @@ import { logger } from '../utils';
...
+    }
+    for (const valueSet of this.fshPackage.valueSets) {
+      exports.set(`${valueSet.name}-ValueSet.fsh`, valueSet.toFSH());
+    }
+    for (const instance of this.fshPackage.instances) {
+      exports.set(`${instance.name}-Instance.fsh`, instance.toFSH());
+    }
+    for (const instance of this.fshPackage.instances) {
+      exports.set(`${instance.name}-Instance.fsh`, instance.toFSH());
+    }
```
jafeltra, line 136: It looks like maybe the instance code is duplicated unnecessarily here.

jafeltra (commented):
This looks really nice! I had one small comment inline but didn't notice anything else besides that. The one other thing that kind of keeps throwing me for a mini-loop is that the `by-category` and `by-type` names feel really similar to me and I keep forgetting which is which. But I've been trying to think of a more descriptive/different name for one of them and I can't really come up with any. Do they feel clear to others? Because if so, I think this is just a me-problem!

ngfreiter:
I'm going to be making changes to address @cmoesel's comment, so hold off on reviewing for now.

julianxcarter (approved):
Everything looks good and works well for me, so I'm gonna go ahead and approve! 

I kind of wish that the `by-type` and `by-category` options were more descriptively named since I don't think either is wholly representative of what they do. Maybe renaming `by-type` to `by-domain-type` and `by-category` to `by-profile-grouping`? Just a thought on semantics, also any confusion will be cleared up in documentation so it's no big deal. Nice work!

src/exportable/Exportable.ts
```diff
@@ -1,3 +1,18 @@
...
   toFSH(): string;
 }
+
+export type NamedExportable =
+  | ExportableProfile
+  | ExportableExtension
+  | ExportableValueSet
+  | ExportableCodeSystem
+  | ExportableInvariant
+  | ExportableMapping;
```
jafeltra, line 18: Is there a reason `ExportableInstance` is not considered a `NamedExportable`?

src/export/FSHExporter.ts
```diff
@@ -63,150 +75,190 @@ export class FSHExporter {
...
-    files.set(
-      'invariants.fsh',
-      this.fshPackage.invariants.map(invariant => invariant.toFSH())
-    );
-    files.set(
-      'mappings.fsh',
-      this.fshPackage.mappings.map(mapping => mapping.toFSH())
-    );
+    files.set('aliases.fsh', this.fshPackage.aliases);
+    files.set('extensions.fsh', this.fshPackage.extensions);
```
jafeltra, line 215: I can't remember if we already discussed this, but would it make sense to do a similar sorting for extensions, where they are put in the file of a profile where they're used if they're only used on one profile? Or does it make more sense to keep them separate since they are more of their own concept, independent from profiles?

src/export/FSHExporter.ts
```diff
@@ -63,150 +75,190 @@ export class FSHExporter {
...
-    files.forEach((content, fileName) => exports.set(fileName, content.join(`${EOL}${EOL}`)));
-    return exports;
+  private inlineInstanceUsedIn(
+    inlineInstance: ExportableInstance,
+    files: Map<string, Exportable[]>
+  ): string[] {
+    const usedIn: string[] = [];
+    files.forEach((exportables, file) => {
+      exportables
+        .filter(exportable => exportable instanceof ExportableInstance)
```
jafeltra, line 230: I might be missing something here, but why do you filter to only look at other Instance? Couldn't inline instances be used on profiles as well?

src/export/FSHExporter.ts
```diff
@@ -63,150 +75,190 @@ export class FSHExporter {
...
+  }
+
+  private invariantUsedIn(
+    invariant: ExportableInvariant,
+    files: Map<string, Exportable[]>
+  ): string[] {
+    const usedIn: string[] = [];
+    files.forEach((exportables, file) => {
+      exportables
+        .filter(exportable => exportable instanceof ExportableProfile)
```
jafeltra, line 253: I thought invariants could be used on profiles and extensions. Is there a reason you filter to only look profiles here? Is it because we only are concerned with where they are used in relation to profiles since that's the file we're adding them to?

src/export/FSHExporter.ts
```diff
@@ -63,150 +75,190 @@ export class FSHExporter {
...
+    const results: Exportable[] = [];
+    results.push(...this.fshPackage.aliases);
+    results.push(...this.fshPackage.profiles);
+    results.push(...this.fshPackage.extensions);
+    results.push(...this.fshPackage.codeSystems);
+    results.push(...this.fshPackage.valueSets);
+    results.push(...this.fshPackage.instances);
+    results.push(...this.fshPackage.invariants);
+    results.push(...this.fshPackage.mappings);
+    return new Map().set('resources.fsh', results);
```
jafeltra, line 124: When comparing the single-file output between this and the previous version of this branch (and master), there is no additional line between the end of the Alias list and the first profile. Are we okay with that? It feels like there might not be a clean way to get around, in which case, I think this difference is probably okay.

jafeltra (commented):
This looks good. I just had a few questions below. I also still think the `by-type` and `by-category` names are confusing, and I think I figured out a reason why I was confused by it on this PR. I think the "type" description is used when sorting to files when we're sorting based on profile relevance, but now we also use it as the table header in the `index.txt` file to indicate the FSH entity "type". So now I keep thinking of "by-type" as FSH entity sorting. I think keeping the word "type" to mean one thing might be useful. So maybe we can change `by-category` to `by-fsh-type` or `by-fsh-entity-type`, and I like Julian's idea of maybe `by-profile-grouping`.

src/exportable/Exportable.ts
ngfreiter: I'm sure there is a reason, but likely not a good one!

src/export/FSHExporter.ts
ngfreiter: I think it makes sense to leave extensions to themselves, since extensions are meant to stand more alone than something like an inline instance or an invariant. They get their own page in an IG.

src/export/FSHExporter.ts
ngfreiter: Yes they could, will fix

src/export/FSHExporter.ts
ngfreiter: Yes, since the extensions are still sorted into `extensions.fsh`, I wasn't sure it would be helpful to have the invariants follow those extensions around.

src/export/FSHExporter.ts
ngfreiter: Oh that's kinda icky though. I will figure out a way to add a line.

src/export/FSHExporter.ts
jafeltra: I buy it!

ngfreiter:
Ok @jafeltra, your comments are addressed in https://github.com/FHIR/GoFSH/pull/50/commits/e6209237e162993a5c2caab45d1bef7d98b8b1c5. @julianxcarter this commit also updates the `by-type` and `by-category` options per your suggestion. I tweaked them slightly so that `by-type` is now `group-by-profile` and `by-category` is now `group-by-fsh-type`. It just sounded slightly better to me having the word `group` out in front.

jafeltra (approved):
This looks good! And I like the updated names! (And nifty bonus fix for assigned instances on caret value rules)

src/export/FSHExporter.ts
jafeltra: I'm glad this wasn't actually icky.

julianxcarter (approved):
This still looks really good to me, nice work! @ngfreiter

PR #58, opened 2020-12-18
Integrate FSHing Trip

ngfreiter:
This adds a `-f, --fshing-trip` option to GoFSH. If this option is given, after GoFSH runs, SUSHI will run, and then a comparison will be generated showing the differences between the original input to GoFSH, and the final output of SUSHI. The differential is created using the `diff2html` library.

The user can optionally specify `use-local-sushi` after the `-f` flag (e.g. `gofsh -f use-local-sushi`), and this will force GoFSH to use whichever version of SUSHI the user has installed. Otherwise GoFSH will run SUSHI via `npx`, which will run the version of SUSHI that GoFSH depends on.

One issue I've ran into is that when you open the differential file, it takes a long time to load. Like a really long time. At least for me. This appears to be because of the `Diff2HtmlUI` function in `template.html`, so I'm not sure how much we can do about it, since this function comes from the library we are using. Some possibilities would be to live with the loading, use a different library, or put each differential in an individual file. I've also wrapped the diffs in collapsible elements, and this contributes some loading time, but it seems negligible compared to `Diff2HtmlUI`. But open to suggestions on how this can be improved.


src/utils/fshingTrip.ts
```diff
@@ -0,0 +1,82 @@
...
+    logger.info(
+      `Generated comparison file to ${path.join(process.cwd(), 'fshing-trip-comparison.html')}`
+    );
+  } catch (e) {
+    logger.error(`Comparison generation failed with error: ${e.message}`);
+  }
+}
+
+function getFilesMap(dir: string): Map<string, string> {
+  const files = getFilesRecursive(dir).filter(file => file.endsWith('.json'));
```
jafeltra, line 78: I'm not actually sure if this is a problem, but I realized that I added to this filter in a PR to skip over `output/expansions.json`, files in `temp`, and a couple others. I don't think it's necessarily required that the two filters match, but I wonder if would be nice to not include those extra files in the differences you display.

src/utils/fshingTrip.ts
```diff
@@ -0,0 +1,82 @@
...
+  });
+  try {
+    execSync(
+      `npx diff2html -i file -s side -F fshing-trip-comparison.html --hwt ${path.join(
+        __dirname,
+        'template.html'
+      )} -- ${diffFile.path}`
+    );
+    logger.info(
+      `Generated comparison file to ${path.join(process.cwd(), 'fshing-trip-comparison.html')}`
```
jafeltra, line 70: I didn't really expect the output to go to the root of the GoFSH project. I was somehow expecting it to go into the `gofsh` or output directory I specify when running GoFSH. Now that I stop and think about that, I'm not sure that it actually does make sense to be there if the idea is that the output directory is set up to be a FSH Tank. But I think I was looking for it with the rest of my output so I could go back and look at the files that were generated all together. That, and I wanted to delete all the remnants of my first GoFSH run in one clean `rm -rf gofsh` swoop. But that might not be a usual move for most users.

So, if we don't want to put the diff file in the output folder and we want to keep it here, how do you feel about putting this path into the `.gitignore`? We're already ignoring `gofsh` by default and it feels in a similar category of files.

jafeltra (commented):
This looks really good! I had two small questions/comments below.

I also didn't notice anything opening slowly. I ran the round trip on `fhir-mCODE-ig/fsh-generated/resources` and all the diffs open and collapse at a normal speed. So I think that's good!

jafeltra:
I don't feel too strongly about either of my comments, so I'm good with approving this PR either way the comments are resolved. So, assuming those two are resolved (in either direction), consider this my approval!

src/utils/fshingTrip.ts
cmoesel: I talked w/ Julia about this, and she's actually moving some of her filtering logic into `getFilesRecursive` - so this probably is not a concern anymore.

src/utils/fshingTrip.ts
cmoesel: I think I agree w/ @jafeltra on this.  I'd like to see the file in the output folder.  If the person wants to compare it against the actual generated files, or edit the files to account for unintended differences, I think it's easier to have it there -- and it shouldn't hurt SUSHI to have it there either.  It's a nice record of the process that people might want to keep around for a bit.

cmoesel (commented):
This is very cool -- and definitely provides some insight on areas for us to look at w/ GoFSH.  I've already commented on some of the things @jafeltra brought up.  There is one thing that threw me for a loop at first that I think would be good to address.  I think up until now, you could put the commandline flags _before_ the folder to process and it would work fine.  This is actually how I _usually_ do it.  So, naturally, I tried this:
```sh
ts-node src/app.ts -f ~/dev/HL7/fhir-mCODE-ig/fsh-generated/
```
I got back really weird errors and not at all the results I expected.  In fact, it crashed!  I think that the `-f` option slurped up the path to the folder to process and then GoFSH saw no path so decided to operate on the current folder.

I know that the `--help` technically says that the options go _after_ the path, but I can't imagine I'm the only one who might do it backwards.  I'd suggest one of the following:
* detect when the argument after `-f` is _not_ `use-local-sushi` and (somehow) do the right thing
* detect when the argument after `-f` is _not_ `use-local-sushi` and issue an error
* make `-f` a true flag (w/ no extra argument) and add a new flag to indicate that local SUSHI should be used

I'm not marking this as `Request Changes` since I'll be out until Jan 4.  So, if you address my comments, and it's confirmed by another team member, go ahead and merge.

src/utils/fshingTrip.ts
jafeltra: Agreed! With the changes I made, this should work fine, so I'm going to resolve this comment!

ngfreiter:
@jafeltra, https://github.com/FHIR/GoFSH/pull/58/commits/c27a46e2dd408f45ecfea53007717b156bd28713 addresses your comment by making the comparison file go to the output directory instead of the current working directory. And @cmoesel, this commit addresses your comment by adding a `-i` option, which can be used to indicate that the locally installed version of SUSHI should be used (if a comparison is being generated).

cmoesel:
This looks good.  I have one comment and one question.

Comment: I noticed that for a missing file, the diff looks like a file with the literal content `""`.  It seems to me that it would be better if the content was rendered as truly blank (rather than `""`).
image

I also noticed that there are a bunch of files that differ only by the order of properties (which, in JSON, does not matter).  This brings up a few thoughts:
1. Should we first do a deep equals compare and only show diffs of files that aren't deep equal? In some ways, this is good since it removes inconsequential differences -- but on the other hand, I could see it also being confusing to users, especially since the rendered diff (for _actually_ different files) might also highlight diffs that are inconsequential within the file.
2. Or should we update GoFSH and/or SUSHI to render the FSH/JSON in a way that will produce more consistent results?
3. Or should we just ignore it? (for now? forever?)

jafeltra (commented):
This looks good to me! My previous comment was resolved and I didn't notice anything else besides what Chris has mentioned as well.

ngfreiter:
> 1. Should we first do a deep equals compare and only show diffs of files that aren't deep equal? In some ways, this is good since it removes inconsequential differences -- but on the other hand, I could see it also being confusing to users, especially since the rendered diff (for _actually_ different files) might also highlight diffs that are inconsequential within the file.
> 2. Or should we update GoFSH and/or SUSHI to render the FSH/JSON in a way that will produce more consistent results?
> 3. Or should we just ignore it? (for now? forever?)

2. I'm not sure if this would help much. If the original input was made using a tool other than SUSHI, then the JSON could be structured in any way that we can't really control. And if the original input was made using SUSHI, well then I'm not sure why that person is using GoFSH.

As for the other options, I'm not totally sure. I'm not entirely sure what you mean by "but on the other hand, I could see it also being confusing to users, especially since the rendered diff (for actually different files) might also highlight diffs that are inconsequential within the file."?

cmoesel:
Re: 2 -- the one thing that I'm thinking of is following the element order as defined by the spec.  For example, here is the spec ordering of `coding`:
image

So... if SUSHI always honored spec order, then whenever it created a `Coding` JSON object, the properties would be in the order: `system`, `version`, `code`, `display`, `userSelected`.  And maybe (just maybe) the other tools follow spec order too.  At any rate, following spec order of properties seems less arbitrary than whatever we're doing now.

As far as confusing users goes, what I meant was that for a file that has the same content, but properties are in different order, we'll say it's the "same".  But if we detect a file w/ different content, that _diff_ will show not only the actually different content, but will _also_ show the non-meaningful differences.  It might confuse users that in this context, we're _not_ ignoring differences resulting only from order (since there are also non-order-related differences in the same file).

ngfreiter:
Oh I see what you mean. Although I'm definitely not opposed to us using spec order, that does seem like it would be a big lift to do specifically for just GoFSH, but if we thought it would be otherwise helpful that does seem like a nice feature for SUSHI.

> As far as confusing users goes, what I meant was that for a file that has the same content, but properties are in different order, we'll say it's the "same". But if we detect a file w/ different content, that diff will show not only the actually different content, but will also show the non-meaningful differences. It might confuse users that in this context, we're not ignoring differences resulting only from order (since there are also non-order-related differences in the same file).

I could see this maybe being confusing, but it also feels like it may be our best approach. I also wouldn't want to leave matching files out of the diff altogether (aka I would still want to indicate in some way that the two files do not differ, so no diff is shown, in case someone was looking to see if a specific file had differences). Are you ok with me doing it this way? Taking option number 1 from your original comment.

cmoesel:
> I also wouldn't want to leave matching files out of the diff altogether (aka I would still want to indicate in some way that the two files do not differ, so no diff is shown, in case someone was looking to see if a specific file had differences). Are you ok with me doing it this way?

Yeah, OK, as long as it's still easy to quickly see the files w/ actual differences (e.g., I shouldn't have to click on the file name to see if there are differences or not).  Ideally, files w/ differences should probably be grouped together so they don't get lost in the context of tens of files without differences.

ngfreiter:
Ok @cmoesel, the latest commits get rid of that extra `""` that was showing up in blank files, and also handle deep equality. In the end I just filtered out any files that were deeply equal, aka no diffs at all are shown for them. I tried making it so that the diff would just show both copies of the file side by side, to show that there was no diff, but this was an issue. Since the order used in the files themselves are not necessarily equal, I would've had to reorder the contents of the file to render the diff, and then the rendered diff would not match what is actually in the file. Hopefully what I'm saying makes sense, but the result was it just didn't seem like it would actually be helpful, and more likely to confuse.

ngfreiter:
Ok @cmoesel and @jafeltra, https://github.com/FHIR/GoFSH/pull/58/commits/a5eeea72c3a138f3ca7a491ff3ec247bf603f2a5 updates this PR so that if the input and output only differ on `snapshot` or a generated `text` element, no comparison is generated. When input and output differ based on something else, `snapshot` and generated `text` are removed and replaced with a comment indicating that it isn't really meaningful to include them in the comparison. This takes mCODE from having 200+ changed files to only about 50, so that's pretty nice. Let me know what you think.

cmoesel (changes requested):
This is definitely better!  I must be using a different version of mCODE than Nick because mine reduced diffs from over 200 files to just over 100 files.

A couple of observations:
* Generated `text` in bundle entry resources is still causing diffs
* The `date` property is causing diffs (since we exclude it if it looks like it is generated)
* Root elements in the differential (w/ only id and path) are causing diffs.

I did an experiment to see how things would change if I addressed those last two bullets (date and root element) and it reduced my number of diffs to 29!  I think we should do that too.  Here's my code:
```ts
    const generatedDate = inputFileJSON.date && !outputFileJSON.date;
    if (generatedDate) {
      delete inputFileJSON.date;
    }
    const emptyRootDifferential =
      inputFileJSON.differential?.element?.length &&
      isEqual(Object.keys(inputFileJSON.differential.element[0]), ['id', 'path']);
    if (emptyRootDifferential) {
      inputFileJSON.differential.element.splice(0, 1);
    }

// <snip -- then later on, to "restore" it>

    if (generatedDate) {
      inputFileJSON.date =
        'NOTE: Generated date is ignored by GoFSH, so this element is omitted for sake of comparison';
    }
    if (emptyRootDifferential) {
      inputFileJSON.differential.element[0] =
        'NOTE: Root differential is ignored by GoFSH, so this element is omitted for sake of comparison';
    }
```

The other thing is that there are some pretty predictable files that just aren't present in the roundtrip that we should probably be ignoring:
* `.index.json`
* `*.openapi.json`
* `ig-r4.json`

We also know that `ImplementationGuide-hl7.fhir.us.mcode.json` will never be produced (since we create a FSH-only project), but I think it's probably good that we make that obvious by including it in the diff.

ngfreiter:
@cmoesel, I made those suggested updates, and my mCODE is down to just 19 differences. Thanks for the suggested code!

jafeltra (approved):
The changes look good to me!

cmoesel (approved):
Right on!  Great feature!

PR #59, opened 2020-12-21
Additional files to skip while processing

### LINKED ISSUE #53: Files in /output cause GoFSH to hang
There are a set of json files that are generated by the IG Publisher as a matter or course and exist in the /output directory. The files include:
```
qa.json
canonicals.json
expansions.json
usage-stats.json
package.manifest.json
```
If someone wants GoFSH to convert publisher outputs to FSH files, GoFSH will hang. I believe `expansions.json `file is causing this.

My suggestion is to create a hardcoded "files to skip" list, and ignore these files.

jafeltra:
This PR fixes #53 with a few different changes:

- If GoFSH encounters an `expansions.json` file in the `output` directory, we skip processing it. When running GoFSH on the `fhir-mCODE-ig` root directory, GoFSH was hanging because the generated `expansions.json` file was so large. This file is a bundle of ValueSet resources, so I couldn't figure out a smarter way to skip past it without adding it to a list of specific files to skip.
- When running GoFSH on the `fhir-mCODE-ig` root directory, I also noticed there were files with `.escaped.json` file endings, which we recognized as JSON files and tried to parse. However, these are not valid JSON files, so an error would be logged. I decided to add these as files to skip processing.
- If GoFSH encounters a `temp` directory that is a child of the input path specified, we skip processing anything in the directory. If `temp` is used in the input path, we will still process files in it.
- GoFSH now skips any files in a `.git` directory.

These changes should allow GoFSH to run successfully on `fhir-mCODE-ig`. Resources are generated twice because they are in `output` and `fsh-generated`, but we will address that in a different story. It still takes a bit for GoFSH to finish (it takes me about 1 minute), but hopefully when we have additional logging, it will be clearer that things are happening.


src/utils/Processing.ts
```diff
@@ -99,7 +99,22 @@ export function loadExternalDependencies(
...
 function getLakeOfFHIR(inDir: string): LakeOfFHIR {
-  const files = getFilesRecursive(inDir).filter(file => file.endsWith('.json'));
+  // Don't process files in .git folder
+  // Don't process any files in a temp folder, unless the temp folder is used in inDir when running GoFSH
+  const tempDir = `${path.sep}temp${path.sep}`;
+  const gitDir = `${path.sep}.git${path.sep}`;
+  const tempInInDir = inDir.endsWith(`${path.sep}temp`) || inDir.includes(tempDir);
+  const relevantFiles = getFilesRecursive(inDir).filter(
+    file => (tempInInDir ? true : !file.includes(tempDir)) && !file.includes(gitDir)
+  );
```
cmoesel, line 109: I feel like it might be cleaner (and more efficient) to do this in `getFilesRecursive`.  I'm thinking something like:
```ts
function getFilesRecursive(dir: string): string[] {
  if (fs.statSync(dir).isDirectory()) {
    const ancestors = fs.readdirSync(dir, 'utf8').map(f => {
      // NEW CODE STARTS HERE
      if (f === 'temp' || f === '.git') {
        logger.debug(`Skipping ${f} folder: ${path.join(dir, f)}`);
        return [];
      }
      // NEW CODE STOPS HERE
      return getFilesRecursive(path.join(dir, f))
    });
    return [].concat(...ancestors);
  } else {
    return [dir];
  }
}
```
Since it only looks at folder names we recursively find, it still allows for `temp` in the base folder passed in -- but would filter out any nested `temp` paths underneath the base (which I think is OK).

src/utils/Processing.ts
```diff
@@ -99,7 +99,22 @@ export function loadExternalDependencies(
...
+
+  // IG Publisher creates .escaped.json files that are not valid JSON
+  const files = relevantFiles.filter(
+    file =>
+      file.endsWith('.json') &&
+      !file.endsWith('.escaped.json') &&
+      !IGNORED_RESOURCE_LIKE_FILES.some(path => file.endsWith(path))
+  );
```
cmoesel, line 117: We debug log files we skip in line 124-128 below (and in my suggested code above).  I'm wondering, if, for consistency, we should also debug log the `.escaped.json` and `output/expansions.json` files too...  If so, you might want to put those checks in the for-loop below instead (before you read the file in), so you can debug log and early-return when you find them...

cmoesel (commented):
I've suggested a few things that I think might improve upon the solution...  Let me know if you disagree.

src/utils/Processing.ts
cmoesel: This would also allow your tests to be a little more explicit by actively checking for the debug log statement instead of just testing the number of things returned.

cmoesel: Actually... you could just pop these into `getFilesRecursive` too (since they're only based on file name/path).

src/utils/Processing.ts
cmoesel: The other advantage of putting this in `getFilesRecursive` is that we'd have more consistency w/ Fishing Trip.  The comment you brought up [here](https://github.com/FHIR/GoFSH/pull/58#discussion_r546955617) would be moot if the filter were inside `getFilesRecursive`.

jafeltra: I like this idea. I was a little hesitant of making changes in `getFilesRecursive` because I wasn't sure if these were exceptions we wanted to make every time we need to get a list of files. But if you feel like they're common enough, and since it does solve the comment I made on Fishing Trip, I'm good with including them here.

src/utils/Processing.ts
jafeltra: I hadn't added a debug statement for these files because they felt like a slightly different category to me. These seemed like files we definitely didn't want to process, just like files that are other file types that we automatically filter out. However, maybe that's too much of a distinction. A debug statement won't hurt anyone. I can add them into the `getFilesRecursive`, as long as you agree that these are files we'd want to skip any time we try to get a list of files.

jafeltra:
@cmoesel both your comments, plus mine from [here](https://github.com/FHIR/GoFSH/pull/58#discussion_r546955617), should be resolved by 7ee2847.

jafeltra:
I added a few template files that @cmoesel pointed out where still logging errors when running GoFSH on `fhir-mCODE-ig` root directory because they were invalid JSON files. I didn't add a specific test for these files since they are ignored in the same way `output/expansions.json` is, so I didn't think it really tested any new functionality. But I can definitely add that test in if anyone would like it!

Running GoFSH on `fhir-mCODE-ig` should now not log any errors and should process at a reasonable speed. The only remaining issue is that we process all the resources twice since they are in `output` and `fsh-generated`, but I believe that's a separate task.

cmoesel (approved):
This is GREAT and makes a huge difference.  Thank you, @jafeltra!

ngfreiter (approved):
Looks good to me!

PR #60, opened 2020-12-23
Avoid redefining implicit slice elements

### LINKED ISSUE #56: Double Definition Error on Inherited Slices
I am getting the following error in the second step of round-tripping mCODE (FHIR to FSH back to FHIR):

```
error  Slice named conditionCode already exists on element Observation.component:alcoholAbuse.extension of CancerRelatedComorbidities
  File: C:\Users\mkramer\Documents\GitHub\mcode-gofsh\input\fsh\resources.fsh
  Line: 352
```

Here are the profiles generated by GoFSH for CancerRelatedComorbidities and its parent:

```
Profile:  ComorbiditiesParent
Parent: Observation
---snip---
* component.extension contains
    ComorbidConditionCode named conditionCode 0..* and
    ComorbidConditionReference named conditionReference 0..*
---snip---

Profile: CancerRelatedComorbidities
Parent: ComorbiditiesParent
---snip ---
* component contains
    alcoholAbuse 0..1 and ...
--- snip---
* component[alcoholAbuse].extension contains conditionCode 0..*
```

mint-thompson:
Fixes #56 and completes task [CIMPL-617](https://standardhealthrecord.atlassian.net/browse/CIMPL-617)

When slices are defined on a nested element, any slices on an element closer to the root will automatically contain the nested slices. These elements can also be inherited, and further slices near the root may be added. When processing elements that are slices, it is only necessary to create a ContainsRule for that element if the slice will not be automatically created.

Since the above is basically spaghetti, here's an example:
Parent profile states that element at path A.B contains a slice "banana". That element's id is A.B:banana. This would be represented with the contains rule "A.B contains banana".
Child profile states that element at path A contains a slice "apple". That element's id is A:apple.
The child profile needs a contains rule that says "A contains apple", as this is a completely new slice.
The child profile does not need a contains rule that says "A[apple].B contains banana", because this can already be inferred based on the slice defined on the parent profile.


ngfreiter (commented):
Looking good, I had a couple of minor comments.

ngfreiter (approved):
This looks good to me!

jafeltra (approved):
This looks good to me! I was able to use the test fixtures you added to reproduce the error on master but not on this branch. I couldn't think of any test cases that were missing, so all looks good to me!

PR #61, opened 2020-12-29
Description metadata can be multiline string

### LINKED ISSUE #52: Rendering new lines in descriptions
Any newline in a description field is rendering as \r\n.  Is it possible to actually get new lines instead?
Here is what it typically looks like:

```
Description: "Records an alteration in the most common DNA nucleotide sequence. The term variant can be used to describe an alteration that may be benign, pathogenic, or of unknown significance. The term variant is increasingly being used in place of the term mutation.\r\n\r\nConformance statement:\r\n\r\nObservation resources associated with an mCODE patient with Observation.code LOINC 69548-6 MUST conform to this profile. Beyond this requirement, a producer of resources SHOULD ensure that any resource instance associated with an mCODE patient that would reasonably be expected to conform to this profile SHOULD be published in this form."
```

mint-thompson:
Fixes #52 and completes task [CIMPL-613](https://standardhealthrecord.atlassian.net/browse/CIMPL-613)

If the description contains a newline character, treat it as a multiline string instead of a normal FSH string.

Note that this is only for descriptions: the value for an assignment or caret rule can also be a multiline string in FSH, but these are not being changed here.

### DIFF
--- a/src/exportable/common.ts
+++ b/src/exportable/common.ts
@@ -60,7 +60,13 @@ export function metadataToFSH(
     }
   }
   if (definition.description) {
-    resultLines.push(`Description: "${fshifyString(definition.description)}"`);
+    // Description can be a multiline string.
+    // If it contains newline characters, treat it as a multiline string.
+    if (definition.description.indexOf('\n') > -1) {
+      resultLines.push(`Description: """${definition.description}"""`);
+    } else {
+      resultLines.push(`Description: "${fshifyString(definition.description)}"`);
+    }
   }
   if (definition instanceof ExportableInvariant) {
     if (definition.severity) {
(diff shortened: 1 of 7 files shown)


ngfreiter (approved):
This looks good to me. The only thing I noticed was that on one of the descriptions generated for mCODE, the description generated by SUSHI seems to lose a few letters due to an extra space in the description string. However, this seems to be more a bug with SUSHI than an issue with this PR, so I've logged the bug [here](https://github.com/FHIR/sushi/issues/708), and think this PR looks good!

jafeltra (approved):
This looks good to me! I ran a round trip on US Core and everything worked as expected, and the bug fix from SUSHI worked as expected as well!

PR #62, opened 2020-12-30
Support new quantity syntax

mint-thompson:
Completes task [CIMPL-606](https://standardhealthrecord.atlassian.net/browse/CIMPL-606).

Updates to the FSH syntax for specifying a quantity now allow an optional unit string at the end. When combining rules on a quantity element, add in this string if it exists.

This is a FSH STU2 feature, and as such is targeting the `stu2-features` branch.


cmoesel (approved):
Yeah, that looks nice, real nice.

jafeltra (approved):
This looks good and makes sense to me!

Testing this out made me realize that I think I left out a piece when implementing this on the SUSHI side. Since the AssignmentRuleExtractor uses the `FshQuantity.toString` method, I think that method needs to support including the unit value. However, that shouldn't hold up this PR. (I did experiment with it since I had missed this when implementing that, so I can probably put up a quick PR for that.)

PR #64, opened 2021-01-11
Fix FSHing Trip build

ngfreiter:
When using `npm run build`, the `template.html` file that is necessary for FSHing Trip was not getting copied over into the `dist` folder, so running `node dist/app.js ../some-dir -f` would fail. This fixes that error by copying `template.html` into the `dist` folder.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -722,6 +722,22 @@
         }
       }
     },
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
It works!

PR #65, opened 2021-01-12
Use diff2html-cli

mint-thompson:
For some reason, using `diff2html-cli` instead of `diff2html` is necessary to make fshing-trip work correctly on my system. I don't understand it, but here we are.

Since there are no test cases for this, I ask that folks try running fshing-trip to make sure that this change doesn't cause it to stop working on your systems.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -114,7 +114,7 @@ async function app() {
   logger.info('Thank you for using goFSH.');
 
   if (program.fshingTrip) {
-    fshingTrip(inDir, outDir, program.installedSushi);
+    await fshingTrip(inDir, outDir, program.installedSushi);
   }
 
   process.exit(0);
--- a/src/app.ts
+++ b/src/app.ts
@@ -114,7 +114,7 @@ async function app() {
   logger.info('Thank you for using goFSH.');
 
   if (program.fshingTrip) {
-    await fshingTrip(inDir, outDir, program.installedSushi);
+    fshingTrip(inDir, outDir, program.installedSushi);
   }
 
   process.exit(0);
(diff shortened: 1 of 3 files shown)


mint-thompson:
update: i am an oaf and a fool and should remember to actually check for tests before assuming there are no tests. will fix tests now.

cmoesel:
Ugh.  So, this is not working on my system.  I get:
> error Comparison generation failed with error: Command failed: /bin/sh -c npx diff2html-cli -i file -s side -F fshing-trip-comparison.html --hwt /Users/cmoesel/dev/fhir/GoFSH/src/utils/template.html -- /var/folders/wn/kqf6wj156q55_w2w77wncjmw0000gn/T/2021012-9339-1ywdpo0.rk5hg.diff

cmoesel:
Based on npx documentation, I also tried `npx -p diff2html-cli diff2html ...`, but that also did not work.

cmoesel:
@mint-thompson -- So looking at [npx doc](https://www.npmjs.com/package/npx) again, I feel like `npx diff2html` _should_ have worked all along.  I may have led you astray saying to try `diff2html-cli`...  But it's still weird that `diff2html-cli` worked for you (but not for me), and vice-versa for `diff2html`.

`npx diff2html` should be looking for `diff2html` in your `node_modules/.bin` folder.  After running `npm install` on this project, do you _not_ have `node_modules/.bin/diff2html`?

mint-thompson:
@cmoesel `node_modules/.bin/diff2html` is present after running `npm install`, so that's not. It might be related to windows command prompt. `npx diff2html` consistently fails when run directly in the command prompt, and consistently works when run in bash. Other packages do not exhibit this pattern with npx, so maybe it's specific to diff2html?

Also, when running from a packed and globally installed version of goFSH, the result is the same. I'm going to keep digging around to see if I can come up with a consistent solution.

cmoesel:
Closing as it was eclipsed by #67.

PR #67, opened 2021-01-13
Use diff2html API directly instead of npx diff2html

cmoesel:
Using the API directly avoids a problem w/ using npx diff2html on Windows systems.


cmoesel:
Note: If this works, then #65 should be closed.

ngfreiter (approved):
This works for me. Since I still have my windows machine, I tested it there as well, and it worked for me there too.

mint-thompson (approved):
Works on my windows machine as well. Hooray!!

PR #69, opened 2021-01-22
Print a progress counter as GoFSH processes FSH

### LINKED ISSUE #54: Output a heartbeat or progress counter while running
There is no sign of life while GoFSH is executing, after the message `info  Using output directory: gofsh` and before the message `info  Exported 33 Profiles.`  Usually the pause is only a few seconds, so it isn't too disconcerting, but with large IGs, it actually takes some time to generate all the FSH. The user can be left wondering if anything is really happening.
If possible, can you add some sort of output during this period, to let the user know what's happening, and that the process hasn't died?

julianxcarter:
Fixes #54, adding a progress counter to the `FHIRProcessor` that indicates to the user that GoFSH is in fact running as it converts larger IGs to FSH projects.

### DIFF
--- a/src/processor/FHIRProcessor.ts
+++ b/src/processor/FHIRProcessor.ts
@@ -71,13 +71,16 @@ export class FHIRProcessor {
   // Outputs a counter for each resource type being processed
   outputCount(fhirArr: WildFHIR[], index: number, type: string) {
     if (index == fhirArr.length - 1) {
+      // We want the logger to overwite the previous count
       process.stdout.write('\x1b[A\r');
       let finalMessage = `Processed ${fhirArr.length} ${type}`;
       if (fhirArr.length === 1) {
         logger.info(finalMessage.padEnd(45));
+        return;
       } else {
         finalMessage = finalMessage.concat('s');
         logger.info(finalMessage.padEnd(45));
+        return;
       }
     } else if ((index + 1) % 5 == 0) {
       // We want the logger to overwite the previous count
--- a/src/processor/FHIRProcessor.ts
+++ b/src/processor/FHIRProcessor.ts
@@ -6,7 +6,8 @@ import {
   ValueSetProcessor,
   ConfigurationProcessor,
   Package,
-  LakeOfFHIR
+  LakeOfFHIR,
+  WildFHIR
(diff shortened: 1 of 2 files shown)


src/processor/FHIRProcessor.ts
```diff
@@ -67,13 +68,36 @@ export class FHIRProcessor {
...
+
   process(config: ExportableConfiguration): Package {
     const resources = new Package();
     const igForConfig =
       this.lake.getAllImplementationGuides().find(doc => doc.path === this.igPath) ??
       this.lake.getAllImplementationGuides()[0];
     resources.add(config);
-    this.lake.getAllStructureDefinitions().forEach(wild => {
+    const structureDefs = this.lake.getAllStructureDefinitions();
+    logger.info('Processing StructureDefinitions...');
```
ngfreiter, line 99: This applies to all the messages like this, but since these messages are logged before we know that there are `StructureDefinitions`, we get something that looks like this: 
image
This is a case where my input had 1 SD and no other resources, but we still get the messages saying that we are `Processing CodeSystems...` and so on. I think we just want to only log these messages when the array of relevant resources is not empty.

ngfreiter (changes requested):
Noticed one small thing, but looking good besides that.

ngfreiter (approved):
Looks good!

jafeltra (commented):
This looks so cool! I just had one question. I noticed that when I run gofsh on some input that has a lot of instances, the Instance count doesn't start for a while after the `Processing Instances...` message is displayed, so there isn't much feedback about what is going on. I'm not sure if maybe this is just something built into how the InstanceProcessor works, but I was curious if you had noticed it as well. If it's built into how the InstanceProcessor is working, it's probably not a big enough deal to worry about because this is a huge improvement on feedback during processing!

jafeltra (approved):
Okay never mind! I added some extra logs because I was curious and it looks like the first few Instances I have must be really big (they're capability statements and a large code system, so that explains a lot). They just take a long time to process, so since they happen to be the first 5, we don't get to the first 5 threshold to get a new log. I don't think we want to lower that threshold, so I take it all back and I'm approving! 😀

PR #70, opened 2021-01-22
AssignmentRule quantity must have UCUM system

### LINKED ISSUE #63: GoFSH creates Quantities that SUSHI can't convert to single rules (non-UCUM systems)
In FSH, you can do a quantity assignment like such:
```
* valueQuantity = 55 'mm'
```
This assumes that `mm` is a valid UCUM unit.  The `FshQuantity` `toString` function constructs the stringified version of `Quantity` using this format.  In SUSHI, this is fine because the only time the `FshQuantity` is ever constructed is when the units are UCUM.

In GoFSH, however, we can construct `FshQuantity` with non-UCUM unit systems.  In this case, when the rules are created, the `FshQuantity.toString` isn't quite right since the unit is not UCUM.  But there isn't any way to make it right because there is no single string representation of a quantity with non-UCUM units.  So we are at an impasse.

In a future version of FSH, we may support a Quantity representation using non-UCUM units like this: `55 http://foo.org#mm` -- in which case we have a good way to solve this problem.  For FSH 1.0, however, we will need to have GoFSH break up Quantities into multiple statements when the unit system is not UCUM.

(This issue is also noted in FHIR/sushi#716).

mint-thompson:
Fixes #63 and completes task [CIMPL-641](https://standardhealthrecord.atlassian.net/browse/CIMPL-641).

When extracting assignment rules for a quantity or ratio (which contains quantities), construct a FSHQuantity instance only if the system is the url for UCUM. This is because SUSHI currently assumes that quantities and ratios use this system. If the system is not the UCUM url, create separate rules for the code and the value.


src/extractor/AssignmentRuleExtractor.ts
```diff
@@ -58,21 +58,30 @@ export class AssignmentRuleExtractor {
             `${matchingKey}.coding[0].system`,
             `${matchingKey}.coding[0].display`
           );
-          return assignmentRule;
+          return [assignmentRule];
         } else if (matchingKey.endsWith('Quantity') && isQuantity(matchingValue)) {
```
ngfreiter, line 62: I think we may need to update our idea of what a `Quantity` is based on these changes. The issue I'm noticing is that if there is some FSH defined like this:
```
Extension: NonUCUMQuant
* valueQuantity = http://some-url.org#foo "these are other units"
* valueQuantity.value = 12
```
And we run that through SUSHI, the resulting StructureDefinition has `code`, `system`, and `unit` set on `Extension.valueQuantity` via a `patternQuantity`, and `value` set on `Extension.valueQuantity.value` via a `patternDecimal`. Since this code is thinking that a `Quantity` must have `code` and `value`, the `patternQuantity` is not processed as a `Quantity` here, and ends up getting set with a caret rule: 
```
Extension: NonUCUMQuant
Id: NonUCUMQuant
* valueQuantity ^patternQuantity = $some-url#foo "these are other units"
* valueQuantity.value = 12
```
which doesn't seem right. Note that if you update `isQuantity` to include Quantities that don't have a `value`, I think there will need to be some changes further down, in places where it is assumed `value` exists.

ngfreiter (changes requested):
I noticed that this new approach can cause some inconsistencies when the assignments are spread over multiple elements, and that is how we would expect the JSON to look if someone was using SUSHI.

src/extractor/AssignmentRuleExtractor.ts
mint-thompson: That's an excellent point. FSH allows for assigning to a `Quantity` without a value, so goFSH shouldn't be surprised when a patternQuantity doesn't have a value. I'll make this more flexible.

jafeltra (approved):
This looks good to me!

ngfreiter (approved):
Looking good to me!

PR #71, opened 2021-01-25
Ensure Quantity.unit setting rules are printed after Quantity setting rules

julianxcarter:
Addresses [CIMPL-639](https://standardhealthrecord.atlassian.net/browse/CIMPL-639?atlOrigin=eyJpIjoiZDhhNTM3Zjg2YzY3NDAyYmJmNTFlYzQyZTcxMjMxODciLCJwIjoiaiJ9), ensuring that Quantity.unit setting rules are exported _after_ Quantity setting rules to prevent unexpected overwrites of the unit field. A `switchQuantityRules` function has been added that's called during export by `ExportableInstance`, `ExportableProfile`, and `ExportableExtension`. This function simply iterates through the rules on these SDs and moves whole Quantity-setting rules to be just before Quantity.unit-setting rules.


ngfreiter (commented):
I'm wondering again why this happens after the optimizers. The optimizers happen in `getResources`, which is called before `writeFSH`, which calls the various `toFSH` methods via the `FSHExporter`. So with this approach, my understanding is that the FSH definitions will still be incorrect even after going through the optimizers, and are then only fixed in the `toFSH` methods. This seems like a break from our idea that FSH definitions should be at least valid by optimization time, and the optimizers only make them more efficient. To me it feels like if we are going to add an extra step that corrects the rules, it is making more sense in my head that it would happen before the optimizers, and not after.

Let me know if this makes sense, or if you think I'm way off here.

src/processor/StructureDefinitionProcessor.ts
```diff
@@ -21,6 +21,7 @@ import {
 } from '../extractor';
 import { ProcessableElementDefinition } from '.';
 import { getAncestorSliceDefinition } from '../utils';
+import { switchQuantityRules } from './common';
```
ngfreiter, line 24: Minor thing, but if you add:
```javascript
export * from './common';
```
to `src/processor/index.ts`, then on line 22 here, you can do:
```javascript
import { ProcessableElementDefinition, switchQuantityRules } from '.';
```

src/processor/common.ts
```diff
@@ -0,0 +1,16 @@
+import { ExportableAssignmentRule, ExportableSdRule } from '../exportable';
+
+// Places general Quantity-setting rules ahead of Quantity.unit setting rules
+export function switchQuantityRules(rules: ExportableSdRule[]): void {
+  const seenRules = new Map();
+  rules.forEach((rule, index) => {
+    if (!seenRules.has(rule.path) && rule instanceof ExportableAssignmentRule)
+      seenRules.set(rule.path, index);
+    const unitRulePath = rule.path.concat('.unit');
```
ngfreiter, line 9: I think it is fair to assume that `Quantity.unit` is the only usage of `unit` in FHIR right? I think I'm good with assuming that.

ngfreiter: Actually now that I think about it, I'm not worried about this at all, because worst case we would just be re-ordering the rules in a way that I can't picture being harmful.

src/processor/common.ts
```diff
@@ -0,0 +1,16 @@
+import { ExportableAssignmentRule, ExportableSdRule } from '../exportable';
+
+// Places general Quantity-setting rules ahead of Quantity.unit setting rules
+export function switchQuantityRules(rules: ExportableSdRule[]): void {
+  const seenRules = new Map();
+  rules.forEach((rule, index) => {
+    if (!seenRules.has(rule.path) && rule instanceof ExportableAssignmentRule)
```
ngfreiter, line 7: I think we would also want this to work on `ExportableCaretValueRule`s. It is possible that this situation could arise there as well. It's a bit of a pain, since on the `CaretValueRule` we care about the `caretPath` and not the `path`, but there is a `getRulePath` function [here](https://github.com/FHIR/GoFSH/blob/ad7c7f8b4b279a1ff1ecd4ee19cc2d877aef8bab/src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts#L117) that you could probably use to make that easier.

ngfreiter (commented):
Thanks for the refactor! I think this follows our optimizer rules nicely. I noticed a couple things that I think should be updated. Also I think it would be good to add just one test to each processor to make sure that the rule switching is happening as we'd expect.

src/processor/common.ts
julianxcarter: I did a search of [the FHIR datatypes page](https://www.hl7.org/fhir/datatypes.html) and that _seems_ to be the case, but as you said there shouldn't be any adverse effects if that's not the case.

src/processor/common.ts
julianxcarter: You're right about that! I'll add support for `ExportableCaretValueRule`s as well. Would the `caretPath` matter for what we're trying to accomplish though? I'm thinking that only the regular path would matter here even in the case of an `ExportableCaretValueRule`, since that's the portion that specifies which element is being set. I may be wrong here though!

julianxcarter:
So I made some bold moves! When Instances are processed, each individual property of a `Quantity` gets its own assignment rule. For example, if we process this element within an instance...
```
  "valueQuantity": {
    "unit": "C",
    "value": 36.5,
    "system": "http://unitsofmeasure.org",
    "code": "Cel"
  }
```
In lieu of an Optimizer, this will yield the following assignment rules:
```
* valueQuantity.unit = "C"
* valueQuantity.value = 36.5
* valueQuantity.system = "http://unitsofmeasure.org"
* valueQuantity.code = Cel
```

The above is actually valid FHIR, as an overwrite would only happen when the `CombineCodingAndQuantityValuesOptimizer` runs and merges the  latter three rules into one, giving us this: 
```
* valueQuantity.unit = "C"
* valueQuantity= 36.5 'C'
```
In this case, the first rule setting the `unit` property _will_ be overwritten. With this in mind, I actually thought it would be better to update the `CombineCodingAndQuantityValuesOptimizer` to include each property of the Quantity into a single assignment rule on Instances. Given the [changes that we're making on the sushi side to broaden support for non-UCUM systems on quantities](https://github.com/FHIR/sushi/pull/740), I thought that this would be okay to do.

src/processor/common.ts
julianxcarter: I was wrong, and took your suggestion haha. Thank you!

jafeltra (commented):
This makes sense to me. The one thing I was going to suggest was having something that tested switching rules on Instances since `switchQuantityRulesOnInstances` is a separate function for that case. But it looks like from your note in the code that we are expecting that will go away soon. Is that the case? And did you already have something that tested a similar case when you were using the optimizer approach that will get added back when we implement that approach?

src/processor/common.ts
```diff
@@ -0,0 +1,49 @@
...
+export function switchQuantityRulesOnInstances(rules: ExportableSdRule[]): void {
+  let unitIndex: number;
+  let lastSiblingIndex: number;
+  let basePath: string;
+  let siblingPaths: string[];
+  rules.forEach((rule, index) => {
+    if (rule.path.endsWith('.unit')) {
+      unitIndex = index;
+      basePath = rule.path.replace(/\.unit$/, '');
+      siblingPaths = [`${basePath}.system`, `${basePath}.code`, `${basePath}.value`];
```
ngfreiter, line 38: So I think on an Instance, the rules would always come as individual rules, like you are accounting for here. But, I think that the rules could also come as individual rules on a StructureDefinition (if `patternQuantity` is not used, and instead the constraints are individually applied to each element). That makes me think that the processing required for a StructureDefinition is basically a superset of the processing required for an Instance, so I think we could just have one single function, that handles both cases, individual and combined quantity rules. When applied to an instance, the parts of the code handling combined rules would never be hit, but that is fine I think. Does this make sense?

julianxcarter: This makes perfect sense, I'll combine these into one function. Thank you for thinking of this!

ngfreiter (approved):
This looks good to me!

jafeltra (approved):
Looks good to me!

PR #72, opened 2021-01-26
Use unique names on Instances, simplify if possible.

jafeltra:
This PR makes it so that all Instances will have a unique name, while preserving the Instance's id. I was able to do this using a true optimizer approach, so I haven't implemented any new "corrector" concept and didn't break our current definition of "optimizing".

In order to do this, I had the Instance `name` include the InstanceOf information. The `id` is kept the same. Once we have all instances generated, we can use an optimizer to simplify the `name` if we don't need the additional `InstanceOf` information and update it back to the `id`. Because we also have an optimizer to update the `InstanceOf` URLs, I thought it would look nicer (and make more sense) to use the alias if it is used on the `InstanceOf`, so that simplification is included in the optimizer as well.


mint-thompson (approved):
Looks good, thank you!

julianxcarter (approved):
This is a very clever solution, I'm approving! Nice work @jafeltra

PR #73, opened 2021-01-26
Update to SUSHI v1.1.0

jafeltra:
This PR updates SUSHI to the latest version, v1.1.0. I'm not entirely sure why there are so many changes in `package-lock.json`, but I got the same set of changes when I did `npm upgrade fsh-sushi` and when I manually changed the version to `^1.1.0` and installed. My best guess is something with the `antlr4` dependency issue and `npm-shrinkwrap.json`, but that's really just a guess because those commits are introduced in this version upgrade.

This PR also sets the `applyExtensionMetadataToRoot` flag to `false` in all generated `sushi-config.yaml` files.


mint-thompson (approved):
Hooray for updated SUSHI! This means that goFSH can start using some of the new FSH features, correct?

jafeltra:
> Hooray for updated SUSHI! This means that goFSH can start using some of the new FSH features, correct?

I think so! I think the plan was to release a 1.0.0 that's compliant with FSH STU1, and I think we can use cool new features!

ngfreiter (approved):
This looks good to me!

cmoesel:
> > Hooray for updated SUSHI! This means that goFSH can start using some of the new FSH features, correct?
>
> I think so! I think the plan was to release a 1.0.0 that's compliant with FSH STU1, and I think we can use cool new features!

Yeah. GoFSH 1.0 should be SUSHI 1.0 compliant.  After that we can talk about how and when we leverage new SUSHI features in GoFSH.  We might want to let GoFSH settle a little first and work out any bugs that get reported so we're sure we have a stable FSH 1.0-compliant GoFSH before making it work w/ post 1.0 FSH features.

PR #74, opened 2021-01-27
Add fhirToFsh to create a GoFSH API

ngfreiter:
This adds a `fhirToFsh` function that can be called to transform FHIR definitions to FSH.

Inputs: 
* `input` - An array of JSON or strings that can be parsed into JSON. Each element of the array should be a FHIR resource. If an input is not FHIR JSON, or cannot be parsed to FHIR JSON, an error is logged.
* `options` - An object with several parameters.
  * `dependencies` - A list of dependencies in the form `id@version` or `id#version`.
  * `logLevel` - A level to set the GoFSH logger at, can be one of `silly`, `debug`, `verbose`, `http`, `info`, `warn`, `error`, `silent`.
  * `style` - A style for the output. Can be `string` or `map`. When it is `string`, the output FSH will be a single string, containing all definitions. When it is `map`, the output will be an object with `Maps` for each FSH entity, `profiles`, `valueSets`, and so on. Each `Map` contains all the definitions of the corresponding FSH entity type, where the key is the name of the entity, and the value is the actual FSH content. Default is `string`.
 
Outputs:
* `fsh`: The FSH content, styled according to the `exportStyle` param. 
* `configuration`: A JSON representation of the YAML that GoFSH would write to `sushi-config.yaml` in a command line context.
* `errors`: An array containing the error messages logged during the run.
* `warnings`: An array containing the warning messages logged during the run.

The function is exported as a part of the `gofshClient` object, which GoFSH now exports. So in another package which has GoFSH as a dependency, you could import and use this function like this:
```
import { gofshClient } from "gofsh";
gofshClient.fhirToFsh([ /* some FHIR */ ]);
```
Here is a simple script I was using to test:
```javascript
import { gofshClient } from "gofsh";

async function run() {
  const fhir = [
    {
      resourceType: "StructureDefinition",
      id: "MyNormalIntExtension",
      url: "http://example.org/StructureDefinition/MyNormalIntExtension",
      version: "0.2.0",
      name: "MyNormalIntExtension",
      status: "active",
      fhirVersion: "4.5.0",
      mapping: [
        {
          identity: "rim",
          uri: "http://hl7.org/v3",
          name: "RIM Mapping",
        },
      ],
      kind: "complex-type",
      abstract: false,
      context: [
        {
          type: "element",
          expression: "Element",
        },
      ],
      type: "Extension",
      baseDefinition: "http://hl7.org/fhir/StructureDefinition/Extension",
      derivation: "constraint",
      differential: {
        element: [
          {
            id: "Extension.extension",
            path: "Extension.extension",
            max: "0",
          },
          {
            id: "Extension.url",
            path: "Extension.url",
            fixedUri:
              "http://example.org/StructureDefinition/MyNormalIntExtension",
          },
          {
            id: "Extension.value[x]",
            path: "Extension.value[x]",
            slicing: {
              discriminator: [
                {
                  type: "type",
                  path: "$this",
                },
              ],
              ordered: false,
              rules: "open",
            },
            min: 1,
            type: [
              {
                code: "integer",
              },
            ],
          },
          {
            id: "Extension.valueInteger",
            path: "Extension.valueInteger",
            min: 1,
            max: "1",
            patternInteger: 12345,
            mustSupport: true,
          },
        ],
      },
    },
  ];
  const output = await gofshClient.fhirToFsh(fhir, {
    dependencies: ["hl7.fhir.us.mcode#1.0.0"],
    style: "map",
  });
  console.log(output);
}

run();
```
and the corresponding `package.json`:
```json
{
  "name": "gofsh-api-user",
  "version": "1.0.0",
  "description": "",
  "main": "api.js",
  "scripts": {
    "test": "echo \"Error: no test specified\" && exit 1"
  },
  "author": "",
  "license": "ISC",
  "dependencies": {
    "fsh-sushi": "^1.1.0",
    "gofsh": "file:../gofsh"
  },
  "type": "module"
}
```


jafeltra (approved):
This looks incredibly excellent! I didn't find anything and it all works as I'd expect!

The one thing I did notice is that we might want some documentation around how to use `FhirToFsh`. I also realized we don't have much for documentation around `FshToFhir` either. So I don't think this should hold up this PR at all, but maybe in the future we could add documentation for how to use both of these. I feel like your PR description with the various `input` options and `output` is probably almost sufficient.

cmoesel:
I see you already have two reviewers, so I'm just commenting on the PR description with a couple of questions:

> `dependencies` - A list of dependencies in the form `id@version` or `id#version`
Why do we allow two variants that are so similar?  Shouldn't we just pick one?

> `fsh`: The FSH content, styled according to the `exportStyle` param.
In the input description for `options`, you call it just `style`.  The inconsistency probably isn't a big deal, but I mention it in case we lift this description to re-use it as user-facing documentation.

ngfreiter:
> I see you already have two reviewers, so I'm just commenting on the PR description with a couple of questions:
> 
> > `dependencies` - A list of dependencies in the form `id@version` or `id#version`
> 
> Why do we allow two variants that are so similar? Shouldn't we just pick one?

I allowed two versions because one version is the one used on the GoFSH command line (`@`) and one version is the one that I would expect people are used to for FHIR (`#`). I figured it wouldn't hurt. But if I were to restrict it to one, which do you think is best?

> > `fsh`: The FSH content, styled according to the `exportStyle` param.
> 
> In the input description for `options`, you call it just `style`. The inconsistency probably isn't a big deal, but I mention it in case we lift this description to re-use it as user-facing documentation.

Ah yeah, `exportStyle` is the name of the type. My bad.

jafeltra:
Actually, I just realized one other thing. If you specify a `style` that isn't a valid option, we don't output any fsh. Do you think we should default it to `string` if you don't supply a valid option, similar to if you supply no option?

I'm not actually sure if that's something we want to do, so I haven't revoked my approval yet.

cmoesel:
> I allowed two versions because one version is the one used on the GoFSH command line (`@`) and one version is the one that I would expect people are used to for FHIR (`#`). I figured it wouldn't hurt. But if I were to restrict it to one, which do you think is best?

I think we should just use `#` -- because that is what is used by FHIR (in ig.ini and maybe elsewhere) and that is also what is used in FSH Online configuration.  And for consistency, I think I'd also update the GoFSH CLI to take `#` too (although I'd keep _undocumented_ support for `@` in the CLI for a little while so we don't mess up current users).

jafeltra:
> And for consistency, I think I'd also update the GoFSH CLI to take `#` too (although I'd keep _undocumented_ support for `@` in the CLI for a little while so we don't mess up current users)

I think we originally intended to use `#` in the `--dependency` option, but `#` is a special character in bash and zsh, so we had to either escape it or put quotation marks around the dependency. We can always do either of those options, but I think that's how we arrived at the odd `@`.

cmoesel:
Oh... right.  Well then.  OK.  That being the case, then I think maybe it's OK to accept both in the API too.

ngfreiter:
> The one thing I did notice is that we might want some documentation around how to use `FhirToFsh`. I also realized we don't have much for documentation around `FshToFhir` either. So I don't think this should hold up this PR at all, but maybe in the future we could add documentation for how to use both of these. I feel like your PR description with the various `input` options and `output` is probably almost sufficient.

I've added a task to document both APIs on FSHSchool, and I can take that next sprint: https://standardhealthrecord.atlassian.net/browse/CIMPL-657.

ngfreiter:
> Actually, I just realized one other thing. If you specify a `style` that isn't a valid option, we don't output any fsh. Do you think we should default it to `string` if you don't supply a valid option, similar to if you supply no option?

Addressed in https://github.com/FHIR/GoFSH/pull/74/commits/e797f2f41fea31270487cb15b8ac2a76e3b92784.

jafeltra (approved):
Thanks! I'm reapproving!

jafeltra (approved):

julianxcarter (approved):
I tested this out and this works really well, there's not really anything that I'd like to see changed. Great job @ngfreiter!

PR #76, opened 2021-02-03
Url-setting CaretValueRules now included on SDs, CSs, and VSs

### LINKED ISSUE #68: GoFSH always ignores SD/VS/CS url
GoFSH currently assumes that the `url` in a StructureDefinition, ValueSet, or CodeSystem will match whatever SUSHI would construct from the canonical (e.g., `{canonical}/{type}/{id}`).  This, however, is not a guarantee, so GoFSH should check the URL in the JSON against what SUSHI would create -- and if it's different, generate a caret value rule.

See: https://github.com/FHIR/GoFSH/blob/e4821903d8cebb3edf61ee3c472b88debb47c77d/src/extractor/CaretValueRuleExtractor.ts#L147-L173

julianxcarter:
Fixes #68 and addresses [CIMPL-656](https://standardhealthrecord.atlassian.net/browse/CIMPL-656), including CaretValueRules setting the `url` property of StructureDefinitions, ValueSets, and CodeSystems when the pre-existing url is different than the url that sushi would generate automatically.

### DIFF
--- a/src/extractor/CaretValueRuleExtractor.ts
+++ b/src/extractor/CaretValueRuleExtractor.ts
@@ -111,10 +111,8 @@ export class CaretValueRuleExtractor {
     const flatParent = getPathValuePairs(parent);
     Object.keys(flatSD).forEach(key => {
       if (flatParent[key] == null || !isEqual(flatSD[key], flatParent[key])) {
-        if (key === 'url') {
-          const existingUrl = input.url;
-          const generatedUrl = `${config.canonical}/StructureDefinition/${input.id}`;
-          if (existingUrl === generatedUrl) return;
+        if (key === 'url' && this.isStandardURL('StructureDefinition', config, input)) {
+          return;
         }
         const caretValueRule = new ExportableCaretValueRule('');
         caretValueRule.caretPath = key;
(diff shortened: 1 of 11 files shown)


cmoesel (dismissed):
Looks good to me.  I pushed up a small commit to test the url value too, as we want to ensure those come across correctly as well.

src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -110,6 +111,11 @@ export class CaretValueRuleExtractor {
     const flatParent = getPathValuePairs(parent);
     Object.keys(flatSD).forEach(key => {
       if (flatParent[key] == null || !isEqual(flatSD[key], flatParent[key])) {
+        if (key === 'url') {
+          const existingUrl = input.url;
+          const generatedUrl = `${config.canonical}/StructureDefinition/${input.id}`;
+          if (existingUrl === generatedUrl) return;
+        }
```
ngfreiter, line 118: Since this check is repeated on 145-149, it may make sense to extract it out to a little helper function. I also think we could just combine everything into one `if` statement, like this:
```
if (key === 'url' && input.url === `${config.canonical}/${resourceType}/${input.id}` {
  return;
}
```
I think that will work the same, right?

julianxcarter: I believe it'll work the same, I'll condense this down in the next commit!

cmoesel (approved):
I like it even better now.  Thanks @julianxcarter and @ngfreiter.

ngfreiter (approved):
Looks good!

Issue #75, opened 2021-02-02
Missing SD conversions


jvitrifork:
GoFSH'ing on https://github.com/hl7dk/dk-core/tree/master/input/resources results in only 3 converted resources: one instance, a codesystem and a valueset. How far down in the backlog of GoFSH would it be to successfully convert more of the resources within such an IG (that contains a single resource SD and an couple of datatype SD's)?

cmoesel:
Hi @jvitrifork -- thanks for reporting this.  Looking at the repository link you provided, I think the main issue is that GoFSH currently only reads in JSON files; but `dk-core` contains many XML files as well.  We're currently working on a GoFSH 1.0 release that would be feature-complete for JSON definitions; but hadn't yet scheduled support for XML.  That said, we can look into it to see how difficult it would (or wouldn't) be to slip in XML support.

One possible workaround is to run the IG Publisher on your project first and then point GoFSH at the generated `output` folder, which I _think_ should contain JSON definitions for everything.  But if you're not in a big time crunch, you might want to wait for GoFSH 1.0.

jvitrifork:
Thx for the advise @cmoesel  - sure, the output folder contains the files in json format. Trace is the following, which implies an actual bug I believe:

```
output % gofsh . -l debug
info  Starting goFSH v0.4.0 (implements FHIR Shorthand specification v0.13.x)
info  Using input directory: .
info  Found 32 JSON files.
debug Skipping non-FHIR JSON file: canonicals.json
debug Skipping non-FHIR JSON file: package.manifest.json
debug Skipping non-FHIR JSON file: qa.json
debug Skipping non-FHIR JSON file: usage-stats.json
info  Checking local cache for hl7.fhir.r4.core#4.0.1...
info  Found hl7.fhir.r4.core#4.0.1 in local cache.
info  Loaded package hl7.fhir.r4.core#4.0.1
info  Using output directory: gofsh
(node:84693) Warning: Accessing non-existent property 'INVALID_ALT_NUMBER' of module exports inside circular dependency
(Use `node --trace-warnings ...` to show where the warning was created)
(node:84693) Warning: Accessing non-existent property 'INVALID_ALT_NUMBER' of module exports inside circular dependency
warn  InstanceOf definition not found for episodeOfCare-example-wrong. The ResourceType of the instance will be used as a base.
warn  InstanceOf definition not found for episodeOfCare-example. The ResourceType of the instance will be used as a base.
[...]

PR #77, opened 2021-02-04
Fix null dereference

cmoesel:
This fixes and issue that was reported as part of #75 (it's not the original issue, but an issue that came up when attempting the provided workaround).  Basically GoFSH was crashing, and it turns out it was crashing because we tried to access a property on a variable that was null.

If you really want to reproduce:
* Clone [dk-core](https://github.com/hl7dk/dk-core)
* Run the IG publisher on it (`java -jar publisher.jar -ig ig.ini`)
* Run GoFSH on the resulting `output` directory

If you don't want to reproduce, I think that's probably OK -- because no matter how you look at it, `def?.name.match` is definitely a bug.


mint-thompson (approved):

ngfreiter (approved):
👍 👍

PR #80, opened 2021-02-04
Fix npm pack with FSHing Trip

ngfreiter:
FSHing trip was not working correctly when `npm pack`ed, because the `template.html` file was not packaged. This should fix that.


cmoesel (approved):
Yep!

jafeltra (approved):
Truly wonderful.

PR #81, opened 2021-02-05
Empty output directory before use

### LINKED ISSUE #78: Running GoFSH after FSHing trip causes duplicate FSH definitions
If you run GoFSH FSHing Trip (e.g., `gofsh . -f`), it will run SUSHI on the GoFSH output, creating JSON files in `gofsh/input/fsh-generated/resources`.  Then if you run GoFSH again (with or without FSHing Trip), GoFSH will read both the original JSON definitions and the new SUSHI-generated definitions, causing duplicate definitions in the generated FSH files.

I think we need to do one of the following:
* delete the output folder before generating more output (like SUSHI does w/ fsh-generated)
* warn the user if the output folder already exists (but still continue)
* prompt the user to ask if the output folder should be deleted

mint-thompson:
Fixes #78 .

This matches the behavior of SUSHI. Additionally, this will help prevent errors resulting when running a fshing-trip with output in a directory that is inside a directory used as input to a subsequent goFSH run.


cmoesel (changes requested):
Unfortunately, this doesn't actually fix the problem.  Try this:
```
$ ts-node src/app.ts ~/my-fhir-project -o ~/my-fhir-project/gofsh -f
```
and note the GoFSH result counts.  Then so this:
```
$ ts-node src/app.ts ~/my-fhir-project -o ~/my-fhir-project/gofsh
```
and note that they doubled.

This is because the `ensureOutputDir` function gets called _after_ GoFSH has already read in all the stuff from the input folder.  So by the time we clear the folder, all its contents have already been processed.

I also feel like I'd really like a confirmation prompt here.  I know we don't do one for SUSHI, but in that case, it always goes to a `fsh-generated` folder (which is very obviously intended to only have generated code).  In this case, it could be any folder, and it seems more likely that people might edit results in the folder.

So, how would you feel about adding a prompt if the folder exists and asking the user to confirm they want to delete it?  It's fairly simple to do.  There is an example in the regression routine for SUSHI: https://github.com/FHIR/sushi/blob/master/regression/run-regression.ts#L175-L190

jafeltra (approved):
This looks good to me! I tested out and deleting the folder only processes the original set of JSON definitions, even if there was roundtrip, JSON output in the output folder previously. Continue processes the duplicate resources. Quit, well, quits.

cmoesel (approved):
Excellent!

PR #82, opened 2021-02-06
Support XML

ngfreiter:
This PR adds support for GoFSH processing XML inputs. This is handled with the new `-t` option, which can be `json-only`, `xml-only`, or `json-and-xml`.
* `json-only` (default): When this option is selected, only JSON is processed. If a FHIR XML file is detected which doesn't match any of the JSON resources (where matching is determined by a matching `resourceType` and `id`), then a warning is emitted to indicate that if the user wants to process that XML file, they need to switch from `json-only` mode. If all of the XML appears to just be duplicates of resources already defined in JSON, no warning is logged.
* `xml-only`: Works exactly the same as `json-only`, but for `xml`.
* `json-and-xml`: Both JSON and XML are processed. This does not handle duplicates, so those should be handled at some point later in the processing.


src/app.ts
```diff
@@ -86,7 +90,10 @@ async function app() {
   const dependencies = program.dependency?.map((dep: string) => dep.trim());
 
   // Load FhirProcessor and config object
-  const processor = getFhirProcessor(inDir, defs);
+  const fileType = ['json-only', 'xml-only', 'json-and-xml'].includes(program.fileType)
+    ? program.fileType
+    : 'json-only';
```
cmoesel, line 95: I would consider making this case-insensitive:
```ts
/^(json-only|xml-only|json-and-xml)$/i.test(program.fileType)
```

In addition, I think that if the user enters an supported value (e.g., `XML`), we should error out.  Otherwise the output they get may not be the output they expect.

src/app.ts
```diff
@@ -153,6 +160,14 @@ async function app() {
   results.forEach(r => console.log(r));
 
   if (program.fshingTrip) {
+    if (fileType === 'xml-only') {
```
cmoesel, line 163: If we do treat this case-insensitive elsewhere, we need to be case-insensitive everywhere else too; or normalize the `fileType` when we assign it by doing a `toLowerCase()`.

src/utils/Processing.ts
```diff
@@ -99,21 +102,82 @@ export function loadExternalDependencies(
...
+    logger.info(`Found ${jsonFiles.length} JSON files.`);
+    loadPrimaryFiles(jsonFiles, docs);
+    const nonDuplicateXMLFile = findNonDuplicateSecondaryFile(xmlFiles, docs);
+    if (nonDuplicateXMLFile) {
+      // We only find the first non-duplicate file and warn on it, because warnings for every non-duplicate may be annoying
+      logger.warn(
+        `The XML definition at ${nonDuplicateXMLFile} will be ignored since GoFSH is running in "json-only" mode.` +
+          ' To process XML definitions along with JSON, set the "-t" flag to "json-and-xml".' +
+          ' To process only XML definitions, set the "-t" flag to "xml-only".'
+      );
```
cmoesel, line 121: I feel like only printing one could be misleading (it makes it look like there is only one).  Can we get a count and indicate that?
> 17 XML definitions found without corresponding JSON definitions (for example, ${nonDuplicateXMLFile}). These definitions will be ignored since GoFSH is running in "json-only" mode....

src/utils/Processing.ts
```diff
@@ -99,21 +102,82 @@ export function loadExternalDependencies(
...
+    logger.info(`Found ${xmlFiles.length} XML files.`);
+    loadPrimaryFiles(xmlFiles, docs);
+    const nonDuplicateJSONFile = findNonDuplicateSecondaryFile(jsonFiles, docs);
+    if (nonDuplicateJSONFile) {
+      // We only find the first non-duplicate file and warn on it, because warnings for every non-duplicate may be annoying
+      logger.warn(
+        `The JSON definition at ${nonDuplicateJSONFile} will be ignored since GoFSH is running in "xml-only" mode.` +
+          ' To process JSON definitions along with XML, set the "-t" flag to "json-and-xml".' +
+          ' To process only JSON definitions, set the "-t" flag to "json-only", or leave it unset.'
+      );
```
cmoesel, line 133: Same as my comment above, but reverse.

cmoesel (changes requested):
I didn't get to test this, but was able to review the code while in a meeting and have a few suggestions / requests.

ngfreiter:
@cmoesel, comments should be addressed by https://github.com/FHIR/GoFSH/pull/82/commits/aac874b461e1e2ef0cb8989da00ccc33fd0db149.

cmoesel:
@ngfreiter - I tested this on https://github.com/hl7dk/dk-core/ -- and when running on the root (which seems reasonable), I get these errors:
> error Could not load /Users/cmoesel/dev/HL7/dk-core/input/includes/menu.xml: Unknown resource type: ul
> error Could not load /Users/cmoesel/dev/HL7/dk-core/input/pagecontent/extensions.xml: Unknown resource type: div
> error Could not load /Users/cmoesel/dev/HL7/dk-core/input/pagecontent/profiles.xml: Unknown resource type: div
> error Could not load /Users/cmoesel/dev/HL7/dk-core/input/pagecontent/terminology.xml: Unknown resource type: div

Do you think we should intentionally avoid resources whose paths contain `/input/includes/`, `/input/pagecontent/`, `/input/pages/`, `/input/intro-notes/`, `/input/images/`, and `/input/images-source/`?

ngfreiter:
> @ngfreiter - I tested this on https://github.com/hl7dk/dk-core/ -- and when running on the root (which seems reasonable), I get these errors:
> 
> > error Could not load /Users/cmoesel/dev/HL7/dk-core/input/includes/menu.xml: Unknown resource type: ul
> > error Could not load /Users/cmoesel/dev/HL7/dk-core/input/pagecontent/extensions.xml: Unknown resource type: div
> > error Could not load /Users/cmoesel/dev/HL7/dk-core/input/pagecontent/profiles.xml: Unknown resource type: div
> > error Could not load /Users/cmoesel/dev/HL7/dk-core/input/pagecontent/terminology.xml: Unknown resource type: div
> 
> Do you think we should intentionally avoid resources whose paths contain `/input/includes/`, `/input/pagecontent/`, `/input/pages/`, `/input/intro-notes/`, `/input/images/`, and `/input/images-source/`?

Yeah I went back and forth on this. One the one hand the errors are annoying. On the other hand, the user is saying they want us to process XML, and so we try to process XML. I like that the behavior is transparent and simple. That being said I think it is a safe bet to ignore those directories, so I can add that if you say the word.

cmoesel:
I guess if the error message were better ("Not a resource") or if it was only a warning when it's not a resource, I'd feel better about it.  But I don't know if we can easily differentiate the errors like that (as opposed to a FHIR resource that truly can't be parsed for some other reason).

As it is, this is not really an ERROR situation -- the user does not need to be concerned about it.  And I don't really want to confuse the user by making it seem like it is an ERROR situation and they _should_ be concerned about.  So I still think we should skip them.

Would it make you feel better if we logged an INFO that says "skipping ..." so the user knows that even though it is XML, we skipped it?

ngfreiter:
> I guess if the error message were better ("Not a resource") or if it was only a warning when it's not a resource, I'd feel better about it. But I don't know if we can easily differentiate the errors like that (as opposed to a FHIR resource that truly can't be parsed for some other reason).
> 
> As it is, this is not really an ERROR situation -- the user does not need to be concerned about it. And I don't really want to confuse the user by making it seem like it is an ERROR situation and they _should_ be concerned about. So I still think we should skip them.
> 
> Would it make you feel better if we logged an INFO that says "skipping ..." so the user knows that even though it is XML, we skipped it?

No need to add more logs, I will just skip them.

ngfreiter:
@cmoesel, done in https://github.com/FHIR/GoFSH/pull/82/commits/39b87fef6e082762caac62c322bad48005b92816.

cmoesel (approved):
I dig it!

mint-thompson (approved):

PR #83, opened 2021-02-08
Error on Duplicate Names

### LINKED ISSUE #79: GoFSH doesn't error or warn on duplicate names
If the source to GoFSH contains two profiles w/ the same name, GoFSH happily generates both definitions in FSH -- each with the same name.  This also happens for Extensions (and I'm guessing for other types too).  You can easily reproduce this by following the instructions in #78.

GoFSH should either:
* emit an error in this case
* or emit a warning and rename one of them

jafeltra:
This PR fixes #79 by logging and error and not writing FSH definitions that had duplicate names.

In the past, we have had some ability to uniquely identify duplicate names and de-duplicate them (mappings and instances add information to the names and are simplified if possible in optimizers). However, this is not as simple for profiles and for true duplicate definitions, there will be no unique piece of information we can add. In those cases, we log an error when we see the duplicate name and we do not add the definition to the package to be output. This can be replicated by following the steps in the issue.

While doing this, I realized these errors and others that were occurring during the export were being swallowed by the export counter when it tried to log on the same line as the last counting log. For now, I've commented that line out so that we get all the logs, and we can revisit the fancy counter later.


src/processor/Package.ts
```diff
@@ -60,3 +75,17 @@ export class Package {
...
 }
+
+function isDuplicateDefinition(
+  array: NamedExportable[],
+  resource: NamedExportable,
+  type: string
+): boolean {
+  if (array.find(e => e.name === resource.name)) {
+    logger.error(
+      `Encountered ${type} with a duplicate name, ${resource.name}, which GoFSH cannot make unique. Fix the source file to resolve this error. The definition will still be written to a .fsh file.`
```
cmoesel, line 86: > The definition will still be written to a .fsh file.

What does this mean?  I think you mean that only one of the definitions will be written, but I only know that based on your implementation.  Can you clarify this?
> Only one of the duplicate definitions will be written to a .fsh file. Remaining duplicates will be skipped.

src/processor/Package.ts
```diff
@@ -60,3 +75,17 @@ export class Package {
     }
   }
 }
+
+function isDuplicateDefinition(
+  array: NamedExportable[],
+  resource: NamedExportable,
+  type: string
+): boolean {
+  if (array.find(e => e.name === resource.name)) {
```
ngfreiter, line 84: I'm wondering if we want to deduplicate based on name, since SUSHI can produce two profiles with the same `name`, for example, the following:
```
Profile: MyPatient
Parent: Patient
* ^name = "foo"

Profile: MyPatient2
Parent: Patient
* ^name = "foo"
```
is FSH that runs without error, but then causes a duplicate error when I run the output of SUSHI through GoFSH. In SUSHI I think we use the combination of `resourceType` and `id` to detect duplicates. Could we do something similar here?

src/processor/Package.ts
jafeltra: Oops yes, the error message is wrong and nothing is written to a FSH file. It should say 

> The definition will not be written to a .fsh file.

src/processor/Package.ts
cmoesel: Ugh.  So that probably is the best way to do it.  But IIRC there were some complications for implementation in trying to keep both definitions and number them.  I think that was the first approach @jafeltra was experimenting with -- but I'll let her comment on it.

cmoesel: OK, just documented where we landed on this:

- first, deduplicate the incoming JSON by `resourceType` and `id` -- logging an error and tossing the duplicate
- then, on the export side, detect duplicate `name`s and log an error for them -- but still let the duplicates through anyway.  The FSH will be invalid, but these we'll leave it to the author to decide how to resolve that.
- there shouldn't be many duplicate names in the previous step if we've already removed all the duplicates by `resourceType` and `id`

jafeltra:
I believe that 1dc1f5e covers the changes we discussed earlier. 

- The check for deduplicating based on `resourceType` and `id` is done after we `getLakeOfFHIR`, so I think it should work with XML support
- When we still find a duplicate `name` on the export side, an error is logged but the definition is added to the package, so it will be written to a FSH file. I'm not sure if this should be a warning since we'll still write it, but the FSH will be invalid when used in SUSHI, so I kept it at error level.
- I added `id` to a few fixture files that were logging an error when expecting 0 to be logged. I figured we'd always expect resources to have `id` and this was just because we made the test fixtures too simple.

src/processor/Package.ts
jafeltra: I believe that 1dc1f5e covers these changes.

src/processor/Package.ts
jafeltra: I think the error message should be more clear in 1dc1f5e. It should add the definition and write to a FSH file now though.

src/processor/LakeOfFHIR.ts
```diff
@@ -95,4 +97,22 @@ export class LakeOfFHIR implements utils.Fishable {
...
+      const isDuplicate =
+        a.content.id === b.content.id && a.content.resourceType === b.content.resourceType;
+      if (isDuplicate) {
+        logger.error(
+          `Encountered a definition with the same resourceType and id as a previous definition: file "${a.path}" has resourceType ${a.content.resourceType} and id ${a.content.id}. ` +
+            'FHIR definitions should have unique resourceType and id. The duplicate will not be processed by GoFSH.'
+        );
+      }
+      return isDuplicate;
+    });
```
cmoesel, line 116: Two questions/suggestions:

1. You list the path for `a` -- which almost makes it sound like `a` is the duplicate that will be tossed.  But I think `b` is the duplicate that is tossed, right?  If so, maybe it makes sense to tell the user the path of the thing we're tossing?  Or both paths?

2. If this happens a lot, that could be a lot of errors.  Since this is done in a loop (kind of), might it make sense just to collect one massive error something like:
> Encountered 17 definitions with the same resourceType and id as a previous definition.  FHIR definitions should be unique by resourceType and id.  The following duplicate definitions will not be processed by GoFSH:
> - /path/to/foo.xml (StructureDefinition/foo) matches /path/to/foo.json
> - /path/to/bar.xml (StructureDefinition/bar) matches /path/to/bar.json
> - ...

What do you think about that suggestion, @jafeltra?

jafeltra: Yeah I was actually wondering about the number of errors, but when I was wondering that, I must have forgotten about the XML processing that would make this happen more often. I don't mind collecting a running tally and logging once.

As for your first point, I actually thought `a` was the duplicate that is being tossed, but I can double check. But maybe if I log the paths like you did in your example it doesn't matter quite as much.

cmoesel: I think you'll still need to know if the duplicate is `a` or `b` in order to create the list / error message correctly.  I don't think the lodash doc is super clear about it, but I think maybe I misunderstood it and you may be right about `a` being the duplicate.

ngfreiter: I tried it with the REPL in lodash docs, and when I did this:
```
var objects = [{ 'x': 1, 'y': 'first' }, { 'x': 2, 'y': 1 }, { 'x': 1, 'y': 'second' }];
_.uniqWith(objects, (a, b) => a.x === b.x);
```
I got this:
```
[{x: 1, y: "first"}, {x: 2, y: 1}]
```
So I think that shows the second value is tossed right?

ngfreiter: Oh wait maybe that doesn't answer our question I see.

cmoesel: Right...  at the point in the comparator where we're checking for the dup, we only have `a` and `b`.  I guess we could loop through the original array to find each and see which has the lower index, but... that seems like it should be unnecessary (I hope).

jafeltra: I don't know that it truly matters much, but it might help a little when understanding the error message. That said, I think it is working as we'd expect, and this is what the message looks like now with f77d157:

image

jafeltra: Oh no, GitHub did the thing were it didn't show me all the new conversation. Let me read and ignore my message if there are obvious contradictions.

jafeltra:
> Ummm... couldn't you just `dupPaths.length` instead of creating and tracking `dupNum`?

Ha yes. I was going too fast. Fixed in bc5c1f9.

cmoesel (approved):
Looks good to me.  Thanks for picking this one up, @jafeltra!

ngfreiter (approved):
This looks good to me, and looks like it shouldn't conflict with the XML support at all.

PR #84, opened 2021-02-08
be slightly nicer to sea creatures

mint-thompson:
Sea creatures can no longer be harmed or lost, but may be vexed, displeased, irritated, or confused. Additionally, sea creatures may be satisfied or contented.

I just... didn't want to harm them. Please understand.


cmoesel (approved):
I really like your word choice.  Very creative and fun!

17 narwhals vexed!

ngfreiter (approved):
Love it!

PR #85, opened 2021-02-08
Bump SUSHI to 1.2.0

cmoesel:
Bumps SUSHI to v1.2.0.

In the process of testing this, I ran into two bugs, but I don't think either one is related to SUSHI 1.2.0.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2907,9 +2907,9 @@
       "optional": true
     },
     "fsh-sushi": {
-      "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/fsh-sushi/-/fsh-sushi-1.1.0.tgz",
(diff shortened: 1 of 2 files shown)


julianxcarter (approved):

mint-thompson (approved):

cmoesel:
Actually, I take that back.  I think this bug might be caused by SUSHI 1.2.0:
https://github.com/FHIR/GoFSH/issues/87

cmoesel:
If #87 is introduced by SUSHI 1.2.0, then we should hold this PR and try to add a commit that fixes #87, so the update to SUSHI 1.2.0 is accompanied by required fixes.

cmoesel:
OK.  #87 isn't really introduced by SUSHI 1.2.0 in the GoFSH codebase.  Rather, it's introduced when you run GoFSH against an IG that was built using SUSHI 1.2.0.  So there is no harm in merging this PR and operating w/ SUSHI 1.2.0 in the GoFSH code base.

Issue #86, opened 2021-02-08
GoFSH not reflecting `^status` correctly


cmoesel:
If you run SUSHI 1.2.0 against https://github.com/HL7/fhir-mCODE-ig/ and then run GoFSH against the resulting `fsh-generated` folder (using the fshing-trip option), one definition will have the wrong status:
<img width="1259" alt="image" src="https://user-images.githubusercontent.com/2278253/107290339-5ad9d100-6a34-11eb-8a3d-7089c003e1f3.png">

I'll need to look at the SUSHI code, but I think maybe SUSHI defaults `status` to `active`, and `GoFSH` is not creating a `^status` rule even though the SD has it marked as `draft`.

PR #88, opened 2021-02-09
Create a CaretRule when 'status' is not 'active'

ngfreiter:
This fixes https://github.com/FHIR/GoFSH/issues/86.

SUSHI always sets `status` to `active`, so if there is an SD that has a `status` other than `active`, we must always make a CaretValueRule.

To recreate the issue, you can follow the instructions in the issue description on `master`, and you should see that the `mcode-tumor` SD has the incorrect `status`. Then, when use this branch instead of `master`, you should see that the `status` is correctly set to `draft`.

### DIFF
--- a/test/extractor/CaretValueRuleExtractor.test.ts
+++ b/test/extractor/CaretValueRuleExtractor.test.ts
@@ -12,6 +12,7 @@ describe('CaretValueRuleExtractor', () => {
   let looseSD: any;
   let looseVS: any;
   let looseCS: any;
+  let looseBSSD: any;
   let config: fshtypes.Configuration;
   let defs: fhirdefs.FHIRDefinitions;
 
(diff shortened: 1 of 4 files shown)


jafeltra (approved):
Sneaky way to up those lines added.

More seriously, when I run mCODE through GoFSH on master and on this branch, the two diffs I have on `status` in `CTCAdverseEvent` and `Tumor` are gone and the statuses are set correctly. 👍

test/extractor/CaretValueRuleExtractor.test.ts
```diff
@@ -59,6 +65,16 @@ describe('CaretValueRuleExtractor', () => {
       expect(caretRules).toEqual<ExportableCaretValueRule[]>([expectedRule]);
     });
 
+    it('should always extract a caret value rule for status when it is not "active"', () => {
+      const sd = cloneDeep(looseBSSD);
+      sd.status = 'draft';
```
cmoesel, line 70: Do you need to clone `looseBSSD` and set the status to `'draft'`?  Isn't it already draft?

cmoesel (approved):
I approve, although I noticed one thing in the test that seemed unnecessary?

test/extractor/CaretValueRuleExtractor.test.ts
ngfreiter: Not really, I did purposefully set it as 'draft' just so that someone who was looking at the test could see what was going on without having to look at the definition.

ngfreiter: But it is redundant, and I don't mind removing it if you don't like it.

cmoesel: Eh.  It's fine.  Let's just get it merged!

PR #89, opened 2021-02-09
Fix constraint and mapping indices

### LINKED ISSUE #87: GoFSH creating `^` rules against the wrong constraints
If you run SUSHI 1.2.0 against https://github.com/HL7/fhir-mCODE-ig/ and then run GoFSH against the resulting `fsh-generated` folder (using the fshing-trip option), you'll notice that several invariants are produced from the GoFSH definitions that aren't produced from the original FSH definitions.  For example:
image

If you look at the produced FSH for that profile, you'll see this:
image

I think this has something to do w/ SUSHI 1.2.0 differentials no longer carrying the full array of `constraint` and `mapping` -- but only carrying those elements that are different.  It's causing the `^` rules to target the wrong elements -- but really, I don't think we should need `^` rules there at all?  At least not in this case?

cmoesel:
Fixes #87 and other stuff.

There are primarily three things fixed with this PR:
* If a `constraint[n].source` is what SUSHI would have generated for source anyway, exclude it.
* Fix constraint indices to be relative to the constraint array in the snapshot (since the array in the differential can be a subset)
* Fix mapping indices to be relative to the mapping array in the snapshot (since the array in the differential can be a subset)

For the last two bullets, if the snapshot is not available, a warning will be issued and comments will be exported in the FSH indicating that the indices cannot be trusted.

You can test using mCODE if you'd like (as described in #87).  That is actually fixed by the first bullet above (no more caret rules for `constraint.source`.

You can also test against the attached FSH project.  To do so, unzip the project then:
* Run SUSHI 1.2.0 in the FSH project to generate an SD w/ modified constraints and mappings
* Run GoFSH master branch against the project with `-f` flag and notice issues in the round-trip related to indexing trouble
* Run GoFSH from this branch against the project with `-f` flag and notice the new warnings.  Also look at the generated FSH and notice the comments.  And last, you'll still see that the round trip is messed up.
* Run SUSHI 1.2.0 in the FSH project again, but this time with the `-s` flag to generate the snapshot
* Run GoFSH from this branch against the project with the `-f` flag and note that there are no warnings and the round-trip is clean!

[ConstraintsAndMappings.zip](https://github.com/FHIR/GoFSH/files/5954129/ConstraintsAndMappings.zip)

### DIFF
--- a/src/exportable/ExportableCaretValueRule.ts
+++ b/src/exportable/ExportableCaretValueRule.ts
@@ -1,7 +1,9 @@
 import { fhirtypes, fshrules, fshtypes } from 'fsh-sushi';
+import { EOL } from 'os';
 import { ExportableRule } from '.';
 
 export class ExportableCaretValueRule extends fshrules.CaretValueRule implements ExportableRule {
+  comment: string;
   constructor(path: string) {
     super(path);
   }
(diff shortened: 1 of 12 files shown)


ngfreiter (approved):
Looks good, and worked as advertised.

jafeltra (approved):
Works as expected and the mCODE roundtrip looks clean!

PR #91, opened 2021-02-10
If an element references an undefined mapping, create it

cmoesel:
Although it's likely not valid, at least one IG (*ahem*, US Core) contains some elements with `mapping.identity` values that don't match a top-level mapping defined in the SD.  Rather than ignore them or create caret rules, create a top-level Mapping placeholder, log a warning, and write a comment.  That way the user can correct the obviously wrong mistake.  This also avoids some other issues we can run into when expressing all the mappings as caret rules.

If you'd like to test this w/ a real IG, try, oh I don't know, US Core 3.1.1?  On `master`, you'll notice that it creates a bunch of caret rules for mappings on `USCorePatient`.  You'll also notice if you round trip that in `USCorePatient`, the `mapping` array for `Patient.extension:birthsex` is messed up (actually due to another slightly incorrect thing about US Core -- but exacerbated by the mapping caret rules).

Or if you'd like to test is against a simple SD, you can try this one:
```
{
  "resourceType": "StructureDefinition",
  "id": "my-observation",
  "extension": [
    {
      "url": "http://hl7.org/fhir/StructureDefinition/structuredefinition-category",
      "valueString": "Clinical.Diagnostics"
    },
    {
      "url": "http://hl7.org/fhir/StructureDefinition/structuredefinition-security-category",
      "valueCode": "patient"
    }
  ],
  "url": "http://sample.org/StructureDefinition/my-observation",
  "name": "MyObservation",
  "status": "active",
  "fhirVersion": "4.0.1",
  "mapping": [
    {
      "identity": "workflow",
      "uri": "http://hl7.org/fhir/workflow",
      "name": "Workflow Pattern"
    },
    {
      "identity": "sct-concept",
      "uri": "http://snomed.info/conceptdomain",
      "name": "SNOMED CT Concept Domain Binding"
    },
    {
      "identity": "v2",
      "uri": "http://hl7.org/v2",
      "name": "HL7 v2 Mapping"
    },
    {
      "identity": "rim",
      "uri": "http://hl7.org/v3",
      "name": "RIM Mapping"
    },
    {
      "identity": "w5",
      "uri": "http://hl7.org/fhir/fivews",
      "name": "FiveWs Pattern Mapping"
    },
    {
      "identity": "sct-attr",
      "uri": "http://snomed.org/attributebinding",
      "name": "SNOMED CT Attribute Binding"
    }
  ],
  "kind": "resource",
  "abstract": false,
  "type": "Observation",
  "baseDefinition": "http://hl7.org/fhir/StructureDefinition/Observation",
  "derivation": "constraint",
  "differential": {
    "element": [
      {
        "id": "Observation.category",
        "path": "Observation.category",
        "mapping": [
          {
            "identity": "SomeRogueMapping",
            "map": "SomeRogueThing"
          }
        ]
      },
      {
        "id": "Observation.focus",
        "path": "Observation.focus",
        "mapping": [
          {
            "identity": "SomeRogueMapping",
            "map": "SomeOtherRogueThing"
          }
        ]
      }
    ]
  }
}
```

On master, it will create some warnings about mapping indices and you'll see caret rules for mappings in the rules.  On this branch, you'll see a warning about creating a missing mapping, along with a comment telling the author to update it.

### DIFF
--- a/src/extractor/MappingExtractor.ts
+++ b/src/extractor/MappingExtractor.ts
@@ -64,7 +64,6 @@ export class MappingExtractor {
       const mapping = new ExportableMapping(`${m.identity}-for-${sd.name}`);
       mapping.id = m.identity;
       mapping.source = sd.name;
-      mapping.name = `${mapping.id}-for-${mapping.source}`;
       if (m.name) mapping.title = m.name;
       if (m.uri) mapping.target = m.uri;
       if (m.comment) mapping.description = m.comment;
--- a/src/extractor/MappingExtractor.ts
+++ b/src/extractor/MappingExtractor.ts
@@ -6,7 +6,7 @@ import {
   ProcessableStructureDefinition,
   StructureDefinitionProcessor
 } from '../processor';
-import { getPath } from '../utils';
+import { getPath, logger } from '../utils';
 
 export class MappingExtractor {
   static process(
@@ -37,7 +37,7 @@ export class MappingExtractor {
         return ProcessableElementDefinition.fromJSON(rawElement, false);
       }) ?? [];
 
-    const parentMappings = this.extractMappings(parent, parentSDElements);
+    const parentMappings = this.extractMappings(parent, parentSDElements, true);
 
     // Only return mappings that are new to the profile or inherited mappings with new MappingRules
     const newMappings: ExportableMapping[] = [];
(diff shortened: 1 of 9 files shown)


src/extractor/MappingExtractor.ts
```diff
@@ -57,10 +57,12 @@ export class MappingExtractor {
 
   static extractMappings(
     sd: ProcessableStructureDefinition,
-    elements: ProcessableElementDefinition[]
+    elements: ProcessableElementDefinition[],
+    processingParent = false
   ): ExportableMapping[] {
     const mappings = (sd.mapping || []).map(m => {
-      const mapping = new ExportableMapping(m.identity);
+      const mapping = new ExportableMapping(`${m.identity}-for-${sd.name}`);
```
ngfreiter, line 64: Doesn't this line make 67 unnecessary?

ngfreiter (commented):
It looks good and is working as expected. I have one minor question.

src/extractor/MappingExtractor.ts
cmoesel: Maybe... ;-)

cmoesel: Fixed in 8c6d0ab.

ngfreiter (approved):
Looks good to me!

mint-thompson (approved):

PR #92, opened 2021-02-10
Support alternate id syntax for choices when looking for matching id in snapshot

cmoesel:
It turns out that the code I wrote in #89 didn't take into account the fact that choices allow for an alternate `id` and `path` syntax in the differential.  For example, an element with id `Observation.valueQuantity` in the differential actually matches the element with id `Observation.value[x]:valueQuantity` in the snapshot.  Now we take that into account when looking for the matching snapshot element to fix constraint and mapping ids.

You can test this by running GoFSH on the attached StructureDefinition.  On master, you'll see two warnings because it couldn't find the snapshot element (even though it _is_ there) -- and the round-trip is dirty.  On this branch, the warnings go away (and round-trip is clean).

[StructureDefinition-my-profile.json.zip](https://github.com/FHIR/GoFSH/files/5956078/StructureDefinition-my-profile.json.zip)

### DIFF
--- a/src/extractor/CaretValueRuleExtractor.ts
+++ b/src/extractor/CaretValueRuleExtractor.ts
@@ -31,7 +31,7 @@ export class CaretValueRuleExtractor {
         let newIndex: number;
         if (structDef?.snapshot?.element?.length > 0) {
           const diffConstraint = inputJSON.constraint[parseInt(constraintMatch[1])];
-          const snapElement = structDef.snapshot.element.find(el => el.id === inputJSON.id);
+          const snapElement = findMatchingSnapshot(inputJSON.id, structDef);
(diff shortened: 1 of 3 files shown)


ngfreiter (approved):
Looks good to me!

mint-thompson (commented):
Is it possible for the opposite case to occur? That is, `Observation.valueQuantity` exists in the snapshot, and `Observation.value[x]:valueQuantity` exists in the differential.

cmoesel:
I don't think so.  My understanding is that those alternate syntaxes are exclusively allowed in the differential only.

cmoesel:
Found where it is documented.  Not official FHIR documentation, but actually the best available documentation on choice type renaming:
image

mint-thompson (approved):

PR #96, opened 2021-02-12
Assign missing ids on definitions

jafeltra:
This PR assigns an `id` to any definitions that are missing `id`s. If the definition has a `name`, we try to use that. Otherwise, we just assign `id-n`, where `n` counts up for each resource that is missing an `id` and `name`. I also realized that when we write out the FSH files, an error would be thrown if there was no name listed in the `index.txt` file, so this also supports using the `id`, which we now know will exist.

There are some test files in Slack that you can use to try this. But there are also quite a few files in `./test/processor/fixtures` that are missing `id`s and some are missing `name` as well, so you could also use those for testing as well.

### DIFF
--- a/src/processor/LakeOfFHIR.ts
+++ b/src/processor/LakeOfFHIR.ts
@@ -1,6 +1,8 @@
 import { fhirdefs, utils } from 'fsh-sushi';
 import { uniqWith } from 'lodash';
 import { CodeSystemProcessor } from './CodeSystemProcessor';
+import { CONFORMANCE_AND_TERMINOLOGY_RESOURCES } from './InstanceProcessor';
+import { StructureDefinitionProcessor } from './StructureDefinitionProcessor';
 import { ValueSetProcessor } from './ValueSetProcessor';
 import { WildFHIR } from './WildFHIR';
 import { logger } from '../utils';
@@ -99,14 +101,17 @@ export class LakeOfFHIR implements utils.Fishable {
   }
 
   /**
-   * Removes any definitions from this.docs that have the same resourceType and id as a previous definition.
+   * Removes any definitions from this.docs that have the same resourceType and id as
+   * a previous definition, as long as there is a defined id.
    * Logs an error when it finds a duplicate
    */
   removeDuplicateDefinitions() {
     const dupPaths: string[] = [];
     this.docs = uniqWith(this.docs, (a, b) => {
       const isDuplicate =
-        a.content.id === b.content.id && a.content.resourceType === b.content.resourceType;
+        a.content.id != null &&
+        a.content.id === b.content.id &&
+        a.content.resourceType === b.content.resourceType;
       if (isDuplicate) {
         dupPaths.push(`${a.path} (${a.content.resourceType}/${a.content.id}) matches ${b.path}`);
       }
@@ -123,23 +128,32 @@ export class LakeOfFHIR implements utils.Fishable {
   }
 
   /**
-   * All definitions should have a resourceType and id. If any definition is missing an id,
-   * we add one either based on the name or with a simple counter.
-   * Logs a warning if it finds any definitions without an id
+   * All definitions that will be Instances should have a resourceType and id.
+   * If any Instance is missing an id, we add one.
+   * If the instance is a conformance or terminology resource, we try to base the id off the name if it is available.
+   * If there is no name or if it is any other type of resource, we add a clearly generated id with a counter.
+   * Log a warning if it finds any definitions without an id
    */
   assignMissingIds() {
     const createdIdPaths: string[] = [];
     let generatedId = 0;
     this.docs.forEach((d, index) => {
-      if (d.content.id == null) {
-        // Try to be smart and set the id to the existing name
-        d.content.id = d.content.name;
+      const isSpecialFSHType =
+        StructureDefinitionProcessor.isProcessableStructureDefinition(d.content) ||
+        CodeSystemProcessor.isProcessableCodeSystem(d.content) ||
+        ValueSetProcessor.isProcessableValueSet(d.content);
+      if (d.content.id == null && !isSpecialFSHType) {
+        if (CONFORMANCE_AND_TERMINOLOGY_RESOURCES.has(d.content.resourceType)) {
+          // Try to be smart and set the id to the existing name
+          d.content.id = d.content.name?.replace(/_/g, '-').slice(0, 64); // Turn a valid name into a valid id
+        }
+
         if (d.content.id == null) {
-          // If there is no existing name, generate an id with a counter
-          d.content.id = `id-${generatedId++}`;
+          // If a Conformance/Terminology instance didn't have a name or if this any other resourceType, generate an id with a counter
+          d.content.id = `GOFSH-GENERATED-ID-${generatedId++}`;
           // If another definition happen to have the same id as the one we just generated with a counter, increase the counter
           while (this.docs.some((a, i) => a.content.id === d.content.id && index !== i)) {
-            d.content.id = `id-${generatedId++}`;
+            d.content.id = `GOFSH-GENERATED-ID-${generatedId++}`;
           }
         }
         createdIdPaths.push(`${d.path} (${d.content.resourceType}/${d.content.id})`);
--- a/src/processor/LakeOfFHIR.ts
+++ b/src/processor/LakeOfFHIR.ts
@@ -130,9 +130,18 @@ export class LakeOfFHIR implements utils.Fishable {
   assignMissingIds() {
     const createdIdPaths: string[] = [];
     let generatedId = 0;
-    this.docs.forEach(d => {
+    this.docs.forEach((d, index) => {
       if (d.content.id == null) {
-        d.content.id = d.content.name ?? `id-${generatedId++}`;
+        // Try to be smart and set the id to the existing name
+        d.content.id = d.content.name;
+        if (d.content.id == null) {
+          // If there is no existing name, generate an id with a counter
+          d.content.id = `id-${generatedId++}`;
+          // If another definition happen to have the same id as the one we just generated with a counter, increase the counter
+          while (this.docs.some((a, i) => a.content.id === d.content.id && index !== i)) {
+            d.content.id = `id-${generatedId++}`;
+          }
+        }
         createdIdPaths.push(`${d.path} (${d.content.resourceType}/${d.content.id})`);
       }
     });
--- a/src/processor/LakeOfFHIR.ts
+++ b/src/processor/LakeOfFHIR.ts
@@ -121,4 +121,27 @@ export class LakeOfFHIR implements utils.Fishable {
       );
     }
   }
+
+  /**
+   * All definitions should have a resourceType and id. If any definition is missing an id,
+   * we add one either based on the name or with a simple counter.
+   * Logs a warning if it finds any definitions without an id
+   */
+  assignMissingIds() {
+    const createdIdPaths: string[] = [];
+    let generatedId = 0;
+    this.docs.forEach(d => {
+      if (d.content.id == null) {
+        d.content.id = d.content.name ?? `id-${generatedId++}`;
+        createdIdPaths.push(`${d.path} (${d.content.resourceType}/${d.content.id})`);
+      }
+    });
+
+    if (createdIdPaths.length > 0) {
+      logger.warn(
+        `Encountered ${createdIdPaths.length} definition(s) that were missing an id. GoFSH created ids for the following:` +
+          `\n  - ${createdIdPaths.join('\n  - ')}`
+      );
+    }
+  }
 }
--- a/src/export/FSHExporter.ts
+++ b/src/export/FSHExporter.ts
@@ -57,7 +57,7 @@ export class FSHExporter {
         .forEach((exportable: NamedExportable) => {
           // The index will have the name, FSH type, and file of the entity
           index.push([
-            exportable.name ?? exportable.id,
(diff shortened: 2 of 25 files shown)


mint-thompson (changes requested):
This looks good! There's one additional test case I'd like you to consider. It's possible for a resource to be defined with an id of `id-0` or similar. In this case, it will end up with the same id as one that gets assigned a generated id. So, I think it would be useful to check to see if the generated id matches an existing id on a resource of the same type.

jafeltra:
@mint-thompson that was a great catch! I added a test case for that scenario and included a check to see if we already have another `id` that matches the `id-*` value we just assigned and increment the counter if it does. Let me know what you think!

mint-thompson (approved):

src/processor/LakeOfFHIR.ts
```diff
@@ -121,4 +121,36 @@ export class LakeOfFHIR implements utils.Fishable {
...
+  assignMissingIds() {
+    const createdIdPaths: string[] = [];
+    let generatedId = 0;
+    this.docs.forEach((d, index) => {
+      if (d.content.id == null) {
+        // Try to be smart and set the id to the existing name
+        d.content.id = d.content.name;
+        if (d.content.id == null) {
+          // If there is no existing name, generate an id with a counter
+          d.content.id = `id-${generatedId++}`;
```
ngfreiter, line 139: I've found that if there is no existing name, then we don't actually generate a profile at all. Are you seeing the same thing? Are there cases where we generate entities when the `name` is not present? 

This may not necessarily be in the scope of this PR, but if we aren't handling entities that don't have names, then I'm not sure we need the code to handle them here. And confusingly, we log a message indicating that GoFSH created an ID for a profile that then does not exist.

ngfreiter (commented):
I have one question that may be out of scope, but interested to hear what you think.

src/processor/LakeOfFHIR.ts
jafeltra: Hmm that is a great point and a great question. I hadn't actually noticed that when implementing this, but now that I'm looking at it, I notice that we are generating CodeSystems, ValueSets, and Instances that don't have a `name`, but not StructureDefinitions or ImplementationGuides. I'm not entirely sure what the right way forward is, besides that it should probably be consistent. Now that we're setting `id`s on everything, it feels like everything should be generated now, rather than changing some so they do not generate. But I'm also interested in what you and @mint-thompson think. (And anyone else on the team too.)

The other thing that feels somewhat related is that when we do the roundtrip `-f`, the names of the files change when SUSHI builds them because SUSHI uses the `resourceType-id` scheme, which, when you're missing an id on the original input, it is not included in the name either. I'm not sure how to fix that either though. And I'm not sure if either of this should be a separate task or if it makes sense to just do them now.

cmoesel: > I notice that we are generating CodeSystems, ValueSets, and Instances that don't have a name, but not StructureDefinitions or ImplementationGuides.

This is only a problem if people give us bad FHIR.  Whereas `name` is _optional_ in `ValueSet`, `CodeSystem`, and many `Instance`s, it is _required_ (`1..1`) on `StructureDefinition` and `ImplementationGuide`.  So the only way we'd ever end up having one without a name is if we started with invalid FHIR definitions.  (Which has happened before, so...).

> Now that we're setting ids on everything, it feels like everything should be generated now, rather than changing some so they do not generate.

If it's a simple change, that would be fine w/ me.  As I said above, lacking `name` on `StructureDefinition` or `ImplementationGuide` is technically invalid -- so if someone gives us invalid FHIR, I'm OK with making up a name so it is valid and we can write it.

> The other thing that feels somewhat related is that when we do the roundtrip -f, the names of the files change when SUSHI builds them because SUSHI uses the resourceType-id scheme...

Yeah, I think I'm OK with this.  FSHing trip should be useful to the extent that it can be so, but I don't think we should change SUSHI's filename scheme in order to satisfy FSHing trip.  If anything, we should first see if there are things we can do in FSHing trip so that it knows how old and new files map to each other and can correctly line them up when doing the diffs.

jafeltra: So slight snag. After talking at stand up, I was going to just always add a `name` if we tried setting and `id` and there was no `name` set either. But I don't actually think we can take that simplistic of an approach because not all definitions will have a name because not all `Instances` will have a name. So I think that still leaves us with the discrepancy that Nick pointed out - we might log a warning saying we changed the id of a definition that does not actually get exported as FSH because it was also missing a `name`.

So I am back to not being entirely sure what the right approach is. The two thoughts I have are:
- I could log a warning when we skip over StructureDefinitions without names. Maybe then the first warning about updating the `id` would be less confusing if there is a later warning saying why that definition didn't actually get created.
- I could check if the definition is a type `StructureDefinition` or `ImplementationGuide` before setting an `id` in this function since those seem to be the two that require `name` when processing, and then we wouldn't log a warning for that file.

cmoesel: I'm not sure I fully understand the issue.  If you can check if the definition is a `StructureDefinition` or `ImplementationGuide` in order to suppress the warning, then can't you check if the definition is a `StructureDefinition` or `ImplementationGuide` to decide if it is OK to set a name?

I.e., even if you didn't do it in this function, couldn't you create a `LakeOfFHIR.assignMissingNames()` function that is just like this but assigns missing names to SDs and IGs that are missing names?  Or... I think Nick said that this doesn't happen w/ ValueSets/CodeSystems that are missing names.  So... how do they handle it?  Do they handle it in their distinct processors or exporters?

jafeltra: Alright! I think I've got it! (although, who really knows). Instead of checking for and adding `name`s as needed in `assignMissingIds()`, I updated the `StructureDefinition.isProcessableStructureDefinition` function to align more closely with the checks in `CodeSystem` and `ValueSet`. An SD is now considered processable if it has a `resourceType` of `'StructureDefinition'` and if either the `name` or `id` is defined. Then, it uses the input's `name` or `id` to set the Exportable's name. Technically we shouldn't need to worry about having a missing `name` since it is required, but hopefully this will appropriately catch the cases where we have invalid FHIR.

So now I _think_ this thread is set! Let me know what you think!

src/processor/LakeOfFHIR.ts
```diff
@@ -121,4 +121,36 @@ export class LakeOfFHIR implements utils.Fishable {
...
+   * we add one either based on the name or with a simple counter.
+   * Logs a warning if it finds any definitions without an id
+   */
+  assignMissingIds() {
+    const createdIdPaths: string[] = [];
+    let generatedId = 0;
+    this.docs.forEach((d, index) => {
+      if (d.content.id == null) {
+        // Try to be smart and set the id to the existing name
+        d.content.id = d.content.name;
```
ngfreiter, line 136: I think setting the `id` to the existing name is not always a foolproof way of doing things. On many resources, such as a `Patient`, or an `Organization`, the `name` does not mean the same thing as on a `StructureDefinition`. If I give as input a `Patient` without a `name` or `id` (which I think is valid FHIR @cmoesel?) then I get something that looks like this:
image
Since we try to use the `name` array. I'm wondering if it may just be better not to be smart here. What are the downsides of just making every missing `id` into a generated one? I think using `name` given its inconsistent meaning in FHIR is risky.

ngfreiter (changes requested):

src/processor/LakeOfFHIR.ts
cmoesel: Good catch, @ngfreiter.  I like the idea of setting the `id` based on the name, but there are a few gotchas we didn't think of.
* As Nick noted, `name` doesn't always mean what we want it to mean.  But... if we checked if the `resourceType` was a conformance or terminology resource first, then we could more confidently use the name. We already have a list of these resource types [here](https://github.com/FHIR/GoFSH/blob/master/src/processor/InstanceProcessor.ts#L101-L117).
* That said, we'd still need to apply a little more magic to ensure the `id` fits the requirements of `id` in FHIR -- which only allows certain characters and a certain length.  We have code for that in SUSHI [here](https://github.com/FHIR/sushi/blob/master/src/fhirtypes/common.ts#L650-L679).

I guess the upside to always generating the id using a specific pattern (like `id-${count}`) is that you don't need to worry about the stuff above -- and it's easier for an author to find the ids _we_ generated if they can just search for `/id-\d+/`. So I could live with that.

jafeltra: Oh this is a really good catch, thanks Nick! I think it seems feasible to do the smarter `id` setting based on `name` since it looks like we have the functions/types/tools to do this, but now I am thinking that the benefits of a consistent pattern may outweigh that "smarter" approach. This code should really only be reached if we were given invalid FHIR to begin with, which will hopefully never happen, but already has. If we just always set `id` to our pattern, it would make it clear which ids were generated. Another potential approach could be to make the generated id more clearly a GoFSH generated value, like `GOFSH-GENERATED-ID-${count}`. So I think I'm leaning toward that, but happy to do the smarter `name`-based generation if @ngfreiter or @cmoesel prefer it.

cmoesel: I like `GOFSH-GENERATED-ID-${count}`.  Just to be clear, however, `id` is not required on many things, so lack of `id` doesn't necessarily indicate invalid FHIR.  But lack of `name` of a `StructureDefinition` _is_ invalid FHIR.

jafeltra: Hm I wonder if I got confused after all the `name` on `StructureDefinition` thoughts from last week. But now I'm kind of reconsidering... everything. (oops). If `id` isn't required on all definitions, should I even be adding it? The root problem that sparked this PR was that `id` was `undefined` on multiple resources of the same type, so they were getting marked as duplicate and not getting processed. Is the better thing to do just ensure the `id` isn't `null` or `undefined` before checking for a duplicate to filter out? That in itself would be pretty simple, but we would hit the [check for definitions after processing](https://github.com/FHIR/GoFSH/blob/master/src/processor/Package.ts#L77) more often, but at least those resources are output.

cmoesel: Oh.  Hmmm... Yeah, I think that reminds me that the reason we tried to set the `id` to the `name` (where applicable) is because if we do that, then the output FSH can suppress the `Id:` definition since SUSHI defaults `id` to the `name` anyway.  If we replace all missing ids w/ generated `id`s, then all of those will end up w/ `Id:` metadata in the FSH definitions -- and could get ugly.

So if we don't count `null`/`undefined` as a duplicate `id`, what happens after that?  Are there any other bugs/issues that arise in the GoFSH generation or no?  If not, then I might be leaning the same direction as you are, @jafeltra.

Wow, what fun we are having!

jafeltra: Such fun!

The biggest change that I see when running GoFSH on some test files when using the generated `id` vs just not counting `undefined`/`null` as a duplicate `id` are a bunch of errors logged after processing that we found a resource with a duplicate name that we cannot de-duplicate. So for example, I get various errors like this:

```
error Encountered instance with a duplicate name, undefined-of-ConceptMap, which GoFSH cannot make unique. Fix the source file to resolve this error or update the resulting FSH definition.
```
These get logged in [this function](https://github.com/FHIR/GoFSH/blob/master/src/processor/Package.ts#L77) after processing, but we made the decision to add them to the package anyway, so they still get written to the output but with a funky looking name, like `undefined-of-ConceptMap`. As this branch is now, that name would either be the `name` or the `id-${count}` that we add on.

ngfreiter: I think we need to add `id`s to instances that don't have them, because otherwise how do we know what we are going to call that Instance when we write the FSH? My understanding is that SUSHI always generates Instances with an `id`, so no matter what we do the round tripped result won't match the original. So I vote for just generating an `id` and making it obvious it is generated.

cmoesel: `undefined-of-ConceptMap` is kind of ugly, so I agree that we should generate `id`s on instances then -- but I still feel like generating `id`s on SDs, VSes, and CSes might be overstepping where we do not need to overstep (since SUSHI already has an algorithm to generate the `id` from a `name` for those cases).

jafeltra: Okay, I believe that 36a9d21 makes the changes we discussed at stand up, but to summarize:

- We only try to set `id`s on Instances now. This means we only consider definitions that are not going to result in FSH Profiles, Extensions, CodeSystems, or ValueSets. Since there are some `CodeSystem` and `ValueSet` definitions that we do not support creating FSH entities for, we have to check if those definitions will be written as instances or their special FSH entity.
- If we are dealing with a definition that will be a FSH Instance, we only try to set `id` based on `name` if it is a Conformance or Terminology resource. If there is a `name`, we use that as the `id` but we change it slightly to make sure it is a valid `id`.
- If we are dealing with any other type of FSH Instance, we don't try to be smart about setting `id` and just use a GoFSH generated id with a counter.
- Once we've set `id`s on everything that is relevant to have an `id`, we can more safely exclude definitions that do not have an `id` from the duplicate definition check. This lets us process `StructureDefinitions`, `ValueSets`, and `CodeSystems` that have a `name` but no `id`.

I did have to add a handful of test fixtures that have slight variations from each other and from other existing fixtures. If anyone is bothered by having files full of slight changes and knows a way to test these cases without making those files, let me know.

ngfreiter (approved):
These changes make sense to me!

mint-thompson (approved):

PR #99, opened 2021-02-18
Support spaces in ConceptRule codes

### LINKED ISSUE #98: GoFSH does not properly handle codes with spaces in the CodeSystem
Reported on zulip: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Codes.20with.20spaces

jafeltra:
This PR fixes #98 and properly quotes codes that have spaces in ConceptRules. This was the only case that I could find that the Zulip thread might be referencing referencing, so I think this should resolve it.

To test this out, you can add any concept rule that has a `code` with spaces. An example of what I used is below:

```
{
  "resourceType": "CodeSystem",
  "status": "active",
  "content": "complete",
  "name": "MyCodeSystem",
  "id": "MyCodeSystem",
  "version": "0.1.0",
  "url": "http://example.org/CodeSystem/MyCodeSystem",
  "concept": [
    {
      "code": "foo bar",
      "display": "Foo",
      "definition": "Foo"
    }
  ],
  "count": 1
}
```

### DIFF
--- a/src/exportable/ExportableConceptRule.ts
+++ b/src/exportable/ExportableConceptRule.ts
@@ -8,7 +8,7 @@ export class ExportableConceptRule extends fshrules.ConceptRule implements Expor
   }
 
   toFSH(): string {
-    let line = `* #${this.code}`;
+    let line = `* #${this.code.includes(' ') ? `"${this.code}"` : this.code}`;
     if (this.display) {
       line += ` "${fshifyString(this.display)}"`;
     }
--- a/src/exportable/ExportableConceptRule.ts
+++ b/src/exportable/ExportableConceptRule.ts
@@ -8,7 +8,7 @@ export class ExportableConceptRule extends fshrules.ConceptRule implements Expor
   }
 
   toFSH(): string {
-    let line = `* #${this.code.includes(' ') ? `"${this.code}"` : this.code}`;
+    let line = `* #${/\s/.test(this.code) ? `"${this.code}"` : this.code}`;
     if (this.display) {
       line += ` "${fshifyString(this.display)}"`;
     }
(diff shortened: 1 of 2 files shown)


julianxcarter (dismissed):

src/exportable/ExportableConceptRule.ts
```diff
@@ -8,7 +8,7 @@ export class ExportableConceptRule extends fshrules.ConceptRule implements Expor
   }
 
   toFSH(): string {
-    let line = `* #${this.code}`;
+    let line = `* #${this.code.includes(' ') ? `"${this.code}"` : this.code}`;
```
cmoesel, line 11: This works for normal space, but should we also account for tabs or other whitespace characters?  I think maybe we should -- in which case I'd check for a match on `/\s/`.

cmoesel (changes requested):
I'd suggest we should handle other types of whitespace as well.  I think we can pretty easily do that via regular expressions.

src/exportable/ExportableConceptRule.ts
jafeltra: That's a good idea! I was going off of the [FSHCode's toString](https://github.com/FHIR/sushi/blob/master/src/fshtypes/FshCode.ts#L10) in SUSHI, but I am now realizing that change was also me...

cmoesel: Ha.  Maybe.  Chances are I probably approved that other one. ;-)

cmoesel: It would probably be pretty bad for a code to have a tab, but you never know...

jafeltra: 4edae7d tests for `/\s/` when checking to use `" "`. I haven't yet made the equivalent PR for SUSHI, but I can if you'd like!

cmoesel (approved):
Love it.

cmoesel:
@julianxcarter -- any concerns w/ @jafeltra's updates or does this still have your approval?

julianxcarter (approved):
This looks even better to me now! My approval still stands, nice work @jafeltra 👍🏽

PR #100, opened 2021-02-23
Remove whitespace from entity names

### LINKED ISSUE #94: GoFSH allows names with spaces
Running GoFSH on [dk-core](https://github.com/hl7dk/dk-core) created a value set with spaces in the name.  As a result, SUSHI choked on it and the value set was incomplete in the results.  GoFSH should probably handle this.  Maybe detect it, issue a warning, replace spaces with `_`, and set the name back to original value via a caret rule?

mint-thompson:
Fixes #94 and completes task [CIMPL-664](https://standardhealthrecord.atlassian.net/browse/CIMPL-664)

An entity name in FSH cannot contain whitespace. Substitute whitespace with underscore characters so that the generated FSH is syntactically correct. When this substitution occurs, add a caret value rule to preserve the original name. Warn the user about these events.

### DIFF
--- a/src/processor/common.ts
+++ b/src/processor/common.ts
@@ -2,8 +2,13 @@ import { pullAt } from 'lodash';
 import {
   ExportableAssignmentRule,
   ExportableCaretValueRule,
-  ExportableSdRule
+  ExportableSdRule,
+  ExportableProfile,
+  ExportableExtension,
+  ExportableValueSet,
+  ExportableCodeSystem
 } from '../exportable';
+import { logger } from '../utils';
 
 // Places general Quantity-setting rules ahead of Quantity.unit setting rules
 export function switchQuantityRules(rules: ExportableSdRule[]): void {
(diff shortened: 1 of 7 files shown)


src/processor/common.ts
```diff
@@ -31,3 +36,26 @@ export function switchQuantityRules(rules: ExportableSdRule[]): void {
...
+    logger.warn(
+      `${entityType} with id ${entity.id} has name with whitespace. Converting whitespace to underscores.`
+    );
+    const nameRule = new ExportableCaretValueRule('');
+    nameRule.caretPath = 'name';
+    nameRule.value = entity.name;
+    entity.name = entity.name.replace(/\s/g, '_');
+    entity.rules.unshift(nameRule);
+  }
+}
```
cmoesel, line 61: Ha.  Yeah, you're right.  This is a lot simpler (and likely more elegant) than all those TypeScript shenanigans we were thinking of yesterday.  Good thinking!

cmoesel (approved):
I totally dig it.  Nice solution.

ngfreiter (approved):
Isn't a `name` with spaces on a `ValueSet`, `CodeSystem`, or `StructureDefinition` illegal FHIR? Looks good besides that, but just clarifying.

cmoesel:
> Isn't a `name` with spaces on a `ValueSet`, `CodeSystem`, or `StructureDefinition` illegal FHIR? Looks good besides that, but just clarifying.

It's not technically illegal, but it is discouraged.  The wording in the spec is that the name _should_ (not must) be computer-friendly.  And the invariant that checks it is marked as only `warning` (not `error`).

image

PR #101, opened 2021-02-24
Refactor for FSHOnline

julianxcarter:
Addresses [CIMPL-673](https://standardhealthrecord.atlassian.net/secure/RapidBoard.jspa?rapidView=9&modal=detail&selectedIssue=CIMPL-673&assignee=5f7dc49c287870006a40b275), removing uses of the `fs` module in the `loadOptimizers` function and also adding requisite exports to `src/index.ts` for use in FSHOnline. @jafeltra will be testing this out as she implements GoFSH in the browser, but this can also be tested by just ensuring that the optimizers and GoFSH as a whole still work as expected.

### DIFF
--- a/src/optimizer/loadOptimizers.ts
+++ b/src/optimizer/loadOptimizers.ts
@@ -21,6 +21,7 @@ export async function loadOptimizers(
   }
 
   // Import optimizers from the specified folder
+  // relativePath is placed in a dynamic string to allow for FSHOnline compatibility
   const Optimizers: { property: OptimizerPlugin } = await import(`${relativePath}`);
 
   const optimizers = Object.values(Optimizers).filter(
--- a/src/optimizer/loadOptimizers.ts
+++ b/src/optimizer/loadOptimizers.ts
@@ -20,7 +20,9 @@ export async function loadOptimizers(
     relativePath = `./${relativePath}`;
   }
 
+  // Import optimizers from the specified folder
   const Optimizers: { property: OptimizerPlugin } = await import(relativePath);
+
   const optimizers = Object.values(Optimizers).filter(
     // Remove non-optimizers
     o =>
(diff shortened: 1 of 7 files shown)


src/optimizer/loadOptimizers.ts
```diff
@@ -23,30 +20,17 @@ export async function loadOptimizers(
...
 
-  // read the plugins folder, filter to only .ts/.js files, and map to unique module paths (without extensions)
-  const optimizerModules = uniq(
-    fs
-      .readdirSync(folder, { withFileTypes: true })
-      .filter(f => f.isFile() && /^[^.]+.*\.(t|j)s$/.test(f.name))
-      .map(f => `${relativePath}/${f.name.match(/^([^.]+)/)[1]}`)
-  );
+  // Import optimizers from the specified folder
+  const Optimizers: { property: OptimizerPlugin } = await import(relativePath);
```
jafeltra, line 24: I haven't fully reviewed the changes here and I don't fully understand this, but I figured I'd post it in case anyone knew before I looked too much into it. But when I `npm pack` this branch and use it in FSH Online, this line gives me an error: 

```
Uncaught (in promise) Error: Cannot find module './plugins'
```

It appears that it doesn't like the variable passed into `import`. If this was a string path passed directly into `import`, it seems to work just fine. I'm not exactly sure why or what the right change is, but figured I'd at least post this for now!

julianxcarter: Commenting here for documentation purposes. As [noted in webpack's documentation](https://webpack.js.org/api/module-methods/#dynamic-expressions-in-import), performing imports with a variable containing the target path are not allowed. You __can__ perform imports from a dynamic string though, which is the workaround that @jafeltra and I settled on!

jafeltra (approved):
This look good and it works when I locally link it in FSH Online, which is really exciting!

mint-thompson (approved):

PR #103, opened 2021-03-11
GoFSH API map option is now JSON.stringify-able

julianxcarter:
Addresses [CIMPL-685](https://standardhealthrecord.atlassian.net/browse/CIMPL-685), making it so that the object returned by the `map` option of the GoFSH api can be serialized by the `JSON.stringify()` method. As it turns out, `JSON.stringify` is actually incapable of parsing maps, so this is done by creating a child class of the `Map` interface with a `toJSON` method that basically converts the Map to a regular object (credit to @cmoesel for coming up with the idea to this approach). A simple test case was added to verify that running `JSON.stringify()` does in fact convert each resource's map into a JSON object.

### DIFF
--- a/src/api/FhirToFsh.ts
+++ b/src/api/FhirToFsh.ts
@@ -93,14 +93,25 @@ export async function fhirToFsh(
 
 export type fshMap = {
   aliases: string;
-  invariants: Map<string, string>;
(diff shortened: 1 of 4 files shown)


jafeltra (approved):
Looks good to me!

PR #105, opened 2021-03-17
Optimize Quantity values based on new FSH STU2 syntax

mint-thompson:
Completes task [CIMPL-698](https://standardhealthrecord.atlassian.net/browse/CIMPL-698).

New FSH syntax for Quantity values in assignment and caret value rules allow for more elements to be combined into a single rule. The system on the Quantity is no longer required to be UCUM. The Quantity can now contain the display string in its unit element. These changes allow the ExportableAssignmentRule and ExportableCaretValueRule to take advantage of the new implementation of FshQuantity.toString without any further changes. As such, no tests are added for those two classes.

Overall, the removal of the requirement for UCUM makes this implementation a lot simpler. Hooray!

### DIFF
--- a/test/extractor/AssignmentRuleExtractor.test.ts
+++ b/test/extractor/AssignmentRuleExtractor.test.ts
@@ -179,17 +179,13 @@ describe('AssignmentRuleExtractor', () => {
       element.patternQuantity.system = 'http://other-units.org';
       element.patternQuantity.unit = 'Gigawatts';
       const assignmentRules = AssignmentRuleExtractor.process(element);
-      const expectedQuantityRule = new ExportableAssignmentRule('valueQuantity');
-      expectedQuantityRule.value = new fshtypes.FshCode(
-        'GW',
-        'http://other-units.org',
-        'Gigawatts'
+      const expectedRule = new ExportableAssignmentRule('valueQuantity');
+      expectedRule.value = new fshtypes.FshQuantity(
+        1.21,
+        new fshtypes.FshCode('GW', 'http://other-units.org', 'Gigawatts')
       );
-      const expectedValueRule = new ExportableAssignmentRule('valueQuantity.value');
-      expectedValueRule.value = 1.21;
-      expect(assignmentRules).toHaveLength(2);
-      expect(assignmentRules[0]).toEqual<ExportableAssignmentRule>(expectedQuantityRule);
-      expect(assignmentRules[1]).toEqual<ExportableAssignmentRule>(expectedValueRule);
+      expect(assignmentRules).toHaveLength(1);
+      expect(assignmentRules[0]).toEqual<ExportableAssignmentRule>(expectedRule);
       expect(element.processedPaths).toHaveLength(4);
       expect(element.processedPaths).toContain('patternQuantity.value');
       expect(element.processedPaths).toContain('patternQuantity.code');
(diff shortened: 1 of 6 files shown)


ngfreiter (approved):
This looks good to me!

jafeltra (commented):
These changes make a lot of sense to me. I did have one question. I noticed we also do special handling in the `AssignmentRuleExtractor` [here](https://github.com/FHIR/GoFSH/blob/master/src/extractor/AssignmentRuleExtractor.ts#L136) for building quantities for UCUM codes. I'm wondering if there's a case that we could build out the individual rules rather than use the FshQuantities. I'm thinking maybe if we had an assignment rule on a profile, we might not update those rules in the `CombineCodingAndQuantityValueOptimizer`, although I'm a little confused, so I might be missing something.

mint-thompson:
> These changes make a lot of sense to me. I did have one question. I noticed we also do special handling in the `AssignmentRuleExtractor` [here](https://github.com/FHIR/GoFSH/blob/master/src/extractor/AssignmentRuleExtractor.ts#L136) for building quantities. I'm wondering if there's a case that we could build out the individual rules rather than use the FshQuantities. I'm thinking maybe if we had an assignment rule on a profile, we might not update those rules in the `CombineCodingAndQuantityValueOptimizer`, although I'm a little confused, so I might be missing something.

You're absolutely right! Since the requirement for UCUM is no longer around, there's no reason this extractor should be checking for that. I think updating that extractor should be part of these changes as well, since like you said, the optimizer won't check assignment rules on profiles or extensions.

jafeltra (commented):
Thanks for making that change @mint-thompson! It looks good to me! Unfortunately, I just thought of another question. I just remembered we added in the `switchQuantityRules` called [here](https://github.com/FHIR/GoFSH/blob/master/src/processor/InstanceProcessor.ts#L89) as part of #71. Is this necessary with your PR still? Or is it safer to just keep it since it probably isn't affecting much any more?

test/extractor/AssignmentRuleExtractor.test.ts
```diff
@@ -179,17 +179,13 @@ describe('AssignmentRuleExtractor', () => {
...
+        1.21,
+        new fshtypes.FshCode('GW', 'http://other-units.org', 'Gigawatts')
       );
-      const expectedValueRule = new ExportableAssignmentRule('valueQuantity.value');
-      expectedValueRule.value = 1.21;
-      expect(assignmentRules).toHaveLength(2);
-      expect(assignmentRules[0]).toEqual<ExportableAssignmentRule>(expectedQuantityRule);
-      expect(assignmentRules[1]).toEqual<ExportableAssignmentRule>(expectedValueRule);
+      expect(assignmentRules).toHaveLength(1);
+      expect(assignmentRules[0]).toEqual<ExportableAssignmentRule>(expectedRule);
```
ngfreiter, line 188: Since we now don't special case UCUM, I don't think we need this test at all anymore, since it just tests the same path of code as the test above.

test/extractor/AssignmentRuleExtractor.test.ts
```diff
@@ -223,23 +219,16 @@ describe('AssignmentRuleExtractor', () => {
...
+          new fshtypes.FshCode('cm', 'http://other-units.org', 'centimeters')
+        ),
+        new fshtypes.FshQuantity(1, new fshtypes.FshCode('s', 'http://unitsofmeasure.org'))
       );
-      expect(assignmentRules).toHaveLength(3);
-      expect(assignmentRules[0]).toEqual<ExportableAssignmentRule>(expectedNumeratorQuantity);
-      expect(assignmentRules[1]).toEqual<ExportableAssignmentRule>(expectedNumeratorValue);
-      expect(assignmentRules[2]).toEqual<ExportableAssignmentRule>(expectedDenominatorRule);
+      expect(assignmentRules).toHaveLength(1);
+      expect(assignmentRules[0]).toEqual<ExportableAssignmentRule>(expectedRule);
```
ngfreiter, line 231: Similar to the test I commented on above, I'm not sure we really need this test.

test/extractor/AssignmentRuleExtractor.test.ts
```diff
@@ -291,27 +270,16 @@ describe('AssignmentRuleExtractor', () => {
...
-        's',
-        'http://other-units.org',
-        'seconds'
+      const expectedRule = new ExportableAssignmentRule('valueRatio');
+      expectedRule.value = new fshtypes.FshRatio(
+        new fshtypes.FshQuantity(
+          5,
+          new fshtypes.FshCode('cm', 'http://other-units.org', 'centimeters')
+        ),
+        new fshtypes.FshQuantity(1, new fshtypes.FshCode('s', 'http://other-units.org', 'seconds'))
```
ngfreiter, line 279: Same comment goes here.

ngfreiter (commented):
I think the tests that were updated in the latest commit could probably just be removed. Since UCUM is no longer a special case in the extractor, I'm not sure those tests really test anything unique. But if you want to leave them in to test the fact that UCUM isn't a special case, that's also fine with me.

mint-thompson:
@ngfreiter I left those test cases in as acknowledgement that UCUM used to be a special case, but that might not really be necessary. If the test cases just use a variety of values for `system`, I feel like that's probably sufficient to demonstrate that the value of `system` does not affect processing.

mint-thompson:
> Thanks for making that change @mint-thompson! It looks good to me! Unfortunately, I just thought of another question. I just remembered we added in the `switchQuantityRules` called [here](https://github.com/FHIR/GoFSH/blob/master/src/processor/InstanceProcessor.ts#L89) as part of #71. Is this necessary with your PR still? Or is it safer to just keep it since it probably isn't affecting much any more?

Removing it doesn't break any tests, so I'm willing to say that it's no longer necessary. The call to `switchQuantityRules` in `StructureDefinitionProcessor` is still needed, though.

jafeltra:
> Removing it doesn't break any tests, so I'm willing to say that it's no longer necessary. The call to `switchQuantityRules` in `StructureDefinitionProcessor` is still needed, though.

Hmm maybe if we need to keep it in `StructureDefinitionProcessor` we should just keep it in the `InstanceProcessor` for simplicity/clarity. I'm okay either way, so whichever you prefer, I'm good with!

mint-thompson:
> > Removing it doesn't break any tests, so I'm willing to say that it's no longer necessary. The call to `switchQuantityRules` in `StructureDefinitionProcessor` is still needed, though.
> 
> Hmm maybe if we need to keep it in `StructureDefinitionProcessor` we should just keep it in the `InstanceProcessor` for simplicity/clarity. I'm okay either way, so whichever you prefer, I'm good with!

Right after making my previous post, I realized it was a little odd for it to still be required in `StructureDefinitionProcessor`. I think the actual answer is: we had a test case for `StructureDefinitionProcessor` that specifically checked for the effects of `switchQuantityRules`, but we didn't have that for `InstanceProcessor`. But, if we keep in both calls, that means that if for whatever reason, the optimizer isn't used, you'll still get correct FSH. So, maybe the right decision is to keep the code in, and just remove the TODO comment? We should probably discuss this with more folks today.

jafeltra:
> But, if we keep in both calls, that means that if for whatever reason, the optimizer isn't used, you'll still get correct FSH. So, maybe the right decision is to keep the code in, and just remove the TODO comment?

I'm a fan! I like this approach and I think it makes the most sense!

jafeltra (approved):
These changes make sense to me! Thanks @mint-thompson!

ngfreiter (approved):
Jealous of all those code deletions. Looks good to me!

PR #106, opened 2021-03-19
Optimize rule paths to use Soft Indexing

julianxcarter:
Completes [CIMPL-699](https://standardhealthrecord.atlassian.net/browse/CIMPL-699), creating the `SimplifyArrayIndexingOptimizer`. 

This optimizer removes `[0]` and `[=]` indices from elements referencing singleton arrays. This optimizer also converts numeric indices to soft indexing where appropriate, meaning only when an array's index remains the same as in the previous rule or is incremented by one. If an array includes gaps in its element references, then soft indexing is eschewed. I thought it would be most intuitive to new users to leave the _first_ index in a soft indexing sequence numeric and use soft indexing for all subsequent indexing, e.g: 

```
* name[0].given = Julian0
* name[0].family = Carter0
* name[1].given = Julian1
* name[1].family = Carter1
```

would convert to:
```
* name[0].given = Julian0
* name[=].family = Carter0
* name[+].given = Julian1
* name[=].family = Carter1
```

This way, if a user is new to FSH then it'll be a bit easier for them to figure out what's going on with the indices.

### DIFF
--- a/src/optimizer/loadOptimizers.ts
+++ b/src/optimizer/loadOptimizers.ts
@@ -39,8 +39,11 @@ export async function loadOptimizers(
   optimizers.forEach(opt => {
     nodes.push(opt.name);
     opt.runAfter?.forEach(dependsOn => {
-      if (optimizers.some(o => o.name === dependsOn)) {
-        edges.push([opt.name, dependsOn]);
+      const preceedingOptimizer = optimizers.find(o =>
+        dependsOn instanceof RegExp ? dependsOn.test(o.name) : dependsOn === o.name
+      );
+      if (preceedingOptimizer) {
+        edges.push([opt.name, preceedingOptimizer.name]);
       } else {
         logger.error(
           `The ${opt.name} optimizer specifies an unknown optimizer in runAfter: ${dependsOn}`
@@ -48,8 +51,11 @@ export async function loadOptimizers(
       }
     });
     opt.runBefore?.forEach(dependedOnBy => {
-      if (optimizers.some(o => o.name === dependedOnBy)) {
-        edges.push([dependedOnBy, opt.name]);
+      const succeedingOptimizer = optimizers.find(o =>
+        dependedOnBy instanceof RegExp ? dependedOnBy.test(o.name) : dependedOnBy === o.name
+      );
+      if (succeedingOptimizer) {
+        edges.push([succeedingOptimizer.name, opt.name]);
       } else {
         logger.error(
           `The ${opt.name} optimizer specifies an unknown optimizer in runBefore: ${dependedOnBy}`
--- a/src/optimizer/loadOptimizers.ts
+++ b/src/optimizer/loadOptimizers.ts
@@ -74,7 +74,6 @@ export async function loadOptimizers(
   let ordered: string[];
   try {
     ordered = toposort.array(nodes, edges).reverse();
-    console.log(JSON.stringify(ordered));
   } catch (e) {
     // This message should be reliably present and reliably in this format, but use '?' defensively just in case
     const nodeMatch = e.message?.match(/"([^"]+)"$/);
--- a/src/optimizer/loadOptimizers.ts
+++ b/src/optimizer/loadOptimizers.ts
@@ -39,11 +39,15 @@ export async function loadOptimizers(
   optimizers.forEach(opt => {
     nodes.push(opt.name);
     opt.runAfter?.forEach(dependsOn => {
-      const preceedingOptimizer = optimizers.find(o =>
-        dependsOn instanceof RegExp ? dependsOn.test(o.name) : dependsOn === o.name
+      const preceedingOptimizers = optimizers.filter(o =>
+        o.name !== opt.name && dependsOn instanceof RegExp
+          ? dependsOn.test(o.name)
+          : dependsOn === o.name
       );
-      if (preceedingOptimizer) {
-        edges.push([opt.name, preceedingOptimizer.name]);
+      if (preceedingOptimizers.length > 0) {
+        preceedingOptimizers.forEach(preceedingOptimizer => {
+          edges.push([opt.name, preceedingOptimizer.name]);
+        });
       } else {
         logger.error(
           `The ${opt.name} optimizer specifies an unknown optimizer in runAfter: ${dependsOn}`
@@ -51,11 +55,15 @@ export async function loadOptimizers(
       }
     });
     opt.runBefore?.forEach(dependedOnBy => {
-      const succeedingOptimizer = optimizers.find(o =>
-        dependedOnBy instanceof RegExp ? dependedOnBy.test(o.name) : dependedOnBy === o.name
+      const succeedingOptimizers = optimizers.filter(o =>
+        o.name !== opt.name && dependedOnBy instanceof RegExp
+          ? dependedOnBy.test(o.name)
+          : dependedOnBy === o.name
       );
-      if (succeedingOptimizer) {
-        edges.push([succeedingOptimizer.name, opt.name]);
+      if (succeedingOptimizers.length > 0) {
+        succeedingOptimizers.forEach(succeedingOptimizer => {
+          edges.push([succeedingOptimizer.name, opt.name]);
+        });
       } else {
         logger.error(
           `The ${opt.name} optimizer specifies an unknown optimizer in runBefore: ${dependedOnBy}`
@@ -66,6 +74,7 @@ export async function loadOptimizers(
   let ordered: string[];
   try {
     ordered = toposort.array(nodes, edges).reverse();
+    console.log(JSON.stringify(ordered));
   } catch (e) {
     // This message should be reliably present and reliably in this format, but use '?' defensively just in case
     const nodeMatch = e.message?.match(/"([^"]+)"$/);
--- a/src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
+++ b/src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
@@ -3,8 +3,6 @@ import { OptimizerPlugin } from '../OptimizerPlugin';
 import { Package } from '../../processor';
 import { ExportableCaretValueRule } from '../../exportable';
 
-const singletonArrayElements: Map<string, fhirtypes.PathPart[]> = new Map();
-
 export default {
   name: 'simplify_array_indexing',
   description: 'Replace numeric indices with soft indexing',
@@ -23,7 +21,8 @@ export default {
       const pathMap: Map<string, number> = new Map();
       const caretPathMap: Map<string, Map<string, number>> = new Map();
       const ruleArr: fshrules.Rule[] = def.rules;
-      const parsedRules = ruleArr.map((rule: fshrules.Rule) => {
+      const singletonArrayElements: Map<string, fhirtypes.PathPart[]> = new Map();
+      const parsedPaths = ruleArr.map((rule: fshrules.Rule) => {
         const parsedPath: { path: fhirtypes.PathPart[]; caretPath?: fhirtypes.PathPart[] } = {
           path: utils.parseFSHPath(rule.path)
         };
@@ -33,16 +32,15 @@ export default {
         }
         return parsedPath;
       });
+      addPrefixes(parsedPaths);
 
-      parsedRules.forEach((parsedRule: any, ruleIndex: number) => {
+      parsedPaths.forEach((parsedRule: any, ruleIndex: number) => {
         const originalRule = def.rules[ruleIndex];
-        parsedRule.path.forEach((element: fhirtypes.PathPart, elementIndex: number) => {
-          // Add a prefix to the current element containing previously parsed rule elements
-          element.prefix = utils.assembleFSHPath(parsedRule.path.slice(0, elementIndex));
-          applySoftIndexing(element, pathMap);
+        parsedRule.path.forEach((element: fhirtypes.PathPart) => {
+          applySoftIndexing(element, pathMap, singletonArrayElements);
         });
 
-        parsedRule.caretPath?.forEach((element: fhirtypes.PathPart, elementIndex: number) => {
+        parsedRule.caretPath?.forEach((element: fhirtypes.PathPart) => {
           // Caret path indexes should only be resolved in the context of a specific path
           // Each normal path has a separate map to keep track of the caret path indexes
           if (!caretPathMap.has(originalRule.path)) {
@@ -50,31 +48,48 @@ export default {
           }
 
           const elementCaretPathMap = caretPathMap.get(originalRule.path);
-          // Add a prefix to the current element containing previously parsed rule elements
-          element.prefix = utils.assembleFSHPath(parsedRule.caretPath.slice(0, elementIndex));
-          applySoftIndexing(element, elementCaretPathMap);
+          applySoftIndexing(element, elementCaretPathMap, singletonArrayElements);
         });
       });
 
-      removeZeroIndices();
+      removeZeroIndices(singletonArrayElements);
 
-      parsedRules.forEach((parsedRule: any, ruleIndex: number) => {
+      parsedPaths.forEach((parsedRule: any, ruleIndex: number) => {
         const originalRule = def.rules[ruleIndex];
         originalRule.path = utils.assembleFSHPath(parsedRule.path);
         if (originalRule instanceof ExportableCaretValueRule) {
           originalRule.caretPath = utils.assembleFSHPath(parsedRule.caretPath);
         }
       });
+      singletonArrayElements.clear();
     });
   }
 } as OptimizerPlugin;
+
+function addPrefixes(rules: any[]) {
+  rules.forEach((parsedPath: any) => {
+    parsedPath.path.forEach((element: fhirtypes.PathPart, elementIndex: number) => {
+      // Add a prefix to the current element containing previously parsed rule elements
+      element.prefix = utils.assembleFSHPath(parsedPath.path.slice(0, elementIndex));
+    });
+    parsedPath.caretPath?.forEach((element: fhirtypes.PathPart, elementIndex: number) => {
+      // Add a prefix to the current element containing previously parsed rule elements
+      element.prefix = utils.assembleFSHPath(parsedPath.path.slice(0, elementIndex));
+    });
+  });
+}
 
 /**
  *
  * @param {PathPart} element - A single element in a rules path
  * @param {Map<string, number} pathMap - A map containing an element's name as the key and that element's updated index as the value
+ * @param {Map<string, PathPart[]} singletonArrayElements - A map containing a
  */
-function applySoftIndexing(element: fhirtypes.PathPart, pathMap: Map<string, number>) {
+function applySoftIndexing(
+  element: fhirtypes.PathPart,
+  pathMap: Map<string, number>,
+  singletonArrayElements: Map<string, fhirtypes.PathPart[]>
+) {
   // Must account for a pathPart's base name, prior portions of the path, as well as any slices it's contained in.
   const mapName = `${element.prefix ?? ''}.${element.base}|${(element.slices ?? []).join('|')}`;
   const indexRegex = /^[0-9]+$/;
@@ -87,11 +102,15 @@ function applySoftIndexing(element: fhirtypes.PathPart, pathMap: Map<string, num
   if (!pathMap.has(mapName)) {
     pathMap.set(mapName, parseInt(currentNumericBracket));
     // If the element contains a zero Index, we'll flag the rule to have the index removed
-    if (currentNumericBracket === '0') singletonArrayElements.set(mapName, [element]);
+    if (currentNumericBracket === '0') {
+      singletonArrayElements.set(mapName, [element]);
+    }
   } else {
     const previousNumericIndex = pathMap.get(mapName);
     if (parseInt(currentNumericBracket) === previousNumericIndex + 1) {
-      if (previousNumericIndex === 0) singletonArrayElements.delete(mapName);
+      if (previousNumericIndex === 0) {
+        singletonArrayElements.delete(mapName);
+      }
       element.brackets[bracketIndex] = '+';
       pathMap.set(mapName, previousNumericIndex + 1);
     } else if (parseInt(currentNumericBracket) === previousNumericIndex) {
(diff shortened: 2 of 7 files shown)


src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
```diff
@@ -0,0 +1,123 @@
+import { fshrules, fhirtypes, utils } from 'fsh-sushi';
+import { OptimizerPlugin } from '../OptimizerPlugin';
+import { Package } from '../../processor';
+import { ExportableCaretValueRule } from '../../exportable';
+
+const singletonArrayElements: Map<string, fhirtypes.PathPart[]> = new Map();
```
ngfreiter, line 6: I'm not sure I entirely understand why this needs to be a global variable. Doesn't that mean it will carry information between optimizations of different resources? Is that something we need?

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
```diff
@@ -0,0 +1,123 @@
...
+  const bracketIndex = element.brackets.indexOf(currentNumericBracket);
+
+  if (!pathMap.has(mapName)) {
+    pathMap.set(mapName, parseInt(currentNumericBracket));
+    // If the element contains a zero Index, we'll flag the rule to have the index removed
+    if (currentNumericBracket === '0') singletonArrayElements.set(mapName, [element]);
+  } else {
+    const previousNumericIndex = pathMap.get(mapName);
+    if (parseInt(currentNumericBracket) === previousNumericIndex + 1) {
+      if (previousNumericIndex === 0) singletonArrayElements.delete(mapName);
```
mint-thompson, line 94: The single-line style for `if` statements is a little bit tricky to read, especially in a position where it is right after another `if` statement and followed by additional statements. Could you put braces around all of your `if`-blocks?

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
```diff
@@ -0,0 +1,123 @@
...
+import { OptimizerPlugin } from '../OptimizerPlugin';
+import { Package } from '../../processor';
+import { ExportableCaretValueRule } from '../../exportable';
+
+const singletonArrayElements: Map<string, fhirtypes.PathPart[]> = new Map();
+
+export default {
+  name: 'simplify_array_indexing',
+  description: 'Replace numeric indices with soft indexing',
+  optimize(pkg: Package): void {
```
mint-thompson, line 11: I think it might be necessary to declare a `runsAfter` for this extension. Specifically, I think it needs to run after `CombineCodingAndQuantityValuesOptimizer`. That optimizer relies on rule path equality by simple string comparison, and does not attempt to perform path calculations that involve soft indices. So, if the soft indexing were applied before that optimizer, it would be unable to combine any rules.

Thinking about this scenario more generally, this optimizer should probably be run near the end. Otherwise, we'd need to change any optimizer that modifies the list of rules to be able to handle path comparisons involving soft indices. Fortunately, we've done a good job with naming our optimizers, so it looks like this would need to run after the combiners:
- combine_card_and_flag_rules
- combine_coding_and_quantity_values
- combine_contains_rules
and removers:
- remove_choice_slicing_rules
- remove_default_extension_context_rules
- remove_extension_slicing_rules
- remove_extension_url_assignment_rules
- remove_generated_text_rules
- remove_implied_zero_zero_card_rules
- remove_publisher_derived_date_rules

I'm not 100% that every optimizer in those lists can interact unfavorably with soft indices, but it seems like a safe bet to include all of those as preconditions for applying soft indexing.

mint-thompson (changes requested):

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
```diff
@@ -0,0 +1,123 @@
...
+    pathMap.set(mapName, parseInt(currentNumericBracket));
+    // If the element contains a zero Index, we'll flag the rule to have the index removed
+    if (currentNumericBracket === '0') singletonArrayElements.set(mapName, [element]);
+  } else {
+    const previousNumericIndex = pathMap.get(mapName);
+    if (parseInt(currentNumericBracket) === previousNumericIndex + 1) {
+      if (previousNumericIndex === 0) singletonArrayElements.delete(mapName);
+      element.brackets[bracketIndex] = '+';
+      pathMap.set(mapName, previousNumericIndex + 1);
+    } else if (parseInt(currentNumericBracket) === previousNumericIndex) {
```
ngfreiter, line 97: I think I am running into some tricky issues with how this `pathMap` setup deals with nested paths. I've been trying it out on the US Core Client Capability Statement resource. The first time we see this path:
```
* rest[0].resource[0].supportedProfile[0] = "http://hl7.org/fhir/us/core/StructureDefinition/us-core-allergyintolerance"
```
It is correctly translated into:
```
* rest.resource[=].supportedProfile[0] = "http://hl7.org/fhir/us/core/StructureDefinition/us-core-allergyintolerance"
```
But then the second time we run into a path ending in `supportedProfile`:
```
* rest[0].resource[1].supportedProfile[0] = "http://hl7.org/fhir/us/core/StructureDefinition/us-core-careplan"
```
It gets translated into
```
* rest.resource[=].supportedProfile[=] = "http://hl7.org/fhir/us/core/StructureDefinition/us-core-careplan"
```
Which isn't really right, and causes SUSHI to emit an error like:
image

It is correct to assume the `0` index, but still we don't want to write FSH that produces a bunch of errors. From what I can tell this happens because the second time we hit a path ending in `supportedProfile`, we already have `rest[=].resource[=].supportedProfile|` in the `pathMap`. But _this_ path really has nothing to do with that original `supportedProfile`. The `pathMap` does not account for the fact that the index of `resource` has been incremented since the first time we saw `supportedProfile`. At least I think that is what is happening. If I am way off base on my explanation let me know. But this is definitely a bug, regardless of if I've explained it all wrong, since it is causing a lot of errors when I try to round trip US Core.

ngfreiter (changes requested):
Dang this is tricky, good work figuring out an approach, but I did find some issues that cropped up when round tripping US Core.

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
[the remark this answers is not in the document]
ngfreiter: Pst please ignore this because @mint-thompson made the same point before me.

cmoesel: The problem w/ `runBefore` and `runAfter` is that if we want something to run last, then we need to try to keep up as things get added.  @mint-thompson, @ngfreiter, @julianxcarter -- what do you think about adding a `runLast` boolean?  The only tricky thing is if there are multiple things labeled `runLast` -- but in that case the tie could be broken using `runBefore` and `runAfter`.  The only other thing I can think of would be supporting regex on `runBefore` and `runAfter`, so you could do something like: `runAfter: /^combine/` or even `runAfter: /.*/` (although that still could lead to ties).

mint-thompson: I like the regex approach because it should be able to work well with the existing ordering strategy we have implemented. Ties are fine as long as they are between optimizers that don't need a strict ordering. While `runLast` is convenient for this case, I think if we were to take that approach, I'd prefer to take it one step further and define subgroups for strict ordering, with partial ordering happening within each subgroup. But, maybe just two subgroups would be enough? In that case, the boolean makes sense.

I feel like it comes down to a choice of one level of sorting or more than one level of sorting, and my preference is to keep it at one level. Subgroups can naturally emerge from having just one level of sorting without having to explicitly define a subgroup.

julianxcarter: @cmoesel I really like the idea of supporting regex in `runBefore` and `runAfter`. That sounds like it could be pretty useful in other optimizers down the line, and it'll also allow us to effectively have a `runLast` option. If we don't build in an explicit `runLast` option though, then breaking ties could be more difficult. I'd be in favor of adding both.

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
julianxcarter: Yes I can!

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
cmoesel: Since you both like the approach of using regular expressions, let's go ahead and give that a whirl.  I think I agree with @mint-thompson that we want to try to keep this at one level -- so let's not add `runLast` just yet; let's see how far we can get w/ the regular expressions. It may be that they're enough for any actual use cases we'll ever have.

ngfreiter: Didn't see your comment @cmoesel until after I posted mine. Regexs work for me too.

cmoesel: Ah, OK.  I actually think that implementation of regex should be pretty simple.  It might just be changing the types like:
```
runBefore?: string[];
```
to:
```
runBefore?: (string|RegExp)[];
```

And then changing where we do the comparison (example below) to either do equality (for string) or test (for RegEx):
```
if (optimizers.some(o => o.name === dependsOn)) {
```

I think it might be even easier than adding a third field (`runLast`) to the mix.

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
julianxcarter: Ahhhh I see what's going on here. Cases like this are supposed to be caught by adding the `prefix` property to each PathPart in a path. Problem was, this property was being added _after_ each property's numeric indexes were converted to soft indexes. This can create situations like what you ran into here, thanks for catching this! This should be addressed in the next commit.

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
```diff
@@ -0,0 +1,123 @@
...
+      ...pkg.extensions,
+      ...pkg.valueSets,
+      ...pkg.codeSystems,
+      ...pkg.instances,
+      ...pkg.mappings
+    ].forEach(def => {
+      const pathMap: Map<string, number> = new Map();
+      const caretPathMap: Map<string, Map<string, number>> = new Map();
+      const ruleArr: fshrules.Rule[] = def.rules;
+      const parsedRules = ruleArr.map((rule: fshrules.Rule) => {
```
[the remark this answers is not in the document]
julianxcarter: This is correct!

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
julianxcarter: Addressed in [`d6e165e`](https://github.com/FHIR/GoFSH/pull/106/commits/d6e165e5611b739e6fae61fe8ba1ca39183768ae).

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
julianxcarter: Added regex support in [`77ddd51`](https://github.com/FHIR/GoFSH/pull/106/commits/77ddd51820580042db41050654e61133fa74121b).

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
julianxcarter: Addressed in [`d6e165e`](https://github.com/FHIR/GoFSH/pull/106/commits/d6e165e5611b739e6fae61fe8ba1ca39183768ae).

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
julianxcarter: Addressed in [`d6e165e`](https://github.com/FHIR/GoFSH/pull/106/commits/d6e165e5611b739e6fae61fe8ba1ca39183768ae).

ngfreiter (changes requested):
I'm still noticing some bugs when I run this on US Core. For reference, I'm running:
```
ts-node src/app.ts -f ~/.fhir/packages/hl7.fhir.us.core\#3.1.0/package/
```
on this branch, and on master, and just comparing how things look going through a full round-trip. In terms of meaning, we want the two round-trips to be exactly the same. If some elements get re-ordered here and there, that's not an issue. But I'm seeing pretty significant regressions on this branch compared to master. As one example, we get SearchParameters which end up missing data, and not correctly having Soft Indexing applied. On this branch we generate the following FSH:
```
Instance: us-core-allergyintolerance-clinical-status
InstanceOf: SearchParameter
Usage: #definition
* modifier[0] = #missing
* modifier[+] = #text
* modifier[+] = #not
* modifier[+] = #in
* modifier[+] = #not-in
* modifier[+] = #below
* modifier[+] = #above
* modifier[+] = #ofType
* modifier[0].extension.url = "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation"
* modifier[0].extension.valueCode = #MAY
* modifier[1].extension.url = "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation"
* modifier[1].extension.valueCode = #MAY
* modifier[2].extension.url = "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation"
* modifier[2].extension.valueCode = #MAY
* modifier[3].extension.url = "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation"
* modifier[3].extension.valueCode = #MAY
* modifier[4].extension.url = "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation"
* modifier[4].extension.valueCode = #MAY
* modifier[5].extension.url = "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation"
* modifier[5].extension.valueCode = #MAY
* modifier[6].extension.url = "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation"
* modifier[6].extension.valueCode = #MAY
* modifier[=].extension.url = "http://hl7.org/fhir/StructureDefinition/capabilitystatement-expectation"
* modifier[=].extension.valueCode = #MAY
```
The first issue is that I think we could expect that the `modifier.extension` lines could use soft indexing, after the first one resets the context to the 0th element. But the real problem is that the `pathMap` doesn't reset its context once we see that 0th element. So the final element should have index `7`, and GoFSH recognizes that, and thinks that the most recently see `modifier` element was at index `7`, so it uses the `=`. But this is a problem, because the true most recently seen `modifier` element was at index `6`, on the previous line.

Something is also going on with the CapabilityStatements. I haven't dug deep enough to figure out exactly why, but they are really good tests cases of soft indexing, since they use so much soft indexing. The issue seems somewhat similar to what I detailed above. There are cases where the `=` sign is being incorrectly used. and we are overwriting previous elements.

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
```diff
@@ -0,0 +1,170 @@
...
+    }
+    const filteredElements = referencePathElements.filter(
+      e =>
+        e.base === element.base &&
+        e.prefix === element.prefix &&
+        e.brackets &&
+        !(e.brackets.includes('0') || e.brackets.includes('='))
+    );
+    if (filteredElements.length === 0) {
+      if (element.brackets.includes('0')) {
```
ngfreiter, line 137: Something is still going wrong with CapabilityStatements on US Core. I'm not entirely sure, but I think it is likely due to this line, since the error starts at index `10`, and this line is only looking for `0` to be included, which is included in `10`.

ngfreiter (commented):
I think there are still some issues with US Core. I'd definitely suggest using the command I listed above and double checking that everything works the same as on `master`, even once this bug is fixed it's possible there could be some more hiding out. If you want me to show you what I've been doing to test, I'm happy to, just let me know.

julianxcarter:
I've gotten to the root of what was going on here. It all has to do with [this line within sushi's parseFSHPath function](https://github.com/FHIR/sushi/blob/9dc8af003772f1398e8f01a5213f5813a95fb028/src/utils/PathUtils.ts#L14). The regex expression within that function is currently `/^[0-9]$/` when it should really be `^[0-9]+$`. This is a minor change that actually has some pretty big effects.

This regex statement is used within that function to determine if a PathPart's bracket is a slice name, or a numeric index. If a bracket is a slice name, we'd want to add it to that PathPart's `slices` array so that it can be taken into account during soft indexing conversion (e.g. we would want `item[Slice1][+]` and `item[Slice2][+]` to be indexed separately since they're two different arrays). Because the regex statement was missing that plus though, double digit indices were recognized as slice names and added to the slices array when they shouldn't have been. 

In regards to this optimizer, this resulted in incremented indices greater than 9 being printed as numbers rather than being converted to `+`s as one would expect in a soft indexing sequence. This is because `parseFSHPath` was adding the numeric index to each PathPart's `slices` array, which in turn made the `mapName` different for each succeeding element after the 9th element. The same problem occurred on the sushi side when this fsh was processed during a fishing trip, resulting in cascading problems.

I've put up a PR on the sushi side to fix this issue there, however in the meantime I've added a separate `parseFSHPath` helper function to this optimizer so that we don't have to wait until we release a new version of sushi to take advantage of this fix. When we release another version of sushi with that hot fix, we can go back and remove the duplicate `parseFSHPath` function from this optimizer.

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
julianxcarter: I've detailed what was going on here in a separate comment, this should be addressed in `ba71e9f`.

src/optimizer/loadOptimizers.ts
```diff
@@ -39,17 +39,23 @@ export async function loadOptimizers(
...
+      const preceedingOptimizer = optimizers.find(o =>
+        dependsOn instanceof RegExp ? dependsOn.test(o.name) : dependsOn === o.name
+      );
+      if (preceedingOptimizer) {
+        edges.push([opt.name, preceedingOptimizer.name]);
       } else {
         logger.error(
           `The ${opt.name} optimizer specifies an unknown optimizer in runAfter: ${dependsOn}`
         );
       }
```
ngfreiter, line 51: I don't think this handling of `runAfter` totally works for `RegEx`. If the `runAfter` is a `RegEx`, then there may only be one expression in the array of `runAfter`. I think this code will only find one `preceedingOptimizer` that matches the `runAfter`, not all possible ones.

ngfreiter (changes requested):
I don't think the RegEx implementation is working quite right, so some of the errors referenced in this comment: https://github.com/FHIR/GoFSH/pull/106#discussion_r597816936 I am still seeing. Running this on US Core with FSHing Trip is really your best bet at catching any bugs, and I think you should give that a shot to double check before we re-review.

src/optimizer/loadOptimizers.ts
julianxcarter: You're right! Commit `8b519d4` switches this `find` to a `filter`, so it should organize the optimizers correctly now.

src/optimizer/loadOptimizers.ts
```diff
@@ -60,6 +74,7 @@ export async function loadOptimizers(
   let ordered: string[];
   try {
     ordered = toposort.array(nodes, edges).reverse();
+    console.log(JSON.stringify(ordered));
```
ngfreiter, line 77: We want to take this out (good news is though that it does confirm for me that "simplify_array_indexing" is running last).

ngfreiter (changes requested):
I'm still seeing some minor bugs when generating US Core. For example, here is a comparison of how the FSH looks on master, vs how it looks on this branch:
image

This results from this generated block of soft indexing:
image

I've done some digging, and the way this works in SUSHI is that the `=` indices will refer to index `0`, when we really want those rules to apply to `1`. GoFSH seems to think that SUSHI will resolve the `=` to index `1`. I wasn't sure what the intended behavior is. Should SUSHI be resolving those `=` to `1`? Or should GoFSH be generating something different?

cmoesel:
> I've done some digging, and the way this works in SUSHI is that the = indices will refer to index 0, when we really want those rules to apply to 1. GoFSH seems to think that SUSHI will resolve the = to index 1. I wasn't sure what the intended behavior is. Should SUSHI be resolving those = to 1? Or should GoFSH be generating something different?

I think that `[=]` should refer to the most recently used index for that path.  So:
```
* foo[0] // foo[0]
* foo[1] // foo[1]
* foo[=] // foo[1]
* foo[+] // foo[2]
* foo[=] // foo[2]
```

That's what makes the most sense to me.  Would you agree?

julianxcarter:
> I've done some digging, and the way this works in SUSHI is that the `=` indices will refer to index `0`, when we really want those rules to apply to `1`. GoFSH seems to think that SUSHI will resolve the `=` to index `1`. I wasn't sure what the intended behavior is. Should SUSHI be resolving those `=` to `1`? Or should GoFSH be generating something different?

The FSH that GoFSH is producing here is correct, Sushi should be resolving the `=` to index 1 rather than to index 0 as it's doing now. [This bug is occurring due to an assumption that I made on this line within sushi](https://github.com/FHIR/sushi/blob/e93ae9a0506a2dc707ac0a377ae7b3a2e934a25f/src/utils/PathUtils.ts#L77) that the first element assigned within an array would always be of index 0, which is not correct. I'll update that line within sushi so that cases like this will be properly accounted for.

src/optimizer/loadOptimizers.ts
julianxcarter: Removed!

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
```diff
@@ -0,0 +1,213 @@
...
+    } else {
+      pathMap.set(mapName, parseInt(currentNumericBracket));
+    }
+  }
+}
+
+function splitOnPathPeriods(path: string): string[] {
+  return path.split(/\.(?![^\[]*\])/g); // match a period that isn't within square brackets
+}
+
```
ngfreiter, line 120: Could you just add a comment here indicating that this function should be removed once we update GoFSH to use a version > 1.3.1 of SUSHI?

ngfreiter (commented):
I have one more teeny tiny comment, but with the updates you made to SUSHI, this looks good to me. It had no diffs when I ran this on mCODE and US Core. This was a tricky task, awesome work on figuring it out.

src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
julianxcarter: Sounds reasonable!

ngfreiter (approved):
Looks good! 🎆

mint-thompson (approved):

PR #108, opened 2021-04-12
Process Configuration and Infer missing IG properties

jafeltra:
This PR updates how we extract configuration information from IG resources. The changes came up after conversation on a [FSH Online PR](https://github.com/FSHSchool/FSHOnline/pull/46).

If an IG resource is missing any properties, including the required properties of `fhirVersion` and `url`, the ConfigurationProcessor now tries to infer them from the other resources, aligning with how properties are inferred from resources when there is not IG resource present. If the required properties `fhirVersion` and `url` are missing, a warning is still logged before inferring from the resources.

This PR accomplishes this inference by merging the old `ConfigurationProcessor` with the `ConfigurationExtractor`. This actually aligns more closely with our current definitions of extractor and processor. It also allows me to use the other existing inference functions and pass along all other resources to use when inferring.

This also updates the default `canonical` value that GoFSH creates to be `http://example.org`, as discussed [here](https://github.com/FSHSchool/FSHOnline/pull/46#discussion_r611595150).


src/processor/FHIRProcessor.ts
```diff
@@ -30,21 +29,16 @@ export class FHIRProcessor {
...
-          ...this.lake.getAllValueSets()
-        ].map(wild => wild.content)
-      );
-    }
+    const resources = [
+      ...this.lake.getAllStructureDefinitions(),
+      ...this.lake.getAllCodeSystems(),
+      ...this.lake.getAllValueSets()
+    ].map(wild => wild.content);
+    if (igForConfig) resources.push(igForConfig.content);
```
ngfreiter, line 40: I think stylistically we usually avoid one line if statements? Not sure I'm pretty sure @cmoesel, gave me a stern talking to one time for using one. So consider removing this or risk his _wrath_.

src/extractor/ConfigurationExtractor.ts
```diff
@@ -1,34 +1,63 @@
...
+      igResource?.status ??
+      (ConfigurationExtractor.inferString(resources, 'status') as ImplementationGuideStatus);
     if (status) {
       config.config.status = status;
     }
-    const version = ConfigurationExtractor.inferString(resources, 'version');
+    const version = igResource?.version ?? ConfigurationExtractor.inferString(resources, 'version');
     if (version) {
       config.config.version = version;
     }
```
ngfreiter, line 56: This also would apply to `status` and `dependencies`, but is there a reason why we can't do:
```javascript
config.config.version = igResource?.version ?? ConfigurationExtractor.inferString(resources, 'version');
```
and let the value just be not written if it is blank?

src/extractor/ConfigurationExtractor.ts
```diff
@@ -1,34 +1,63 @@
...
+    if (igResource && !igResource.url) {
+      missingIGProperties.push('url');
+    }
+    if (igResource && !igResource.fhirVersion?.length) {
+      missingIGProperties.push('fhirVersion');
+    }
+    if (missingIGProperties.length > 0) {
+      logger.warn(
+        `ImplementationGuide missing properties needed to generate configuration file: ${missingIGProperties.join(
+          ', '
```
ngfreiter, line 21: I think it may be useful to add to this message to indicate that the properties are going to be inferred. Just because it may be confusing the see that the `url` is missing, but then see a `canonical` still show up in your generated config.

ngfreiter (commented):
A few minor comments.

src/extractor/ConfigurationExtractor.ts
jafeltra: I was mostly just following the existing pattern, but I don't see any reason why we couldn't it this way, and the code looks a little cleaner this way, so I'll update!

src/processor/FHIRProcessor.ts
jafeltra: 😳 Do not want to risk his _wrath_ 😳 

(Also pretty sure I've made the same comment before, so I'm risking my own _wrath_)

src/extractor/ConfigurationExtractor.ts
jafeltra: I have run into a snag. By doing this, `config.config.version` and `config.config.status` become `''`, which don't get written to `sushi-config.yaml`, which is good. However, there are various `FhirToFsh` tests that compare the value of a generated config object. This means that I would now need to add in those blank values to the expected object, but the snag is that `status` has a very nicely defined type that includes `'draft' | 'active' | 'retired' | 'unknown'` but not `''`, which means TypeScript is not happy if I try to set the default configuration object in that test to have a `status: ''`. I'm leaning towards just leaving these lines how they are so we just don't include the property if it is not defined, but I wanted to check if you or anyone else had clever ideas for getting the test to pass. I could compare each individual property, but I think that actually weakens the tests a bit since a change like this wouldn't cause that type of test to break. I haven't thought of another way to fix it though.

src/extractor/ConfigurationExtractor.ts
jafeltra: Added in 20d75fa!

src/processor/FHIRProcessor.ts
jafeltra: Fixed in 20d75fa!

src/extractor/ConfigurationExtractor.ts
ngfreiter: I'm fine with just leaving it how it is. I don't think it's worth the trouble of those changes for such a minor thing.

ngfreiter (approved):
Looks good to me!

julianxcarter (approved):
This works well for me!

PR #110, opened 2021-04-21
Update README

ngfreiter:
This updates the GoFSH README.md file to be more accurate. The README had indicated that GoFSH was still a work in progress and was incomplete/didn't support generating all desired FSH. Since we have done the 1.0 release, I didn't think any of that was still true, so this PR removes all that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,24 +1,10 @@
-# GoFSH (WORK IN PROGRESS)
+# GoFSH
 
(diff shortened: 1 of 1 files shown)


jafeltra (approved):
It's o-fish-ally o-fish-al!

mint-thompson (approved):

PR #112, opened 2021-05-04
Support fhirVersions > 4.0.1

### LINKED ISSUE #107: GoFSH assumes 4.0.1
We've had a report that GoFSH does not work with FHIR 4.5.0 IGs, and honestly, that's not surprising at all.  We should add support for the same FHIR versions as SUSHI supports (e.g., >= 4.0.1).

ngfreiter:
This fixes #107 by adding support for versions of FHIR > 4.0.1. 

# Changes
There are a few things that had to change to do this. First there were changes related to allowing loading versions of FHIR > 4.0.1:
* `ConfigurationExtractor.ts`was updated so that only supported versions of FHIR will be used as the `fhirVersion` in the configuration.
* `app.ts` was updated so that the package corresponding to the given `fhirVersion` is included in the set of dependencies that is loaded.
* `Processing.ts` was updated so that the `loadExternalDependencies` function now only defaults to FHIR 4.0.1 if no other version of FHIR is already specified in the dependencies. I don't think this check should be necessary in any situation anymore, but I think it doesn't hurt to keep it.

Then there were some changes to make `CodeableReference` work:
* `OnlyRuleExtractor.ts` was updated to check if the type of an element is `Reference` or `CodeableReference`.

Then changes to make `integer64` work:
* `element.ts` was updated so that `getFSHValue` will convert an `integer64` string value to a `bigint`
* `ExportableAssignmentRule.ts` was updated so that its `toFSH` method will print a `bigint` without `""` around it
* `AssignmentRuleExtractor.ts` was updated so that a string value representing an `integer64` will be extracted to a `bigint`, so that when it is printed with `toFSH`, it will not have `""` around it.

We actually could leave out the `integer64` stuff if we are fine with writing an `integer64` in FSH as a string, which SUSHI does support. I'm not really sure what the "preferred" style is in this case. An `integer64` is serialized in JSON as a string, so maybe that is preferred, but all other numbers are written as numbers, so maybe that is preferred.

# Testing
Here is a FSH Online [link](https://fshschool.org/FSHOnline/#/share/3egoUc0) to the FSH I used to test this, by running through SUSHI and then GoFSH and comparing with the original. The things to test are:
* If `fhirVersion` is something other than 4.0.1, we load that version of FHIR
* If `fhirVersion` is not present and cannot be inferred, we still default to 4.0.1
* When a `fhirVersion` other than 4.0.1 is given, you can run GoFSH on SDs that inherit from Resources only in that version of FHIR
* When a VS binding or a Reference type constraint is placed on a `CodeableReference`, GoFSH processes that without resorting to `^` rules
* When an `integer64` is assigned on an `Instance` or a `Profile/Extension`, GoFSH writes it as a number, not a string (but the number loses no precision)

This feels a little too easy, but I cannot find evidence of anything else we've done to SUSHI relating to FHIR R5, so I'm not sure what else there would be to do here. But please as part of the review make sure to also consider anything that may be missing from this PR.

### DIFF
--- a/src/api/FhirToFsh.ts
+++ b/src/api/FhirToFsh.ts
@@ -77,10 +77,12 @@ export async function fhirToFsh(
   const dependencies = configuration.config.dependencies?.map(
     (dep: fhirtypes.ImplementationGuideDependsOn) => `${dep.packageId}@${dep.version}`
   );
-  const fhirPackageId = configuration.config.fhirVersion[0].startsWith('4.0')
-    ? 'hl7.fhir.r4.core'
-    : 'hl7.fhir.r5.core';
-  dependencies.push(`${fhirPackageId}@${configuration.config.fhirVersion[0]}`);
+  if (dependencies) {
+    const fhirPackageId = configuration.config.fhirVersion[0].startsWith('4.0')
+      ? 'hl7.fhir.r4.core'
+      : 'hl7.fhir.r5.core';
+    dependencies.push(`${fhirPackageId}@${configuration.config.fhirVersion[0]}`);
+  }
   await Promise.all(loadExternalDependencies(defs, dependencies));
 
   // Process the FHIR to rules, and then export to FSH
--- a/src/api/FhirToFsh.ts
+++ b/src/api/FhirToFsh.ts
@@ -77,6 +77,10 @@ export async function fhirToFsh(
   const dependencies = configuration.config.dependencies?.map(
     (dep: fhirtypes.ImplementationGuideDependsOn) => `${dep.packageId}@${dep.version}`
   );
+  const fhirPackageId = configuration.config.fhirVersion[0].startsWith('4.0')
+    ? 'hl7.fhir.r4.core'
+    : 'hl7.fhir.r5.core';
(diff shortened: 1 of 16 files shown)


mint-thompson (commented):
Overall, this looks pretty great. Thank you! My one question is for cases where the user has specified FHIR R4 in their configuration, but are using R5 types in their resources. If this kind of scenario is possible, I think it would be helpful to warn the user that they have likely made a mistake in their configuration.

ngfreiter:
> Overall, this looks pretty great. Thank you! My one question is for cases where the user has specified FHIR R4 in their configuration, but are using R5 types in their resources. If this kind of scenario is possible, I think it would be helpful to warn the user that they have likely made a mistake in their configuration.

I _think_ this should generally be covered. Like if you make a sample IG with a resource that is only in R5, say `MedicinalProductDefinition`, and then point the IG `fhirVersion` at 4.0.1, you'll get a warning about not being able to find the definition:
image

jafeltra (commented):
This looks really good! I just had one really minor question. I also looked through the SUSHI PRs we had for R5 features and didn't notice anything missing.

jafeltra (approved):
Thanks! Looks great to me!

mint-thompson (approved):

Issue #122, opened 2021-06-03
Failure translating extension definitions to FSH


markkramerus:
This is a simple extension that should be readily convertible to FSH. GoFSH is coming up with the following FSH that is rejected by SUSHI:

```
* valueReference contains valueReference
* valueReference only Reference(HealthcareService)
```
The two files are attached (remove the .txt extensions).

[AssociatedHealthcareService-Extension.fsh.txt](https://github.com/FHIR/GoFSH/files/6592106/AssociatedHealthcareService-Extension.fsh.txt)

[structuredefinition-associated-healthcareservice.xml.txt](https://github.com/FHIR/GoFSH/files/6592096/structuredefinition-associated-healthcareservice.xml.txt)

markkramerus:
Note that this is just one of many errors translating extensions in [hl7au/au-fhir-base](https://github.com/hl7au/au-fhir-base)

markkramerus:
Here's another interesting one, translating AustralianIndigenousStatus extension:
```
* valueCoding contains valueCoding 1..
* valueCoding[valueCoding] only Coding
* valueCoding[valueCoding] from $australian-indigenous-status-1_1 (required)
* valueCoding[valueCoding] ^short = "Indigenous Status Code"
* valueCoding[valueCoding] ^definition = "NHDD based indigenous status code."
* valueCoding[valueCoding] ^binding.description = "Indigenous Status"
```

cmoesel:
Nick and I have produced a fix for this in the PR linked above -- but it does not fix the last case you note (for AustralianIndigenousStatus).  The reason is because that case uses an `id` that I'm not even sure is valid.  It has this element:
```xml
<element id="Extension.valueCoding:valueCoding">
  <path value="Extension.valueCoding"/>
  <sliceName value="valueCoding"/>
  <short value="Indigenous Status Code"/>
  <definition value="NHDD based indigenous status code."/>
  <min value="1"/>
  <type>
    <code value="Coding"/>
  </type>
  <binding>
    <strength value="required"/>
    <description value="Indigenous Status"/>
    <valueSet value="https://healthterminologies.gov.au/fhir/ValueSet/australian-indigenous-status-1"/>
  </binding>
</element>
```

[...]

PR #123, opened 2021-06-03
Fix bug with choice slices

ngfreiter:
Fixes https://github.com/FHIR/GoFSH/issues/122.

When someone has a `sliceName` specified in the differential on the choice slice, SUSHI doesn't really know how to handle that situation, see this example (pardon my xml):
```xml
<element id="Extension.value[x]:valueReference">
      <path value="Extension.valueReference"/>
      <sliceName value="valueReference"/>
      <min value="1"/>
      <type>
        <code value="Reference"/>
        <targetProfile value="http://hl7.org/fhir/StructureDefinition/HealthcareService"/>
      </type>
    </element>
```
I wasn't sure there was a good way to make this work in SUSHI using `contains` rules, so in this case I felt like it results in more clear FSH to just not process this as a `contains` rule. But I may be missing something about how this should be written as FSH, so let me know.

### DIFF
--- a/src/processor/StructureDefinitionProcessor.ts
+++ b/src/processor/StructureDefinitionProcessor.ts
@@ -90,8 +90,8 @@ export class StructureDefinitionProcessor {
     // Then extract rules based on the differential elements
     elements.forEach(element => {
       const ancestorSliceDefinition = getAncestorSliceDefinition(element, input, fisher);
-      // if there is a slice, but no ancestor definition, capture with a contains rule
-      if (element.sliceName && ancestorSliceDefinition == null) {
+      // if there is a slice, which is not a choice slice, but no ancestor definition, capture with a contains rule
+      if (element.sliceName && !element.id.includes('[x]:') && ancestorSliceDefinition == null) {
         newRules.push(
           ContainsRuleExtractor.process(element, input, fisher),
           OnlyRuleExtractor.process(element),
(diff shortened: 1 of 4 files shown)


cmoesel (changes requested):
This fix is a bit too agressive now.  If you run it on the [au-base xml definitions](http://build.fhir.org/ig/hl7au/au-fhir-base/definitions.xml.zip) and compare to what is produced by gofsh master, you'll see that some `contains` statements are dropped when they shouldn't be.

For example:
image

and

image

I think this is because the new code looks for `[x]:` _anywhere_ in the path when it really should only be looking at it in the tail of the path.

cmoesel:
I've pushed a fix for the issue I identified above.  Note that this doesn't address _all_ extension-related issues in the au-base IG.  There are some extensions where they have ids that combine the shortcut syntax _and_ the slicename (e.g., `Extension.valueIdentifier:valueIdentifier`.  This PR does not fix that case because the `id` does not have the `[x]` in it.

I'm not sure that is even a valid `id` anyway, so I do not plan to address it in this PR.

cmoesel (approved):
I'm approving this now that it contains my additional fix.  While there are still some issues in the au-base IG, they're related to ids that I'm not sure are valid.  We should find out if they are before spending too much time trying to fix them.  So I propose we accept this PR as it is now.

We still need another reviewer on this.

jafeltra (approved):
This looks good to me. I tested it out using the XML definitions and comparing the differences to what the current version of gofsh generated, and the differences all made sense to me!

PR #124, opened 2021-06-04
References to contained resources should refer by id (w/ no #)

### LINKED ISSUE #120: Invalid Reference(#name)
This example: [http://hl7.org/fhir/us/core/2021Jan/MedicationRequest-uscore-mo2.json](http://hl7.org/fhir/us/core/2021Jan/MedicationRequest-uscore-mo2.json)
has a contained instance of a medication `med2`.  This line in the generated FSH looks wrong:
```
* medicationReference.reference = "#med2"
```
because it has an extra # sign. Clearly this is not a code. Is this a bug?

cmoesel:
Prior to this fix, when GoFSH created references to contained instances, it generated references like this:
```
* medicationReference = Reference(#mymed)
```
SUSHI did not like this because the parser tried to interpret `#mymed` as a code.  In FSH, you reference contained resources simply by their id (without the `#` prefix).  This PR fixes it so that the above reference is now rendered as:
```
* medicationReference = Reference(mymed)
```

The unit test shows the fix but you can also manually test w/ a simple JSON instance like:
```json
{
  "resourceType": "MedicationRequest",
  "id": "MyMedRequest",
  "status": "completed",
  "intent": "order",
  "contained": [
    {
      "resourceType": "Medication",
      "id": "mymed",
      "code": {
        "coding": [
          {
            "code": "bar",
            "system": "http://foo"
          }
        ]
      }
    }
  ],
  "subject": {
    "reference": "Patient/Bob"
  },
  "medicationReference": {
    "reference": "#mymed"
  }
}
```

If you want to test it on the package that exposed this bug, you can run GoFSH on the [US Core 3.2.0 package](http://hl7.org/fhir/us/core/2021Jan/package.tgz) to see it in action.  After running GoFSH, run SUSHI.  If you used the master version of GoFSH, _one of the errors_ SUSHI outputs will be:
> error Resolved value "Reference(" is not a valid URI.
>  File: /Users/cmoesel/dev/fhir/GoFSH/gofsh/input/fsh/instances.fsh
>  Line: 6838

If you use this branch's version of GoFSH and run SUSHI on the results, that error should go away.

Fixes #120

### DIFF
--- a/src/optimizer/plugins/AddReferenceKeywordOptimizer.ts
+++ b/src/optimizer/plugins/AddReferenceKeywordOptimizer.ts
@@ -27,7 +27,9 @@ export default {
             const parentPath = splitOnPathPeriods(rule.path).slice(0, -1).join('.');
             const searchablePath = parentPath.replace(/\[\d+\]/g, '');
             if (sd?.findElementByPath(searchablePath, fisher)?.type?.[0]?.code === 'Reference') {
-              const reference = new fshtypes.FshReference(rule.value);
+              // Remove '#' prefix from references to contained resources (if applicable)
+              const referenceTarget = rule.value.replace(/^#/, '');
+              const reference = new fshtypes.FshReference(referenceTarget);
               // If we are converting the reference, look for a matching display rule, using the
               // original numerical parent path
               const matchingDisplayRuleIndex = instance.rules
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
Makes sense to me! Tested with the provided example and US Core and worked as described.

mint-thompson (approved):

PR #127, opened 2021-06-07
Export LogicalModels and Resources as Instances

### LINKED ISSUE #125: GoFSH translates logical models and resources as FSH Profiles
I know that we're working on GoFSH support for the FSH STU2 Logical and Resource authoring -- but assuming that still has some work to do, I think we should release a quick fix on this ASAP.

The problem is that if you pass a logical model or resource to GoFSH, it generates a FSH profile with invalid paths.  For example, the following logical model:
```json
{
  "resourceType": "StructureDefinition",
  "id": "LogicalExample",
  "url": "http://example.org/StructureDefinition/LogicalExample",
  "version": "1.0.0",
  "name": "LogicalExample",
  "status": "active",
  "fhirVersion": "4.0.1",
  "mapping": [
    {
      "identity": "rim",
      "name": "RIM Mapping",
      "uri": "http://hl7.org/v3"
    }
  ],
  "kind": "logical",
  "abstract": false,
  "type": "http://example.org/StructureDefinition/LogicalExample",
  "baseDefinition": "http://hl7.org/fhir/StructureDefinition/Base",
  "derivation": "specialization",
  "differential": {
    "element": [
      {
        "id": "LogicalExample",
        "path": "LogicalExample"
      },
      {
        "id": "LogicalExample.foo",
        "path": "LogicalExample.foo",
        "short": "A foo",
        "definition": "A foo",
        "min": 0,
        "max": "1",
        "type": [
          {
            "code": "string"
          }
        ]
      }
    ]
  }
}
```
gets translated as:
```
Profile: LogicalExample
Parent: Base
Id: LogicalExample
* ^version = "1.0.0"
* ^kind = #logical
* ^abstract = false
* ^type = "http://example.org/StructureDefinition/LogicalExample"
* foo 0..1
* foo only string
* foo ^short = "A foo"
* foo ^definition = "A foo"
```
This is obviously wrong.

The correct behavior is that if GoFSH encounters a `StructureDefinition` whose `derivation` is `specialization`, then it should output it as an `Instance` of a `StructureDefinition`.  I'm assuming this is a fairly easy fix to hold us over until we have proper support for `Logical` and `Resource`.

jafeltra:
This PR supports exporting `LogicalModel`s and `Resource`s as `Instance`s. This is a temporary fix that should be removed when full support for `LogicalModel` and `Resource` is implemented in GoFSH. I tried to mark the places that will need to be updated again when that support is included.

Fixes #125

### DIFF
--- a/src/processor/LakeOfFHIR.ts
+++ b/src/processor/LakeOfFHIR.ts
@@ -16,7 +16,9 @@ export class LakeOfFHIR implements utils.Fishable {
    * @returns {WildFHIR[]}
    */
   getAllStructureDefinitions(): WildFHIR[] {
-    return this.docs.filter(d => d.content.resourceType === 'StructureDefinition');
+    return this.docs
+      .filter(d => d.content.resourceType === 'StructureDefinition')
+      .filter(d => !this.isSpecialization(d)); // TODO: Remove this filter when full support for LogicalModels and Resources is implemented
   }
 
   /**
@@ -62,8 +64,11 @@ export class LakeOfFHIR implements utils.Fishable {
     return this.docs.filter(d => {
       switch (d.content.resourceType) {
         case 'ImplementationGuide':
-        case 'StructureDefinition':
           return false;
+        case 'StructureDefinition':
+          // TODO: This is a temporary fix to support exporting LogicalModels and Resources as Instances.
+          // This should return false when full support is implemented.
+          return this.isSpecialization(d);
         case 'CodeSystem':
           return (
             includeUnsupportedTerminologyResources &&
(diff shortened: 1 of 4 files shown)


src/processor/LakeOfFHIR.ts
```diff
@@ -16,7 +16,9 @@ export class LakeOfFHIR implements utils.Fishable {
    * @returns {WildFHIR[]}
    */
   getAllStructureDefinitions(): WildFHIR[] {
-    return this.docs.filter(d => d.content.resourceType === 'StructureDefinition');
+    return this.docs
+      .filter(d => d.content.resourceType === 'StructureDefinition')
+      .filter(d => !this.isSpecialization(d)); // TODO: Remove this filter when full support for LogicalModels and Resources is implemented
```
mint-thompson, line 21: Somehow, I didn't realize until you pointed it out here: Profiles and Extensions are constraints, while Logicals and Resources are specializations. A delightfully concise implementation.

mint-thompson (approved):

src/processor/LakeOfFHIR.ts
cmoesel: Data types are also `specialization`s (just FYI).

cmoesel: Except a few variants on `Quantity` that are actually profiles...

julianxcarter (approved):
This looks good and works well!

PR #128, opened 2021-06-07
Update SUSHI dependency to 2.0.0 beta

mint-thompson:
The new major version of SUSHI contains a lot of new features relevant to goFSH. FSH types and rules now have toFSH methods, which makes some code in goFSH redundant. ExportableValueSetComponentRule needed to be changed immediately due to the use of a private function, and other classes will be changed in later commits.

Modify the mock of loadDependency used in Processing tests. The previous mock was failing due to the fhirdefs object not being able to set a value on its loadDependency property.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2907,21 +2907,20 @@
       "optional": true
     },
     "fsh-sushi": {
-      "version": "1.3.2",
-      "resolved": "https://registry.npmjs.org/fsh-sushi/-/fsh-sushi-1.3.2.tgz",
-      "integrity": "sha512-KkLxG9vyjuxh0GZpQYD80ETMVqWnOnKUjCGtxfC2F3bAiyViRUKWiSZTEuPlcUiYMolDpabnGqrsc/n6nHHtpw==",
+      "version": "2.0.0-beta.1",
+      "resolved": "https://registry.npmjs.org/fsh-sushi/-/fsh-sushi-2.0.0-beta.1.tgz",
+      "integrity": "sha512-O9PJt+U/CYzWC83RgGNlEHWf76t7uAc+wU9pb+24nxhP4Qt/8pkzoEPkNd/UchmNf9fk7nRgc90USD/8zQqVKw==",
       "requires": {
         "antlr4": "~4.8.0",
         "axios": "^0.21.1",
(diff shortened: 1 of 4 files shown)


jafeltra:
I'm not exactly sure what happened in the CI, but it looks like some of the check aren't passing. I can help look into that if you'd like! Besides that, I noticed two other small things that I'm curious what you think about:

1. When I ran `npm install`, my `package-lock.json` changes compared to what is committed here. I did have `package-lock.json` change unexpectedly a couple times on FSH Online, so I'm not sure if it's me, but if it isn't, we might want to include those differences also.
2. It looks like Processing.test.ts now logs messages from the logger during the tests. I thought normally importing the loggerSpy suppressed these messages, but it looks like we already import that in those tests, so I'm not sure why that's happening.

mint-thompson:
> I'm not exactly sure what happened in the CI, but it looks like some of the checks aren't passing. I can help look into that if you'd like! Besides that, I noticed two other small things that I'm curious what you think about:
> 
> 1. When I ran `npm install`, my `package-lock.json` changes compared to what is committed here. I did have `package-lock.json` change unexpectedly a couple times on FSH Online, so I'm not sure if it's me, but if it isn't, we might want to include those differences also.
> 2. It looks like Processing.test.ts now logs messages from the logger during the tests. I thought normally importing the loggerSpy suppressed these messages, but it looks like we already import that in those tests, so I'm not sure why that's happening.

The short answer for the cause of failure is: the mock I added isn't actually getting used, so I'm re-wrangling the mock so that it'll actually get used.

jafeltra (approved):
So I have no idea what is going on with my npm (or my understanding of npm) today, but I don't see any changes in `package-lock.json` and the updates to the mock seem to have resolved the stray logs in the `Processing.test.ts` tests, so this looks good to me!

cmoesel (approved):
This works great, @mint-thompson.  I tested it against US Core 3.2.0 and it did very well.

@jafeltra -- I did not experience either one of the two issues you reported.  My package-lock.json is the same as is checked in and I saw no logs during the unit test run.

cmoesel:
Nevermind.  I see those issues are resolved for @jafeltra too.  Great!

PR #130, opened 2021-06-08
Upgrade SUSHI dependency to 2.0.0, refactor toFSH implementations

mint-thompson:
Completes task [CIMPL-747](https://standardhealthrecord.atlassian.net/browse/CIMPL-747).

The first part of this task was already completed by #128, which sets the new dependency. What this PR does is update the `toFSH` implementations in the various Exportable classes to take advantage of the fact that the classes provided by SUSHI now have their own `toFSH` methods. In many cases, these implementations can be used as-is. In other cases, the SUSHI implementation is very close to what goFSH wants. Note that `Logical`, `Resource`, and `AddElementRule` are not in this PR, as support for those classes is coming as part of [another task](https://standardhealthrecord.atlassian.net/browse/CIMPL-729).

These classes have their `toFSH` methods removed and use the inherited implementation:
```
ExportableAssignmentRule
ExportableBindingRule
ExportableCardRule
ExportableCodeSystem
ExportableConceptRule
ExportableFlagRule
ExportableInstance
ExportableInvariant
ExportableMappingRule
ExportableOnlyRule
ExportableProfile
ExportableValueSet
ExportableValueSetConceptComponentRule
ExportableValueSetFilterComponentRule
```

These classes have their `toFSH` methods refactored to make use of the inherited implementation:
```
ExportableCaretValueRule
ExportableExtension
ExportableMapping
```

Other exportable classes maintain their current implementations:
* `ExportableAlias` doesn't extend a SUSHI-provided class, so it needs its own implementation.
* `ExportableCombinedCardFlagRule` could make use of `toFSH`, but just uses some other helper methods on component rule classes instead. I feel like it's about equally complex either way.
* `ExportableContainsRule` contains logic to find and use the cardinality of and flags for each contained item. Although the logic is similar to the SUSHI implementation, I feel that writing an implementation that finds and replaces the SUSHI-provided `0..` cardinality and inserts the appropriate flags is getting a bit deeper than I'd like with string manipulations.
* `ExportableObeysRule` operates on multiple invariants, which is fundamentally different from the base class.

Some helper functions that were used by removed `toFSH` implementations are also removed, as they are not used anywhere else.


ngfreiter (commented):
I'm seeing some diffs where it _seems_ like we are now adding too many `\` when we are trying to escape things. Here is an example on a `CaretValueRule`: 
image

and another example that was on an `AssignmentRule`: 
image

ngfreiter (commented):
I think it is because when we extract values on `Assignment` and `Caret` rules, we `fshify`, but now we also `fshify` in the `toFSH` method, so we've got twice the `fshification` on our hands.

cmoesel:
> we've got twice the `fshification` on our hands

I don't know why.  But I really like this phrase.  In some ways it feels whimsical.  In other ways, it feels like the major plot point of a very suspenseful movie.

src/utils/element.ts
```diff
@@ -92,7 +92,8 @@ export function getFSHValue(
   } else if (element?.type?.[0]?.code === 'integer64') {
     return BigInt(value);
   }
-  return typeof value === 'string' ? fshifyString(value) : value;
+  // return typeof value === 'string' ? fshifyString(value) : value;
```
ngfreiter, line 95: Should this have been deleted?

ngfreiter (commented):
Looks good, just one maybe typo?

ngfreiter (approved):
Good stuff!

julianxcarter (approved):
Looks good to me!

PR #131, opened 2021-06-10
Log an error when a sliceName is illegal

### LINKED ISSUE #114: Generating illegal FSH
Running GoFSH on [us-odh](https://github.com/afourquet/us-odh/tree/master).  GoFSH is generating the following rule in translating odh-PastOrPresentJob.json:

```
* extension[odh-isCurrentJob-extension] contains odhIsCurrentJob
* extension[odh-isCurrentJob-extension] only ODHIsCurrentJobExtension
```
The "contains" rule is broken because (1) it refers to a slice that does not exist, and (2) it does not have the "named foo" clause.

ngfreiter:
Fixes #114 by preventing generating of illegal FSH in the case that the `sliceName` in an element, `sliceName` implied by the `id` of an element do not match. The `id` is supposed to be set according to [this](https://www.hl7.org/fhir/elementdefinition.html#id) portion of the spec, but if it is not, GoFSH was generating illegal FSH, as is the case in the linked bug.

Now, we prefer the value implied by the `id` to the `sliceName`, so the generated FSH is legal (but will differ from the input JSON). An error is logged to warn the user why they are going to see this difference in the JSON.


jafeltra (approved):
Makes sense to me and resolves the issue reported.

mint-thompson (dismissed):
This looks good! The only thing I am thinking of is: do we want to provide a link to the spec when showing this error?

ngfreiter:
> This looks good! The only thing I am thinking of is: do we want to provide a link to the spec when showing this error?

Yeah good call, done in https://github.com/FHIR/GoFSH/pull/131/commits/0a9aa25fb180c42b17cc237723469e6d70fbecd8.

ngfreiter:
@jafeltra and @mint-thompson, the update in https://github.com/FHIR/GoFSH/commit/0a9aa25fb180c42b17cc237723469e6d70fbecd8 dismissed your approvals, so this needs a re-review, but it was a very minor change.

jafeltra (approved):

mint-thompson (approved):

PR #132, opened 2021-06-14
Generate Logicals and Resources

mint-thompson:
Completes task [CIMPL-729](https://standardhealthrecord.atlassian.net/browse/CIMPL-729).

The StructureDefinitionProcessor now figures out when a StructureDefinition represents a Logical or a Resource. In order to correctly do this, the logic for deciding what a StructureDefinition represents is more precise. The LakeOfFHIR also implements this logic in order to provide entities of the correct type. To go along with these code updates, many of the test fixtures needed to be updated in order to provide all the necessary values to be categorized correctly.

Logicals and Resources have all the usual StructureDefinition rules, and they also have AddElementRules. So, when extracting rules for a Logical or Resource, AddElementRuleExtractor is used. Cardinality, flag, and type information is contained within these rules, so the extractors for those rules are not used in the main rule extractor. Slices on new elements and assigned values on new elements are not allowed, but rules for them are still made. When these rules are made, the user is warned that they're going to have problems very soon.

FSHExporter is also updated to handle the two new entity types. They don't need any particular special handling here, since they have a `toFSH` method like all the other entities. Since both Logical and Resource have a default parent, they get to leave it out the same way that Extension does.


cmoesel (changes requested):
Good work!  This PR ended up being bigger than I expected (but it makes sense)!

I ran this against the [HIV Case Reporting IG's definitions](https://build.fhir.org/ig/openhie/hiv-ig/downloads.html) and it came out clean, but the `Logical`s included generated narratives in the FSH.  We remove those for profiles and extensions; I think we need to remove them for `Logical`s and `Resource`s too.

We also need to update the results summary at the end of the run to include logicals and resources in the counts:
image

mint-thompson:
@cmoesel Thank you for the review. The only thing that's giving me a hard time right now is figuring out how to restructure the result printout: we now have 9 FSH types to report, plus the mandatory sea creature count, which means a total of 10 output items. I feel like a 3/3/4 split is probably the best option here, or 3/3/3 if we remove the mandatory sea creature count. I'll try a few things out and ask the team what looks good.

mint-thompson:
Here are two potential output styles I've come up with. I didn't want to try anything that went with 5 fields in a row, since I think the output would be too wide at that point.

3/3/3 with no sea creatures
```
╔═════════════════════════ GoFSH RESULTS ═════════════════════════╗
║ ╭────────────────────┬───────────────────┬────────────────────╮ ║
║ │      Profiles      │    Extensions     │      Logicals      │ ║
║ ├────────────────────┼───────────────────┼────────────────────┤ ║
║ │         2          │         0         │         0          │ ║
║ ╰────────────────────┴───────────────────┴────────────────────╯ ║
║ ╭────────────────────┬───────────────────┬────────────────────╮ ║
║ │     Resources      │     ValueSets     │     CodeSystems    │ ║
║ ├────────────────────┼───────────────────┼────────────────────┤ ║
║ │         0          │         0         │         1          │ ║
║ ╰────────────────────┴───────────────────┴────────────────────╯ ║
║ ╭────────────────────┬───────────────────┬────────────────────╮ ║
║ │     Instances      │    Invariants     │      Mappings      │ ║
║ ├────────────────────┼───────────────────┼────────────────────┤ ║
║ │         1          │         1         │         0          │ ║
║ ╰────────────────────┴───────────────────┴────────────────────╯ ║
║                                                                 ║
╠═════════════════════════════════════════════════════════════════╣
║ Unfor-tuna-tely, there are errors.      5 Errors     0 Warnings ║
╚═════════════════════════════════════════════════════════════════╝
```

3/3/4 with sea creatures
```
╔═════════════════════════ GoFSH RESULTS ═════════════════════════╗
║ ╭────────────────────┬───────────────────┬────────────────────╮ ║
║ │      Profiles      │    Extensions     │      Logicals      │ ║
║ ├────────────────────┼───────────────────┼────────────────────┤ ║
║ │         2          │         0         │         0          │ ║
║ ╰────────────────────┴───────────────────┴────────────────────╯ ║
║ ╭────────────────────┬───────────────────┬────────────────────╮ ║
║ │     Resources      │     ValueSets     │     CodeSystems    │ ║
║ ├────────────────────┼───────────────────┼────────────────────┤ ║
║ │         0          │         0         │         1          │ ║
║ ╰────────────────────┴───────────────────┴────────────────────╯ ║
║ ╭──────────────┬───────────────┬──────────────┬───────────────╮ ║
║ │  Instances   │  Invariants   │   Mappings   │  Monk Seals   | ║
║ ├──────────────┼───────────────┼──────────────┼───────────────┤ ║
║ │      1       │       1       │      0       │  48 insulted  | ║
║ ╰──────────────┴───────────────┴──────────────┴───────────────╯ ║
║                                                                 ║
╠═════════════════════════════════════════════════════════════════╣
║ You spawned some errors.                5 Errors     0 Warnings ║
╚═════════════════════════════════════════════════════════════════╝
```

Please let me know what y'all think of these, or if there are other changes you think I should try for this.

ngfreiter:
> Please let me know what y'all think of these, or if there are other changes you think I should try for this.

I'm going to vote +1 for the 3x3. It's sad to see the animals go, but I think those tended to confuse people anyway.

jafeltra:
> > Please let me know what y'all think of these, or if there are other changes you think I should try for this.
> 
> I'm going to vote +1 for the 3x3. It's sad to see the animals go, but I think those tended to confuse people anyway.

Yeah I think I'm going to vote for 3x3 too. I'm sad for the sea creatures, but I think they might be too confusing to warrant making an extra box for them now.

cmoesel:
> > > Please let me know what y'all think of these, or if there are other changes you think I should try for this.
> > 
> > 
> > I'm going to vote +1 for the 3x3. It's sad to see the animals go, but I think those tended to confuse people anyway.
> 
> Yeah I think I'm going to vote for 3x3 too. I'm sad for the sea creatures, but I think they might be too confusing to warrant making an extra box for them now.

The sea creatures had a good run - but their purpose here on earth has now been fulfilled.  It is time for them to grace other lands.

jafeltra:
> The sea creatures had a good run - but their purpose here on Earth has now been fulfilled. It is time for them to grace other lands.

Do you mean their _porpoise_ here on Earth has been fulfilled?

src/app.ts
```diff
@@ -16,9 +16,7 @@ import {
   logger,
   stats,
   fshingTrip,
-  getRandomPun,
-  getRandomSeaCreatures,
-  getRandomSeaCreaturesStat
```
cmoesel, line 21: I think this means we can delete these functions from `utils.ts` too, right?

cmoesel (commented):
Thanks for addressing my feedback.  I've left a comment on one thing that I think you missed.  It's harmless, but probably should be done.

src/optimizer/plugins/RemoveGeneratedTextRulesOptimizer.ts
```diff
@@ -18,7 +20,14 @@ export default {
   description: 'Remove rules on "text" which represent generated content.',
 
   optimize(pkg: Package): void {
-    [...pkg.profiles, ...pkg.extensions, ...pkg.valueSets, ...pkg.codeSystems].forEach(resource => {
+    [
```
ngfreiter, line 23: I can't comment on the actual line, but I think we want to apply some other optimizers:
 * `remove_publisher_derived_date_rules` to at least `Logical`s as well. I'm not sure if `Resource`s can even make it through the publisher, so they probably don't matter so much.
 * `resolve_parent_urls` to both `Logical` and `Resource` I think.
 * `simplify_array_indexing` to both `Logical` and `Resource` for caret rules.

src/processor/StructureDefinitionProcessor.ts
```diff
@@ -84,18 +105,53 @@ export class StructureDefinitionProcessor {
...
+      parentDefinition = fisher.fishForFHIR(input.baseDefinition);
+    } else if (target instanceof ExportableResource) {
+      parentDefinition = fisher.fishForFHIR(
+        'http://hl7.org/fhir/StructureDefinition/DomainResource'
+      );
+    } else if (target instanceof ExportableLogical) {
+      parentDefinition = fisher.fishForFHIR('http://hl7.org/fhir/StructureDefinition/Base');
+    }
     // First extract the top-level caret rules from the StructureDefinition
     newRules.push(...CaretValueRuleExtractor.processStructureDefinition(input, fisher, config));
```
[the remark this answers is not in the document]
ngfreiter: Also I think this applies to `kind` as well? Since we set `kind` of a `Logical` automatically to `logical`, do we need to generate a `^` rule that sets `kind`? I think we may just want to ignore that property.

ngfreiter (changes requested):
Looking good, I had a few questions/comments.

src/processor/StructureDefinitionProcessor.ts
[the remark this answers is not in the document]
mint-thompson: You are correct! There's a little bit of this logic in `ExportableExtension`, and similar logic can be added to `ExportableLogical` and `ExportableResource`.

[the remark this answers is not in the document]
mint-thompson: Setting `type` automatically is correct, but there are multiple valid values for `kind` for a Profile, so that property can't be ignored entirely. But, we can still check for defaults and avoid making the rule for certain `kind` values.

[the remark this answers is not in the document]
mint-thompson: Wait, I just realized that `kind` is already handled by existing logic: since the parent provides a `kind`, it already gets set if the value is different from the parent value. So there's no need for extra logic on Profile. Still need to check things on `Logical`, though.

src/export/FSHExporter.ts
```diff
@@ -219,9 +219,13 @@ export class FSHExporter {
   private groupByProfile(): Map<string, Exportable[]> {
     const files: Map<string, Exportable[]> = new Map();
 
-    // Group profiles and examples of those profiles into individual files
-    this.fshPackage.profiles.forEach(profile => {
-      files.set(`${profile.name}.fsh`, [profile]);
+    // Group profiles, logicals, and resources with their respective examples into individual files
+    [
+      ...this.fshPackage.profiles,
+      ...this.fshPackage.logicals,
```
ngfreiter, line 225: This is a bit of a tough call, because a `Logical` cannot have `Instances` made of it, but it can have `Invariants`. So I could go either way on if it should have its own file, or just be in one big `logicals.fsh` file. But if you do want to keep each `Logical` in its own file, I think we will want to remove `Logicals` from the logic that finds out if there are any instances of a `Logical`, since that isn't possible.

ngfreiter (commented):
Just one more minor comment.

src/export/FSHExporter.ts
mint-thompson: My opinion is that since a Logical can have invariants on it, each one should be allowed to have its own file in the `group-by-profile` strategy. I'll update the strategy to skip checking for Instances of a Logical.

mint-thompson: Upon a closer read, no code change is necessary, but I have updated the comments to be more accurate.

ngfreiter (approved):
Looks good to me!

cmoesel (changes requested):
I think there may still just be a few small things to resolve.

1. When I run `npm install` I end up w/ a different `package-lock.json` than what is checked in.
2. When I ran this on a custom resource w/ a user-defined `short` and `definition` on the _root element_, GoFSH did not generate any rules for those.  Oddly enough, it _did_ generate a rule for the root `alias` attribute though.
3. When I ran this on the [hiv-ig](http://build.fhir.org/ig/openhie/hiv-ig/branches/master/definitions.json.zip), it ran clean, but then running SUSHI on the output caused _lots_ of errors. They were all due to actual issues in the source definitions, but one that came up a lot was due to some elements not specifying a `type` (but _used_ like a `BackboneElement`).  These caused GoFSH to create `AddElement` rules with a blank type, which is not valid FSH.  What do you think about detecting this situation (no type on a new element) and defaulting to using `BackboneElement` in the AddElementRule (while issuing a warning about it)?  I think that would be a good change, as we try to at least create syntactically valid FSH.

mint-thompson:
@cmoesel The second item you reported is due to the fact that `AddElementRuleExtractor` is run on the root element, but the rule is discarded. So the paths for `short` and `definition` were marked as processed, and thus they were ignored by the `CaretValueRuleExtractor`. I'll change the strategy for handling the root element, since I think cardinality-related attributes are still supposed to be ignored. Are there attributes other than cardinality on the root element that should get ignored?

For the third issue: I think that defaulting to `BackboneElement` and issuing a warning is a reasonable strategy, so I'll implement that.

cmoesel:
@mint-thompson -- Cardinality attributes are the only attributes I can think of that are basically meaningless on the root element of a logical/resource.  So I think ignore those but anything else is fair game?  At least try it that way and see if the output seems reasonable.

mint-thompson:
@cmoesel I tried using these updates with the IG you mentioned above, and I think it's doing what we want it to do: I saw warnings on elements I expected for defaulting to `BackboneElement` on the `HIVCaseReport` logical model, and I saw a `CaretValueRule` for the root element's `short` text on each of the applicable logical models.

src/processor/StructureDefinitionProcessor.ts
```diff
@@ -134,14 +134,22 @@ export class StructureDefinitionProcessor {
...
+
         if (isNewSlice) {
           logger.warn(
             `${target.constructorName} ${target.name} contains a slice definition for ${element.sliceName} on ${element.path}. This is not supported by FHIR.`
           );
           newRules.push(ContainsRuleExtractor.process(element, input, fisher));
-        } else if (element.path !== input.name) {
-          newRules.push(addElementRule);
+        } else {
+          if (element.path === input.name) {
```
cmoesel, line 144: Sorry, I did not catch this before, but strictly speaking, I'm not sure that the root element path _has_ to match the SD `name`.  I think it is supposed to match the `type` or the tail end of the `type` if the `type` is a URL (as is the case for logical models). You could check against that, but I think it is easier and just as valid to check if the `element.path` contains a `.`.  If it doesn't, then it's a root element.

cmoesel (changes requested):
@mint-thompson -- I agree, the HIV IG is round-tripping better now and all errors are true errors that we do not need to address.  Nice work!

I still get a different result on `package-lock.json` when I run `npm install`.  Maybe someone else can try and let me know if the issue is on my side or Mint's?

I added one more small comment related to something I hadn't caught the first time (sorry).

src/processor/StructureDefinitionProcessor.ts
mint-thompson: I think it does have to match, based on the [the definition of Element.path](http://hl7.org/fhir/R4/elementdefinition-definitions.html#ElementDefinition.path).

cmoesel:
The `package-lock.json` mismatch turned out to be on my side.  Once I deleted the `node_modules` folder and ran `npm install` again, everything line up correctly.  Sorry about that!

cmoesel (approved):
I approve!  Hurray for Logicals and Resources!

ngfreiter (approved):
Looking good to me.

PR #137, opened 2021-06-25
Organize FSH files per definition by type

### LINKED ISSUE #126: Add Style (-s) option for separate files organized by type
It would be useful to have an option where each item is in its own file (like the `file-per-definition` option), but arrange those files into folders according to type.

I also think this would be the best default behavior.

jafeltra:
This PR fixes #126 by adding a new output style to GoFSH. The new output style is called `files-organized-by-type` and it outputs an individual file per FSH definition, and each file is organized into a folder based on its FSH entity type. For example, if there are two profiles created, they will be in separate files, both under the `profiles` directory.

There are two questions I'd like input on:
- Is there a clearer name than `files-organized-by-type`? I was trying to come up with a short, but clear, description for this style and that was the best I came up with.
- The issue mentions that this might be the best default option. Is there any reason we would want to avoid updating the default option? If not and if we like this style, I can update the default as part of this PR too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Options:
   -d, --dependency <dependency...> specify dependencies to be loaded using format dependencyId@version (FHIR R4 included by default)
   -v, --version                    print goFSH version
   -h, --help                       display help for command
-  -s, --style                      specify how the output is organized into files: group-by-fsh-type (default), group-by-profile, single-file, file-per-definition, files-organized-by-type
+  -s, --style                      specify how the output is organized into files: file-per-definition (default), group-by-fsh-type, group-by-profile, single-file
   -f, --fshing-trip                run SUSHI on the output of GoFSH and generate a comparison of the round trip results
   -t, --file-type                  specify which file types GoFSH should accept as input: json-only (default), xml-only, json-and-xml
   -i, --installed-sushi            use the locally installed version of SUSHI when generating comparisons with the "-f" option
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Options:
   -d, --dependency <dependency...> specify dependencies to be loaded using format dependencyId@version (FHIR R4 included by default)
   -v, --version                    print goFSH version
   -h, --help                       display help for command
-  -s, --style                      specify how the output is organized into files: group-by-fsh-type (default), group-by-profile, single-file, file-per-definition
+  -s, --style                      specify how the output is organized into files: group-by-fsh-type (default), group-by-profile, single-file, file-per-definition, files-organized-by-type
   -f, --fshing-trip                run SUSHI on the output of GoFSH and generate a comparison of the round trip results
   -t, --file-type                  specify which file types GoFSH should accept as input: json-only (default), xml-only, json-and-xml
   -i, --installed-sushi            use the locally installed version of SUSHI when generating comparisons with the "-f" option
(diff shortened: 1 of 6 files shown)


cmoesel:
> Is there a clearer name than `files-organized-by-type`?

This feels a little inconsistent compared to the other style names, _but_ I tried to think of something better and I couldn't.  So here's another option.  What if we didn't introduce a new style but just changed the `file-per-definition` style to group into folders (retiring the current `file-per-definition` behavior)?  Is that a bad idea?

As for what the default option should be... I don't have a strong opinion on the matter, but I think I do like this new behavior _better_ than `group-by-fsh-type`, so... I guess I'm in favor of changing the default?

Also... let's not forget to update FSHSchool whenever we make this change.

README.md
```diff
@@ -29,7 +29,7 @@ Options:
   -d, --dependency <dependency...> specify dependencies to be loaded using format dependencyId@version (FHIR R4 included by default)
   -v, --version                    print goFSH version
   -h, --help                       display help for command
-  -s, --style                      specify how the output is organized into files: group-by-fsh-type (default), group-by-profile, single-file, file-per-definition
+  -s, --style                      specify how the output is organized into files: group-by-fsh-type (default), group-by-profile, single-file, file-per-definition, files-organized-by-type
```
ngfreiter, line 32: I agree with @cmoesel, I think it makes sense to just make this `file-per-definition`.

src/export/FSHExporter.ts
```diff
@@ -150,32 +154,46 @@ export class FSHExporter {
     return new Map().set('resources.fsh', results);
   }
 
-  private groupAsFilePerDefinition(): Map<string, Exportable[]> {
+  private groupAsFilePerDefinition(byGroup = false): Map<string, Exportable[]> {
```
ngfreiter, line 157: If you do make this `file-per-definition`, added bonus is this function gets simpler.

src/export/FSHExporter.ts
```diff
@@ -150,32 +154,46 @@ export class FSHExporter {
...
     const files: Map<string, Exportable[]> = new Map();
     // Aliases, still get grouped into one file
     files.set('aliases.fsh', this.fshPackage.aliases);
 
     // Other definitions are each placed in an individual file
     for (const invariant of this.fshPackage.invariants) {
-      files.set(`${invariant.name}-Invariant.fsh`, [invariant]);
+      const fileDir = byGroup ? 'invariants' : '';
+      const filename = path.join(fileDir, `${invariant.name}-Invariant.fsh`);
```
ngfreiter, line 165: I don't think we still need to put the `-Invariant` after the name of the entity (this applies to all other types as well). Since the entities are in a folder which gives their type, for me that is sufficient, and having the name appended ends up being a bit redundant.

jafeltra:
I agree with you both that it makes sense to just update the `file-per-definition` style because there probably isn't much of a need for the separate flat folder vs the nested directories. I've also updated it to be the default style and adjusted the tests accordingly. This should be ready for re-review!

src/export/FSHExporter.ts
jafeltra: That makes sense! I've removed them alL!

ngfreiter (dismissed):
Looks good to me!

julianxcarter (dismissed):
This looks good to me! I also really like this new option as our new default.

jafeltra:
@ngfreiter and @julianxcarter I rebased to resolve the conflicts, but that resulted in your reviews being dismissed. I did make updates to support exporting Logicals and Resources within the rebase, but I mostly just pulled over changes @mint-thompson had added in other tests. Besides that, it should all be the same.

julianxcarter (approved):
This still looks good to me!

ngfreiter (approved):
👍 ⏲️

PR #138, opened 2021-06-29
Add option to indent rules

mint-thompson:
Completes task [CIMPL-745](https://standardhealthrecord.atlassian.net/browse/CIMPL-745).

A new command line option `--indent` is defined. When this option is provided, the `SimplifyRulePathContextsOptimizer` runs, which modifies rules on Profiles, Extensions, Logicals, Resources, and Instances to take advantage of context paths. CodeSystems are not included, as support for hierarchical codes are part of [another task](https://standardhealthrecord.atlassian.net/browse/CIMPL-744).

In order to allow for optional optimizers, optimizers can provide an `enable` function. If this function is provided, it must return true in order for the optimizer to be included in the set of optimizers that will run. The function takes one argument: an `OptimizerOptions` object. Currently, this object will just have information about the `--indent` flag, but it is designed to easily expand to handle more complex options. Having optional optimizers complicates the topological sorting of the optimizers somewhat, since now it is possible that an optimizer specified in `runBefore` or `runAfter` exists but is not enabled. Different messages are provided to the user in this situation. In general, the rule I recommend following is that if you want to put an optional optimizer into a `runAfter` list, your optimizer should also be optional, and should turn itself off when its predecessor is turned off. Putting optional optimizers into `runBefore` does not create any additional logging messages when the optional optimizers are turned off.

The optimizer will use as much of a context path as possible when modifying rules. In the future, we may wish to consider if we want to provide more detailed options for this behavior, but I think that this is fine for the standard behavior.

The new `ExportableRule` interface provides a numeric `indent` field. Rules now implement this interface instead of `Exportable`, as `ExportableRule` extends `Exportable`. Since our rules are now potentially indented, they specify their own `toFSH` implementations. The inherited implementations are used by some, but not all, of our rules.


julianxcarter (approved):
This looks good to me! Works as expected, the test cases cover everything that you've added, and the code makes sense to me. 

The only potential addition that I could think of would be making it so that this new optimizer also inserts `pathRules` where possible so that indentation could be used in more places. I don't think that juice is worth the squeeze though, and it's also not necessarily a simplification of the FSH that's being generated so I see why that wasn't included here.

src/optimizer/loadOptimizers.ts
```diff
@@ -49,9 +56,29 @@ export async function loadOptimizers(
...
+          .map(o => o.name);
+        if (offPrecedingOptimizers.length > 0) {
+          const plural = offPrecedingOptimizers.length > 1;
+          logger.error(
+            `The ${opt.name} optimizer specifies ${
+              plural ? 'optimizers' : 'an optimizer'
+            } in runAfter that ${plural ? 'are' : 'is'} not enabled: ${offPrecedingOptimizers.join(
+              ', '
+            )}`
+          );
```
cmoesel, line 76: Since this is in an `else` block, it will only be run if the `runAfter` value had zero matches.  What if a regular expression was used and some active optimizers matched but some disabled optimizers also matched?  On one hand, it feels like that's potentially still an issue and we want to log an error.  On the other hand, we wouldn't want it to trigger an error on an optimizer that basically says "run me last" because it doesn't really care what came before it; only that it runs last.  So... maybe this is fine as-is?

Or maybe we should consider if this is really an error condition at all?  Sometimes we say "runAfter" because there is a true dependency on the previous optimizer (if it doesn't run, we are toast), but sometimes we might say "runAfter" because we know that if the previous optimizer will be run, it might affect our output, so it's best to run it first (but if it's not run, that's not a big deal). So... maybe we're being too strict?  I'm honestly not sure.

src/optimizer/loadOptimizers.ts
```diff
@@ -65,9 +92,18 @@ export async function loadOptimizers(
...
-        logger.error(
-          `The ${opt.name} optimizer specifies an unknown optimizer in runBefore: ${dependedOnBy}`
+        // a succeeding optimizer may exist, but be optional and not enabled.
+        // this does not represent an error, so nothing is logged.
+        const offSucceedingOptimizers = offOptimizers.filter(o =>
+          o.name !== opt.name && dependedOnBy instanceof RegExp
+            ? dependedOnBy.test(o.name)
+            : dependedOnBy === o.name
         );
+        if (offSucceedingOptimizers.length == 0) {
```
cmoesel, line 102: Seeing this code pattern for the 3rd time, I'm starting to wonder if maybe it would be better to use `allOptimizers` when filtering for `preceedingOptimizers` and `succeedingOptimizers` and then just check which are enabled as necessary.  I think that you should at least do that here (in the `runBefore` section) as it would remove the need for most of this new code between lines 95 and 102.

src/optimizer/plugins/SimplifyInstanceNameOptimizer.ts
```diff
@@ -2,11 +2,13 @@ import { OptimizerPlugin } from '../OptimizerPlugin';
 import { Package } from '../../processor';
 import { ExportableAssignmentRule } from '../../exportable';
 import ResolveInstanceOfURLsOptimizer from './ResolveInstanceOfURLsOptimizer';
+import SimplifyArrayIndexingOptimizer from './SimplifyArrayIndexingOptimizer';
 
 export default {
   name: 'simplify_instance_names',
   description:
     'Simplify Instance names by removing the appended InstanceOf information from the name (where possible)',
+  runBefore: [SimplifyArrayIndexingOptimizer.name],
```
cmoesel, line 11: Why?  Is there any real interdependence between these two optimizers? As far as I can tell I can't see how the output of this optimizer would affect the operation of the `SimplifyArrayIndexingOptimizer`... but maybe I am missing something.

src/optimizer/plugins/SimplifyRulePathContextsOptimizer.ts
```diff
@@ -0,0 +1,62 @@
...
+          // we didn't find a context to use. so, the indent will be 0.
+          rule.indent = 0;
+          // get rid of all existing contexts, and push this rule's path on to pathContext
+          pathContext.splice(0);
+          // change soft-index marker because a matching path should use [=]
+          pathContext.push(rule.path.replace(/\[\+\]/g, '[=]'));
+        }
+      });
+    });
+  },
```
cmoesel, line 58: I had to work through this algorithm w/ pen and paper and an example to verify it makes sense... but, it does indeed make sense!  Nice job!

cmoesel (changes requested):
Nice work.  That's a great set of test cases too!

I have noted a few things to consider (some more strongly than others).  Let me know if you want to discuss any of it!

src/exportable/ExportableAddElementRule.ts
```diff
@@ -1,8 +1,15 @@
...
+export class ExportableAddElementRule extends fshrules.AddElementRule implements ExportableRule {
+  indent: number;
 
-export class ExportableAddElementRule extends fshrules.AddElementRule implements Exportable {
   constructor(path: string) {
     super(path);
   }
+
+  toFSH(): string {
+    return `${repeat(' ', INDENT_SIZE * (this.indent ?? 0))}${super.toFSH()}`;
```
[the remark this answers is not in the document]
mint-thompson: Currently, all of the `toFSH()` methods in SUSHI for rules return one FSH rule. The other consideration is cases where `toFSH()` will contain output with newlines. Currently, this is only the case for SUSHI's `ContainsRule` class, and goFSH's `ExportableContainsRule` doesn't call `super.toFSH()` since an `ExportableContainsRule` contains instances of `ExportableCardRule` and `ExportableFlagRule`.

src/optimizer/loadOptimizers.ts
mint-thompson: This is a good point. I had always interpreted `runAfter` as saying "these are the necessary preconditions in order to correctly run," but that's just my interpretation, and that doesn't allow for differentiating between the cases you pointed out. In order to be able to represent those ideas, we might need something like `tryRunAfter` for things that are okay to turn off and `mustRunAfter` for things that are not okay to turn off. So something that just wants to get run after other things, but doesn't have any hard preconditions, can have a `tryRunAfter` that will match most other things.

The more I think about your second paragraph here, the more I like that idea. That is, there shouldn't be a case where running optimizer A, but not optimizer B, results in incorrect output. If that statement is correct, then it shouldn't be necessary to check for these errors at all. I'll take a look at our current set of optimizers and see if there are any that break this rule. If nothing breaks this rule, then I think a slight change to this implementation and a removal of the error case is fine.

julianxcarter (approved):

cmoesel (changes requested):
Thanks for addressing my previous comments.  I think it's looking good.  Unfortunately, however, it seems there are some issues that the tests are not catching and that I didn't catch in my review either -- but running against a real IG highlighted them.

I ran this against the [US Core 4.0.0 package](http://www.hl7.org/fhir/us/core/package.tgz) and while the GoFSH process ran pretty smoothly, SUSHI failed pretty spectacularly when I tried to compile the generated code:
image

It seems there are several things going on here, a sample of which are shown in the screenshot below (by way of doing a diff between output with and without `--indent` flag):
image

cmoesel:
I will be out for the next 4 business days.  If the issues I've raised are addressed and two reviewers approve, feel free to force-merge this ignoring my "request changes".  I just can't figure out how to remove my "request changes" now without explicitly "approving" the PR.

mint-thompson:
The latest changes have resolved all of the issues in USCore related to this optimizer. So, it should now be ready for re-review.

julianxcarter:
I had approved prior to these new changes, but I can confirm that a fishing trip with and without the indentation flag looks the same so my approval still stands!

cmoesel (changes requested):
I ran GoFSH master against US Core and then this against US Core _without_ the `--indent` flag.  I then did a diff on the resulting FSH.  There should be no meaningful differences, but I found that references were being processed differently.  For example (master on left, this PR on right):
image

Perhaps the optimizer order was affected resulting in this difference?

src/exportable/ExportableObeysRule.ts
```diff
@@ -12,7 +12,7 @@ export class ExportableObeysRule extends fshrules.ObeysRule implements Exportabl
 
   toFSH(): string {
     return `${repeat(' ', INDENT_SIZE * this.indent)}* ${
-      this.path === '.' || this.path === '' ? '' : `${this.path} `
+      this.path === '' ? '' : `${this.path} `
```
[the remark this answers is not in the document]
mint-thompson: It's mostly to avoid an issue that can arise in situations like this one:
```
* . obeys inv-1
* . ^short = "Short"
```
With these two rules in sequence, the second rule would attempt to indent, since the previous rule's path is a match of the current rule's path. So after indenting, you'd get this:
```
* . obeys inv-1
  * ^short = "Short"
```
Which is correct. However, if we output the obeys rule with no path, we get a problem, where this:
```
* obeys inv-1
* . ^short = "Short"
```
becomes this:
```
* obeys inv-1
  * ^short = "Short"
```
Since the actual paths still match, the second rule still gets indented.

It ends up as sort of a toss-up: either keep the `.` path and enable an indent (which only matters if the user asked for indentation and the next rule can use it), or go with an empty path and have additional checks in the indentation optimizer based on what we know the path will actually output as, not what the path's value is.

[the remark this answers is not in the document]
mint-thompson: Alternatively, we could avoid having to include special logic in the indentation optimizer if the path on the Obeys rule gets changed from `.` to an empty string before any optimizers run. Then the paths won't match, and you'll get valid non-indented output.

mint-thompson:
@cmoesel The output on the right side is very unusual: it looks like `AddReferenceKeywordOptimizer` didn't get run at all. I'll investigate and see what could be causing that.

src/exportable/ExportableObeysRule.ts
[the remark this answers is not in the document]
cmoesel: I think I like that last option (change Obeys rules with path "." to empty path)... Is it feasible without being messy?  Could it possibly be an optimizer itself?  It is, after all, taking valid FSH and changing it to also-valid FSH that follows best practices more closely.

mint-thompson:
Okay, I figured out the cause of the difference: on `master`, `ConstructInlineInstanceOptimizer` runs before `AddReferenceKeywordOptimizer`. On this branch, the order is reversed. `AddReferenceKeywordOptimizer` relies on checking types in the instance's StructureDefinition. When `ConstructInlineInstanceOptimizer` runs first, the rules in question are in an instance of `USCorePractitionerRoleProfile`. However, if this doesn't happen, the rules are in their original location: a `Bundle` instance, which has no idea about these contained resources. There is no enforced ordering for these two optimizers, so it's a coinflip as to whether they'll be in the order we want. The fix is easy enough: add another entry to `ConstructInlineInstanceOptimizer`'s `runBefore` list so that it always runs before `AddReferenceKeywordOptimizer`.

cmoesel (approved):
Thanks, @mint-thompson!  This looks and works great!  I did three runs of GoFSH:
* GoFSH master branch on US Core
* GoFSH PR branch _without indent_ on US Core
* GoFSH PR branch _with indent_ on US Core

I confirmed that:
* master and PR produce the same FSH (except alias ordering) when no indent is used
* running SUSHI (on PR 888 branch) results in same exact FHIR JSON between master GoFSH and PR GoFSH with indents

So-- it's all good!  Thanks!

cmoesel:
@julianxcarter -- if you have time to review this again this morning, then we can unblock Mint's other task once this is merged.  If you don't have time, I am comfortable merging as-is since you have already reviewed all the tricky bits before.  Just let me know what you want to do!

julianxcarter (approved):

PR #139, opened 2021-07-07
Fix errors in slice cardinality when extracting a contains rule

### LINKED ISSUE #135: Incorrect slice cardinality in generated FSH
If an element in the differential does not contain a `min` value, GoFSH appears to use the `min` value from the base (what is being sliced). Sometimes this results in a `slice` being marked as required when it shouldn't be.  See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Optional.20slices.20and.20goFSH

The proper behavior for determining `min` of a slice should be:
* If the differential element for the slice has a `min`, use it;
* Otherwise, if the snapshot element for the slice has a `min`, use it;
* Otherwise, default the slice min to `0`.

The proper behavior for determining `max` of a slice should be:
* If the differential element for the slice has a `max`, use it;
* Otherwise, if the snapshot element for the slice has a `max`, use it;
* Otherwise, default the `max` to `'*'`.

mint-thompson:
Fixes #135 and completes task [CIMPL-781](https://standardhealthrecord.atlassian.net/browse/CIMPL-781).

When trying to determine the cardinality of a slice, check for min and max in these places:
1. differential
2. snapshot
3. the sliced element's cardinality (only for the max) or a default 0 (only for the min)

Keep looking for a min and a max until both of them are defined. This means that the min and max may come from different sources. If we end up with an undefined max after checking all three places, we give up. This should only happen when the sliced element is not actually defined.

### DIFF
--- a/src/extractor/ContainsRuleExtractor.ts
+++ b/src/extractor/ContainsRuleExtractor.ts
@@ -32,31 +32,37 @@ export class ContainsRuleExtractor {
     // 1. Check the element differential
     // 2. Check the element snapshot
     // 3. Use defaults: min = 0, max = sliced element's max
-    const cardRule = CardRuleExtractor.process(input, structDef, fisher, false);
-    if (cardRule) {
-      containsRule.cardRules.push(cardRule);
-    } else {
-      // can we get information from the snapshot?
+    // Once both min and max are defined, the CardRule is ready to use.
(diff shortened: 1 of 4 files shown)


ngfreiter (approved):
This looks good to me!

julianxcarter (approved):

PR #140, opened 2021-07-14
Debug message added for large instances

julianxcarter:
Addresses [CIMPL-780](https://standardhealthrecord.atlassian.net/browse/CIMPL-780), adding a debug message for especially large Instances with the potential to appear stalled out during processing. A file is considered "large" if the length of its buffer array exceeds 200,000 entries when its file is first read in, in which case a `large` property is added to the corresponding WildFHIR object. This `large` property triggers the warning when instances are being processed.

This can be tested by running GoFSH against the [UTG IG](https://github.com/HL7/UTG) with the debug flag turned on.

### DIFF
--- a/src/processor/FHIRProcessor.ts
+++ b/src/processor/FHIRProcessor.ts
@@ -135,7 +135,7 @@ export class FHIRProcessor {
     instances.forEach((wild, index) => {
       try {
         if (wild.large) {
-          logger.debug(
+          logger.info(
             `Instance ${wild.content.id} is especially large. Processing may take a while.`
           );
         }
--- a/src/processor/FHIRProcessor.ts
+++ b/src/processor/FHIRProcessor.ts
@@ -134,6 +134,11 @@ export class FHIRProcessor {
     if (instances.length > 0) logger.info('Processing Instances...');
     instances.forEach((wild, index) => {
       try {
+        if (wild.large) {
+          logger.debug(
+            `Instance ${wild.content.id} is especially large. Processing may take a while.`
+          );
+        }
         resources.add(InstanceProcessor.process(wild.content, igForConfig?.content, this.fisher));
         this.outputCount(instances, index, 'Instance');
       } catch (ex) {
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -4,7 +4,14 @@ import ini from 'ini';
 import readlineSync from 'readline-sync';
 import { fhirdefs } from 'fsh-sushi';
 import { logger } from './GoFSHLogger';
-import { Package, FHIRProcessor, LakeOfFHIR, WildFHIR, FHIRResource } from '../processor';
+import {
+  Package,
+  FHIRProcessor,
+  LakeOfFHIR,
+  WildFHIR,
+  FHIRResource,
+  FileImport
+} from '../processor';
 import { FSHExporter } from '../export/FSHExporter';
 import { loadOptimizers } from '../optimizer';
 import { MasterFisher } from '../utils';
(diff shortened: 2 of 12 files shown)


src/utils/Processing.ts
```diff
@@ -201,9 +208,19 @@ function findNonDuplicateSecondaryFiles(files: string[], docs: WildFHIR[]): stri
 
 function readJSONorXML(file: string): any {
```
jafeltra, line 209: This is pretty small, but can the return type of this function be `FileImport` now?

test/helpers/stockLake.ts
```diff
@@ -9,6 +9,6 @@ export function stockLake(...paths: string[]): LakeOfFHIR {
 
 export function restockLake(lake: LakeOfFHIR, ...paths: string[]): void {
   paths.forEach(p => {
-    lake.docs.push(new WildFHIR(fs.readJSONSync(p), p));
+    lake.docs.push(new WildFHIR({ content: fs.readJSONSync(p) }, p));
```
jafeltra, line 12: I'm not sure how much it really matters since it looks like `restockLake` and `stockLake` are only used in tests, but I think this function would miss adding the `large` boolean to large files. Do you think it's worth adding a similar check for the buffer length there?

jafeltra (commented):
This looks good! I think the message is really useful, which lead to one of my questions below.

Also, the one other thought I had is about testing. Do you think it makes sense to test that that the warning is logged? That might be overkill, so I'm just curious what you think.

src/utils/Processing.ts
julianxcarter: It certainly can be! Addressed in commit `969399b`.

src/processor/FHIRProcessor.ts
```diff
@@ -134,6 +134,11 @@ export class FHIRProcessor {
     if (instances.length > 0) logger.info('Processing Instances...');
     instances.forEach((wild, index) => {
       try {
+        if (wild.large) {
+          logger.debug(
```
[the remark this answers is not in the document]
julianxcarter: I agree! I switched this to an info message in commit `969399b`.

test/helpers/stockLake.ts
julianxcarter: It wouldn't hurt, and it may actually come in handy for unit tests in the future so I like the idea. I've made this addition in `88c8e31`.

julianxcarter:
In addition to the resolved changes I've also added some unit tests that test that the `large` flag is added in correct circumstances, and that the info message is logged when the `large` flag is encountered on instances.

jafeltra (dismissed):
This looks good to me! I think this is super helpful when processing the large instances. I also tried out mCODE with this branch and none of the instances we have there trigger the flag, so the size we set as "large" seems appropriate.

ngfreiter (changes requested):
This seems to work well. How did you settle on the `buffer.length` that was a cutoff point?

src/utils/Processing.ts
```diff
@@ -199,11 +206,21 @@ function findNonDuplicateSecondaryFiles(files: string[], docs: WildFHIR[]): stri
   });
 }
 
-function readJSONorXML(file: string): any {
+export function readJSONorXML(file: string): FileImport {
   if (file.endsWith('.json')) {
-    return fs.readJSONSync(file);
+    const buffer = fs.readFileSync(file);
+    const importedFile: FileImport = { content: JSON.parse(buffer.toString()) };
+    if (buffer.length > 200000) {
```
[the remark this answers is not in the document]
julianxcarter: Addressed in commit `1959d5a`.

julianxcarter:
> This seems to work well. How did you settle on the `buffer.length` that was a cutoff point?

@ngfreiter Thanks! To find the cutoff point I just added some log statements to determine which instances specifically would cause a noticeable delay in the USCore and UTG IGs. Then I ran both of those IGs again using the debugger so I could stop and see what the buffer length of those source files were. It seemed like a buffer length of 250,000 is where I started to notice a pause in GoFSH, so I dropped it down to 200,000 just to be safe.

ngfreiter (approved):
Looks good to me!

jafeltra (approved):

PR #143, opened 2021-07-26
Add options for handling meta.profile on instances

### LINKED ISSUE #136: Handling `meta.profile` in instances
Currently, if an instance JSON/XML has any `meta.profile` entries, GoFSH will use the _first_ entry as `InstanceOf`. There are three potential problems with this:
1. If that `meta.profile` URL doesn't resolve to something in the IG or its dependencies, SUSHI will fail.
2. If there are multiple `meta.profile` entries, GoFSH is making an assumption by choosing the first one.
3. If the instance doesn't really conform to its advertised profile, SUSHI will either force it to or report errors. (Some might argue this is not actually a problem).

We should consider changing the default behavior and/or making it configurable.

**Change Default Behavior**

If the `meta.profile` has a single URL and it can be resolved to an SD, then use it as `InstanceOf`.

If the `meta.profile` has a single URL and it _cannot_ be resolved to an SD, issue a warning, set the `InstanceOf` to the base resource instead, and add the `meta.profile` entry as a caret rule.

If the `meta.profile` has multiple URLs, do not use any of them as `InstanceOf`.  Instead just use the base resource and add caret rules for all `meta.profile` entries.

**Make It Configurable**

Allow users to indicate how they want GoFSH to translate `meta.profile` entries:
* default behavior (as described above)
* always set first `meta.profile` entry as `InstanceOf`
* never set any `meta.profile` entry as `InstanceOf`

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Handling.20meta.2Eprofile.20in.20instances

mint-thompson:
Fixes #136 and completes task [CIMPL-778](https://standardhealthrecord.atlassian.net/browse/CIMPL-778).

A new command line option is added so that the user can specify their desired strategy for dealing with the `meta.profile` array on Instances. The strategies are:

* `only-one`: If there is exactly one entry in the `meta.profile` array, try to use it as InstanceOf. If it is found when fishing, do not add an assignment rule to set it in the Instance's `meta.profile` array. If it is not found when fishing, log a warning and use `resourceType` as InstanceOf. If there is not exactly one entry in the `meta.profile` array, use the `resourceType` as InstanceOf. This is the default strategy used when the command line option is not provided.
* `first`: If there is at least one entry in the `meta.profile` array, try to use it as InstanceOf. If it is found when fishing, do not add an assignment rule to set it in the Instance's `meta.profile` array. If it is not found when fishing, log a warning and use `resourceType` as InstanceOf. If there are no entries in the `meta.profile` array, use the `resourceType` as InstanceOf.
* `none`: Ignore `meta.profile` when determining InstanceOf. Always use `resourceType` as InstanceOf.

The two places where we construct Instances are `InstanceProcessor.process` and `ConstructInlineInstanceOptimizer.optimize`. These functions now receive the processing options as an argument so that they can use the selected strategy.

### DIFF
--- a/src/app.ts
+++ b/src/app.ts
@@ -16,7 +16,8 @@ import {
   logger,
   stats,
   fshingTrip,
-  getRandomPun
+  getRandomPun,
+  ProcessingOptions
 } from './utils';
 import { Package } from './processor';
 
@@ -58,6 +59,10 @@ async function app() {
       'specify which file types GoFSH should accept as input: json-only (default), xml-only, json-and-xml'
     )
     .option('--indent', 'output FSH with indented rules using context paths')
+    .option(
+      '--meta-profile <mode>',
+      'specify how meta.profile on Instances should be applied to the InstanceOf keyword: only-one (default), first, none'
+    )
     .version(getVersion(), '-v, --version', 'print goFSH version')
     .on('--help', () => {
       console.log('');
@@ -110,10 +115,19 @@ async function app() {
     process.exit(1);
   }
 
-  // Get options for optimizers (currently just the indent flag)
+  const metaProfileBehavior = program.metaProfile?.toLowerCase() ?? 'only-one';
+  if (!['only-one', 'first', 'none'].includes(metaProfileBehavior)) {
+    logger.error(
+      `Unsupported "meta-profile" option: ${metaProfileBehavior}. Valid options are "only-one", "first", and "none".`
+    );
+    process.exit(1);
+  }
+
+  // Get options for processors and optimizers
   const processingOptions = {
-    indent: program.indent === true
-  };
+    indent: program.indent === true,
+    metaProfile: metaProfileBehavior
+  } as ProcessingOptions;
 
   const processor = getFhirProcessor(inDir, defs, fileType);
   const config = processor.processConfig(dependencies);
--- a/src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
+++ b/src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
@@ -11,7 +11,7 @@ import { hasGeneratedText } from './RemoveGeneratedTextRulesOptimizer';
 import RemoveGeneratedTextRulesOptimizer from './RemoveGeneratedTextRulesOptimizer';
 import ResolveInstanceOfURLsOptimizer from './ResolveInstanceOfURLsOptimizer';
 import AddReferenceKeywordOptimizer from './AddReferenceKeywordOptimizer';
-import { MasterFisher, logger } from '../../utils';
+import { MasterFisher, logger, ProcessingOptions } from '../../utils';
 import { utils } from 'fsh-sushi';
 
 export default {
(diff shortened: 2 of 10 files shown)


src/app.ts
```diff
@@ -58,6 +59,10 @@ async function app() {
       'specify which file types GoFSH should accept as input: json-only (default), xml-only, json-and-xml'
     )
     .option('--indent', 'output FSH with indented rules using context paths')
+    .option(
+      '--meta-profile <mode>',
+      'specify how meta.profile on Instances should be applied to the InstanceOf keyword: only-one (default), first, none'
+    )
```
ngfreiter, line 65: I think this needs to be documented in the README.md (as does `--indent`). And then both options need to be documented in FSHSchool as well.

src/utils/Processing.ts
```diff
@@ -308,6 +308,8 @@ const IGNORED_NON_RESOURCE_DIRECTORIES = [
 ];
 
 export type ProcessingOptions = {
+  indent?: boolean;
+  metaProfile?: 'only-one' | 'first' | 'none';
   [key: string]: boolean | number | string;
```
ngfreiter, line 313: Why do the `ProcessingOptions` need the
```javascript
  [key: string]: boolean | number | string;
```
line? I'm not seeing where that is used.

src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
```diff
@@ -72,22 +74,54 @@ export default {
...
           }
         });
 
         let newInstance = new ExportableInstance(
           id ?? `Inline-Instance-for-${resource.id}-${++generatedIdCount}`
         );
+
+        // if a profile is available and the user wants it, try to fish it up and use it as InstanceOf
+        if (
+          (profileIndices.length === 1 && options.metaProfile !== 'none') ||
```
ngfreiter, line 94: Is this the same as `(profileIndices.length === 1 && options.metaProfile === 'only-one')`? And if so would it be more clear to use the positive?

ngfreiter (changes requested):

src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
mint-thompson: It's not exactly the same, since it's also okay if `options.metaProfile` is undefined. This shouldn't ever happen, but I wanted to implement it like this in order to represent the idea that the `only-one` behavior is the default that should happen.

src/utils/Processing.ts
mint-thompson: This is mostly just some anticipatory implementation on my part, where we might want a way to set additional primitive values so that they could be used in custom optimizers. Right now, the only place they're actually used is in a few of the `Processing` tests.

src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
ngfreiter: Ok, that makes sense.

ngfreiter (approved):
Looks good to me, thanks!

julianxcarter (approved):

PR #144, opened 2021-07-26
Skip IG Publisher validation bundle

### LINKED ISSUE #142: GoFSH Replacing Many #inline Instances w/ Blank Operation Outcomes
Running GoFSH against US Core -- and then running SUSHI -- I noticed a lot of errors like this:
```
error Element OperationOutcome.issue has minimum cardinality 1 but occurs 0 time(s).
  File: /Users/cmoesel/dev/fhir/GoFSH/gofsh-master/input/fsh/instances/AllergyIntolerance-example.fsh
  Line: 1 - 3
```

Looking at the generated FSH, it is:
```
Instance: AllergyIntolerance-example
InstanceOf: OperationOutcome
Usage: #inline
```

This is not correct.  The `AllergyIntolerance-example` resource in the US Core package I used was definitely NOT and `OperationOutcome`.  Looking at more instances, it seems that this is happening a LOT!  The pattern seems to be that it happens on instances w/ `Usage: #inline`.

Actually, looking into it further, I think the `other/validation-oo.json` file in the package is being processed.  It is a bundle and actually redefines many of these resources as OperationOutcomes!

mint-thompson:
Fixes #142 and completes task [CIMPL-799](https://standardhealthrecord.atlassian.net/browse/CIMPL-799).

The IG publisher puts some validation information into the package it produces. One of the files with validation information is a Bundle resource. However, it should be ignored when processing input.

### DIFF
--- a/src/utils/Processing.ts
+++ b/src/utils/Processing.ts
@@ -237,6 +237,10 @@ export function isProcessableContent(content: any, source?: string): content is 
     // The IG Publisher creates weird "Intersection" and "Union" SD files, so this check filters them out
     logger.debug(`Skipping temporary "comparison" resource created by IG Publisher: ${source}`);
     return false;
+  } else if (source?.endsWith(path.join('other', 'validation-oo.json'))) {
+    // The IG Publisher creates a Bundle of OperationOutcome resources based on validation results, which we don't want
+    logger.debug(`Skipping validation outcome resource created by IG Publisher: ${source}`);
+    return false;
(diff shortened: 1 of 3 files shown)


ngfreiter (approved):
Looks good to me 👍

cmoesel (approved):
Yeah, this definitely checks out!  I did all the same things I did when reporting this issue, but now all the bad stuff doesn't happen.  Yay!

PR #145, opened 2021-08-02
Hierarchical concepts and CodeCaretValueRule support

julianxcarter:
Addresses [CIMPL-744](https://standardhealthrecord.atlassian.net/browse/CIMPL-744), adding support for hierarchical concepts and concepts with `designation` or `property` values assigned. These newly supported concepts are also indented by the `SimplifyRulePathContextsOptimizer`.


src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -231,6 +240,29 @@ export class CaretValueRuleExtractor {
...
+      if (isFSHValueEmpty(rule.value)) {
+        logger.error(
+          `Value in CodeSytem ${
+            codeSystem.name ?? codeSystem.id
+          } at concept ${conceptHeirarchy.join('.')} for element ${
+            rule.caretPath
+          } is empty. No caret value rule will be created.`
+        );
+        delete caretValueRules[index];
+      }
```
mint-thompson, line 261: Could you add a test case for an empty value in a concept's caret rule? I don't think any of the current test cases will go into this block.

mint-thompson (changes requested):
Hooray for hierarchical concepts! I have a few small changes to request.

src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -231,6 +240,29 @@ export class CaretValueRuleExtractor {
     return caretValueRules;
   }
 
+  static processConcept(
+    input: any,
+    conceptHeirarchy: string[],
```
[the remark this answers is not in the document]
julianxcarter: Whoops! Thanks for catching these, this should be addressed in commit `6fe53f2`.

src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -218,6 +222,11 @@ export class CaretValueRuleExtractor {
         const caretValueRule = new ExportableCaretValueRule('');
         caretValueRule.caretPath = key;
         caretValueRule.value = getFSHValue(key, flatVS, resourceType, fisher);
+        if (resourceType === 'Concept') {
+          caretValueRule.isCodeCaretRule = true;
+          caretValueRules.push(caretValueRule);
+          return;
+        }
```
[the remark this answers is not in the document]
julianxcarter: In commit `a1b231b` I opted to just get rid of the `processConcept` function and handle everything in the `processResource` function.

src/extractor/CaretValueRuleExtractor.ts
julianxcarter: Added in commit `a1b231b`, along with some other unit tests.

ngfreiter (changes requested):
I think something seems weird with `package-lock.json`. Why is it changing when this doesn't update `package.json`. And shouldn't `package.json` now point to `2.0.0-beta.3`? Or else I think the new code is referencing properties that are not in `2.0.0-beta.1`

src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -199,16 +199,19 @@ export class CaretValueRuleExtractor {
...
     const flatVS = getPathValuePairs(input);
     Object.keys(flatVS)
       .filter(
         key =>
           !RESOURCE_IGNORED_PROPERTIES[resourceType].some(
-            property => key === property || new RegExp(`${property}(\\[\\d+\\])?\\.`).test(key)
+            property =>
+              key === property ||
+              new RegExp(`${property}(\\[\\d+\\])?\\.`).test(key) ||
+              key.startsWith(`${property}[`)
```
ngfreiter, line 214: This seems really similar to the line above, I'm wondering if maybe something like: 
```
new RegExp(`${property}(\\[\\d+\\]|\\.)`).test(key)
```
could capture all of the cases we need.

src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -199,16 +199,19 @@ export class CaretValueRuleExtractor {
   static processResource(
     input: any,
     fisher: utils.Fishable,
-    resourceType: 'ValueSet' | 'CodeSystem',
-    config: fshtypes.Configuration
+    resourceType: 'ValueSet' | 'CodeSystem' | 'Concept',
```
ngfreiter, line 202: I don't think we want to do this, the function is `processResource`, and the param is `resourceType`, so it feels off to accept a `Concept` here, since Concepts are not a type of FHIR resource.

ngfreiter: And you end up having to do a solid amount of special casing for `Concept` anyway, so I think it's probably best to just make this its own thing.

src/processor/CodeSystemProcessor.ts
```diff
@@ -31,11 +32,40 @@ export class CodeSystemProcessor {
...
+  ): (ExportableCaretValueRule | ExportableConceptRule)[] {
+    const conceptRules: (ExportableConceptRule | ExportableCaretValueRule)[] = [];
+    const newConceptRule = new ExportableConceptRule(
+      concept.code,
+      concept.display,
+      concept.definition
+    );
+    newConceptRule.hierarchy = hierarchy ?? [];
+    concept._parentId = codeSystem.name ?? codeSystem.id;
+    concept._hierarchy = [...newConceptRule.hierarchy, concept.code];
```
ngfreiter, line 54: I think the other advantage of making a new `processConcept` function on the `CaretValueRuleExtractor` is that then you can just (I think) pass hierarchy and the parent ID as params, which makes things a little bit more readable.

ngfreiter (changes requested):

package.json
```diff
@@ -80,4 +80,4 @@
     "winston": "^3.3.3",
     "yaml": "^1.10.0"
   }
-}
+}
```
ngfreiter, line 83: Prettier doesn't like this for me, and puts this newline back when I save.

package.json
```diff
@@ -68,7 +68,7 @@
     "fhir": "^4.8.3",
     "flat": "^5.0.2",
     "fs-extra": "^9.0.1",
-    "fsh-sushi": "^2.0.0-beta.1",
+    "fsh-sushi": "^2.0.0-beta.3",
```
ngfreiter, line 71: I think the package-lock.json is still a little haywire. If I check out this branch and `npm install`, I get a `package-lock.json` with changes. When you run `npm install`, does your lock file change as well? If so I think we want to go ahead and commit those changes.

src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -200,7 +200,7 @@ export class CaretValueRuleExtractor {
     input: any,
     fisher: utils.Fishable,
     resourceType: 'ValueSet' | 'CodeSystem',
-    config: fshtypes.Configuration
+    config?: fshtypes.Configuration
```
ngfreiter, line 203: This can be reverted now right?

src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -231,6 +231,42 @@ export class CaretValueRuleExtractor {
...
   }
 
+  static processConcept(
+    input: any,
+    conceptHierarchy: string[],
+    codeSystemName: string,
+    fisher: utils.Fishable
+  ): ExportableCaretValueRule[] {
+    const caretValueRules: ExportableCaretValueRule[] = [];
+    const flatVS = getPathValuePairs(input);
```
ngfreiter, line 241: Probably want to rename this variable.

ngfreiter (changes requested):

src/extractor/CaretValueRuleExtractor.ts
julianxcarter: It can!

package.json
julianxcarter: I don't get any changes. I think the issue is that I'm running npm version 7 which creates a `package-lock.json` in a format that's ever so slightly different (which can be viewed as the `lockfileVersion` property at the top of the file). This new version is apparently backwards compatible, but I can downgrade my npm version and run `npm install` again for consistency's sake.

src/processor/CodeSystemProcessor.ts
```diff
@@ -31,11 +32,43 @@ export class CodeSystemProcessor {
...
     );
     input.concept?.forEach((concept: any) => {
-      newRules.push(new ExportableConceptRule(concept.code, concept.display, concept.definition));
+      newRules.push(...this.extractConceptRules(concept, input.name || input.id, fisher));
     });
     target.rules = compact(newRules);
   }
 
+  static extractConceptRules(
+    concept: any,
```
ngfreiter, line 41: I think we could use the `fhirtypes.CodeSystemConcept` type here that we've nicely defined in SUSHI. Everything should still work fine, but we just get some of the type checking advantages in the future.

src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -231,6 +231,42 @@ export class CaretValueRuleExtractor {
...
+      .filter(
+        key =>
+          !CONCEPT_IGNORED_PROPERTIES.some(
+            property => key === property || new RegExp(`${property}(\\[\\d+\\])?\\.`).test(key)
+          )
+      )
+      .forEach(key => {
+        const caretValueRule = new ExportableCaretValueRule('');
+        caretValueRule.caretPath = key;
+        caretValueRule.value = getFSHValue(key, flatConcept, 'Concept', fisher);
```
ngfreiter, line 252: I think here you could up front change this to 
```
        caretValueRule.value = getFSHValue(`concept.${key}`, flatConcept, 'CodeSystem', fisher);
```
And then you would not need to change the `getFSHValue` function at all I think?

test/optimizer/plugins/fixtures/unsupported-codesystem.json
```diff
@@ -27,4 +25,4 @@
       }
     }
   ]
-}
+}
```
ngfreiter, line 28: Prettier is mad again. Prettier needs a constant diet of newlines.

ngfreiter (commented):
Looking good, just some final small comments!

mint-thompson (commented):
Nothing to report beyond what @ngfreiter already wrote, so I'll re-review once you handle those findings.

src/extractor/CaretValueRuleExtractor.ts
julianxcarter: This is a clever idea, but doing it this way makes getting the value within the `flatConcept` [on the first line of the `getFSHValue` function](https://github.com/FHIR/GoFSH/blob/14582d5c0ace45370328cd50f15e6edaddee9ae4/src/utils/element.ts#L56) impossible with the key parameter. I could do some clever string manipulation in that function, or change the keys within the `flatConcept` object to also have `concept.` prefixes, but I thought that both of these options felt a bit less intuitive than what I'm currently doing in the `getFSHValue` function. If you prefer either of these approaches then I can make the change!

ngfreiter: Ah good point. How it is now is fine!

ngfreiter (approved):

mint-thompson (approved):

Issue #119, opened 2021-06-01
Change from error to warning?


markkramerus:
I'm converting US Core to FSH. The IG has both JSON and XML definitions. GoFSH finds some XML files that are unrelated to resources (e.g. menu.xml) and reports errors when it can't make sense of them. I'm wondering if these errors should be warnings.
```
error Could not load FHIR-us-core.xml: Unknown resource type: specification
error Could not load fsh-generated\includes\menu.xml: Unknown resource type: ul
error Could not load input-cache\jiraspec.xml: Unknown resource type: specification
error Could not load input-cache\workgroups.xml: Unknown resource type: workgroups
```

cmoesel:
I think that's fair.  The difficulty is that these errors come from the tool we use to translate XML to JSON -- so it's hard to know when a translation exception is expected (like in these cases) or if it truly represents corrupted XML or some other significant problem.  But it's probably safe to assume that if the error we get back from conversion is "Unknown resource type" then it's probably not really supposed to be a FHIR resource anyway, in which case it shouldn't be an error.

We could also completely ignore a few known file names like `menu.xml`, `jiraspec.cml`, and `workgroups.xml`.

Last, I don't know if we provide any guidance regarding how best to use GoFSH, but a few things come to mind:
* Running GoFSH on an entire IG source folder is often not the best way to use GoFSH because it may contain duplicate files (in input and output), incomplete files (no snapshot), extra files (in input-cache and others), etc.
* For best results, run GoFSH in json-only mode (which is the default) against the `output` folder of an IG after it has been built by IG Publisher, or against an unzipped IG download package.
* If running GoFSH against the final output of an IG is not possible, run it against the `input` folder rather than against the entire IG source folder.

PR #146, opened 2021-08-10
Avoid error logging for non-FHIR XML

ngfreiter:
# Description
Fixes https://github.com/FHIR/GoFSH/issues/119.

When the package we used to parse XML FHIR files encounters a non-FHIR XML file, it would log an error like:
```
error Could not load FHIR-us-core.xml: Unknown resource type: specification
```
But when we encounter non-FHIR JSON, we don't log an error, but instead just assume we are supposed to skip this file, and log a debug message. This PR updates the XML handling so that when we encounter non-FHIR XML, we now just log a debug. We can't do this directly, since the package is kind of a black box for us, so instead we just intercept the message coming from the package and log a debug.

# Testing
To test this, you can do the following:
* On `master`, run GoFSH on a folder which contains XML files that are not FHIR. Make sure to process the XML by using `-t json-and-xml`. You should see errors like the one shown above.
* On this branch, run GoFSH on the same folder. The errors should go away (but any generated FSH should not change)
* On this branch, run GoFSH on the same folder, but with `-l debug`, and note the debug messages referring to the invalid XML.


jafeltra (approved):
Checks out ✔️

PR #150, opened 2021-08-18
Avoid error logging for JSON files with BOMs

### LINKED ISSUE #141: GoFSH processing some non-FHIR JSON files (again)
GoFSH used to process the US Core 4.0 package with only one error (related to a duplicate id).  A recent change, however, resulted in the following additional errors when processing US Core 4.0:
image

I'm not positive, but I suspect it may have been introduced w/ the PR that added warnings for large files.

julianxcarter:
# Description
Fixes #141, where GoFSH would log errors for non-FHIR files that were previously ignored. This was due to `JSON.parse()` throwing an error when it encountered files beginning with a Byte Order Mark. We now strip all JSON files of BOMs before attempting to parse them, as was done in `readJSONSync` before we began logging messages for especially large files.

# Testing
1. On the master branch, run GoFSH against the US Core or mCODE packages. You'll encounter several errors like the following: `error Could not load /Users/jacarter/.fhir/packages/hl7.fhir.us.mcode#1.0.0/package/example/.index.json: Unexpected token  in JSON at position 0`.
2. Run this branch against those same packages, and those errors will no longer appear.


mint-thompson (commented):
I think that `badBOM.json` ended up in the repo without the BOM? When I pulled it down locally, it was in regular utf-8 encoding. But, that might have been git's doing. Can you try changing the encoding on that file again? If it still ends up the same, it might be something that git is helpfully doing for us.

julianxcarter:
Thank you for catching that! I took another swing at it in commit `50c09a4`, I think the test fixture should have the correct encoding now. @mint-thompson

mint-thompson (approved):
That fixed it! Thank you!

jafeltra (approved):
BOMs strike again!

I confirmed that the new fixture file is encoded with BOM, and I ran this against mCODE and US Core and saw no errors related to JSON parsing and the output was the same!

Nice work!

PR #152, opened 2021-09-08
1.3.1 version bump

julianxcarter:
Bumps GoFSH's version to `1.3.1`, also updates the `fsh-sushi` package to version `2.0.1`. Consequently, I had to make a small update to an `ExportableAssignmentRule` test case to account for [changes made on the sushi side](https://github.com/FHIR/sushi/pull/896) to the `toString` method on FshCodes.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "gofsh",
-  "version": "1.3.0",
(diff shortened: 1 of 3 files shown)


jafeltra (approved):
Looks good to me!

PR #153, opened 2021-09-15
Update dependencies

jafeltra:
This PR updates our dependencies with reported vulnerabilities using `npm audit fix`.

Note that there is still one dependency with a moderate severity vulnerability. Our dependency `cpy-cli` has a dependency of `cpy`. `cpy` has upgraded the package that has a vulnerability but has not yet released a new version. `cpy-cli` is waiting on this new release to upgrade, as noted [here](https://github.com/sindresorhus/cpy-cli/issues/32).


cmoesel (approved):
Looks good.  Tests pass.  Ran against US Core on this branch and on master -- results were exactly the same.  Confirmed that `npm audit` only reports `glob-parent`, which we cannot fix until `cpy-cli` is updated.  Approved!

PR #155, opened 2021-10-29
Add option to load existing aliases

AymericGrassart:
GoFSH automatically generates an ```aliases.fsh``` file from the urls discovered during the parsing of FHIR resources to convert. The ```-a, --alias-file``` option allows users to specify an existing FSH file to preload aliases in order to generate consistent, predefined alias references.

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ Options:
   -t, --file-type <type>            specify which file types GoFSH should accept as input: json-only (default), xml-only, json-and-xml
   --indent                          output FSH with indented rules using context paths
   --meta-profile <mode>             specify how meta.profile on Instances should be applied to the InstanceOf keyword: only-one (default), first, none
+  -a, --alias-file <alias-filePath> specify an existing FSH file containing aliases to be loaded.
   --no-alias                        output FSH without generating Aliases
   -v, --version                     print goFSH version
   -h, --help                        display help for command
--- a/src/app.ts
+++ b/src/app.ts
@@ -9,6 +9,7 @@ import { fhirdefs, fhirtypes, utils } from 'fsh-sushi';
 import {
   ensureOutputDir,
   getInputDir,
+  getAliasFile,
   getFhirProcessor,
   getResources,
   loadExternalDependencies,
@@ -19,7 +20,8 @@ import {
   getRandomPun,
   ProcessingOptions
 } from './utils';
-import { Package } from './processor';
+import { Package, AliasProcessor } from './processor';
+import { ExportableAlias } from './exportable';
 
 const FSH_VERSION = '1.2.0';
 
@@ -63,6 +65,10 @@ async function app() {
     .option(
       '--meta-profile <mode>',
       'specify how meta.profile on Instances should be applied to the InstanceOf keyword: only-one (default), first, none'
+    )
+    .option(
+      '-a, --alias-file <alias-filePath>',
+      'specify an existing FSH file containing aliases to be loaded.'
     )
     .option('--no-alias', 'output FSH without generating Aliases')
     .version(getVersion(), '-v, --version', 'print goFSH version')
@@ -126,6 +132,12 @@ async function app() {
     process.exit(1);
   }
 
+  // Load alias file
+  let aliases: ExportableAlias[];
+  if (program.aliasFile) {
+    const aliasFile = getAliasFile(program.aliasFile);
+    aliases = AliasProcessor.process(aliasFile);
+  }
   // Get options for processors and optimizers
   const processingOptions = {
     indent: programOptions.indent === true,
@@ -151,7 +163,7 @@ async function app() {
 
   let pkg: Package;
   try {
-    pkg = await getResources(processor, config, processingOptions);
+    pkg = await getResources(processor, config, processingOptions, aliases);
   } catch (err) {
     logger.error(`Could not use input directory: ${err.message}`);
     process.exit(1);
@@ -168,6 +180,7 @@ async function app() {
   const instNum = pad(pkg.instances.length.toString(), 18);
   const invNum = pad(pkg.invariants.length.toString(), 17);
   const mapNum = pad(pkg.mappings.length.toString(), 18);
+  const aliasNum = pad(pkg.aliases.length.toString(), 18);
   const errNumMsg = pad(`${stats.numError} Error${stats.numError !== 1 ? 's' : ''}`, 12);
   const wrnNumMsg = padStart(`${stats.numWarn} Warning${stats.numWarn !== 1 ? 's' : ''}`, 12);
   const aWittyMessageInvolvingABadFishPun = padEnd(getRandomPun(stats.numError, stats.numWarn), 37);
@@ -191,6 +204,11 @@ async function app() {
     clr('║') + ' │     Instances      │    Invariants     │      Mappings      │ ' + clr('║'),
     clr('║') + ' ├────────────────────┼───────────────────┼────────────────────┤ ' + clr('║'),
     clr('║') + ` │ ${    instNum    } │ ${    invNum    } │ ${    mapNum     } │ ` + clr('║'),
+    clr('║') + ' ╰────────────────────┴───────────────────┴────────────────────╯ ' + clr('║'),
+    clr('║') + ' ╭────────────────────┬───────────────────┬────────────────────╮ ' + clr('║'),
+    clr('║') + ' │     Aliases        │                   │                    │ ' + clr('║'),
+    clr('║') + ' ├────────────────────┼───────────────────┼────────────────────┤ ' + clr('║'),
+    clr('║') + ` │ ${    aliasNum   } │                   │                    │ ` + clr('║'),
     clr('║') + ' ╰────────────────────┴───────────────────┴────────────────────╯ ' + clr('║'),
     clr('║') + '                                                                 ' + clr('║'),
     clr('╠'  + '═════════════════════════════════════════════════════════════════' +     '╣'),
--- a/src/app.ts
+++ b/src/app.ts
@@ -134,8 +134,8 @@ async function app() {
 
   // Load alias file
   let aliases: ExportableAlias[];
-  if (program.aliasFile) {
-    const aliasFile = getAliasFile(program.aliasFile);
+  if (programOptions.aliasFile) {
+    const aliasFile = getAliasFile(programOptions.aliasFile);
     aliases = AliasProcessor.process(aliasFile);
   }
   // Get options for processors and optimizers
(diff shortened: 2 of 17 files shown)


cmoesel:
Thanks for the contribution, @AymericGrassart!  We'll take a look!

cmoesel:
I just noticed that some of the CI is failing for this PR.  If you run `npm run check` locally, it should run the same basic tests that happen in CI.  Can you try to get the `npm run check` to come out clean?

AymericGrassart:
> I just noticed that some of the CI is failing for this PR. If you run `npm run check` locally, it should run the same basic tests that happen in CI. Can you try to get the `npm run check` to come out clean?

I pushed some lint fixes. `npm run check` looks clean. Let me know if there is anything else.

src/processor/AliasProcessor.ts
```diff
@@ -0,0 +1,42 @@
...
+      let aliasFileContent;
+      try {
+        aliasFileContent = fs.readFileSync(aliasFile, 'utf8');
+      } catch (e) {
+        logger.warn(`Alias file read failed with error: ${e.message}`);
+      }
+      if (aliasFileContent != null) {
+        const aliasFileContentLines = aliasFileContent.split('\n');
+        for (let i = 0; i < aliasFileContentLines.length; i++) {
+          const aliasFileContentLine = aliasFileContentLines[i];
```
julianxcarter, line 20:
```suggestion
          const aliasFileContentLine = aliasFileContentLines[I].trim();
```
Calling `trim()` on each line as we pull them out will allow for Aliases to still be detected if a user includes a leading space.

src/exportable/ExportableAlias.ts
```diff
@@ -12,4 +12,14 @@ export class ExportableAlias implements Exportable {
   toFSH(): string {
     return `Alias: ${this.alias} = ${this.url}`;
   }
+
+  static fromFSH(fshLine: string): ExportableAlias {
+    //const regex: RegExp = new RegExp('^Alias: (?<alias>\S+) *= *(?<url>\S+).*$');
```
julianxcarter, line 17: Could we remove this comment?

test/processor/AliasProcessor.test.ts
```diff
@@ -0,0 +1,40 @@
...
+      new ExportableAlias('V2-0203', 'http://terminology.hl7.org/CodeSystem/v2-0203')
+    );
+    expect(result[2].alias).toBe('US-NPI');
+    expect(result[2].url).toBe('http://hl7.org/fhir/sid/us-npi');
+    expect(result[3].alias).toBe('invalid-url');
+    expect(result[3].url).toBe('http://invalid-url');
+    expect(result[4].alias).toBe('comments');
+    expect(result[4].url).toBe('http://comments');
+    /*
+    expect(loggerSpy.getLastMessage('warn')).toMatch(
```
julianxcarter, line 35: Could we remove this commented block as well?

julianxcarter (changes requested):
Thank you for contributing this feature @AymericGrassart! I think it'd be a great addition to GoFSH, and the code you wrote fits into our codebase like a glove 🙂. I've left a few minor comments for you, but overall this looks and works great.

AymericGrassart:
Thanks for the idiom @julianxcarter, I was surprised to see how easy it was to integrate. We are using SUSHI and GoFSH and wanted to give back to the great work that you guys are doing with this tooling! Albeit a very minor, yet needed feature.
OT: We also noticed some issues around IG dependencies url resolution. Will be submitting an issue or PR when I have time to debug.

README.md
```diff
@@ -33,6 +33,7 @@ Options:
   -t, --file-type <type>            specify which file types GoFSH should accept as input: json-only (default), xml-only, json-and-xml
   --indent                          output FSH with indented rules using context paths
   --meta-profile <mode>             specify how meta.profile on Instances should be applied to the InstanceOf keyword: only-one (default), first, none
+  -a, --alias-filepath <path-to-alias-fsh-file>   specify an existing FSH file containing aliases to be loaded.
```
julianxcarter, line 36: Thank you for making this change! I was actually referring to the example argument field (`<path-to-alias-fsh-file>`) when I suggested changing the phrase to `<alias-filePath>`. Doing so provides just enough space to allow the description to appear in line with the other arguments in the readme file. I think changing the argument name in other locations in the code was a great idea though. 
```suggestion
  -a, --alias-file <alias-filePath> specify an existing FSH file containing aliases to be loaded.
```

julianxcarter (commented):
Thanks @AymericGrassart! I just had one more comment and this should be ready for an approval.

src/app.ts
```diff
@@ -123,6 +128,10 @@ async function app() {
     process.exit(1);
   }
 
+  // Load alias file
+  const aliasFile = getAliasFile(program.aliasFile);
+  const aliases = AliasProcessor.process(aliasFile);
+
```
mint-thompson, line 134: Since the alias file is optional, I think it would be useful to check to see if the option has a value before calling `getAliasFile`. With the current implementation, an info message of `Using alias file: ` will be logged when this option is not used.

src/exportable/ExportableAlias.ts
```diff
@@ -12,4 +12,13 @@ export class ExportableAlias implements Exportable {
   toFSH(): string {
     return `Alias: ${this.alias} = ${this.url}`;
   }
+
+  static fromFSH(fshLine: string): ExportableAlias {
+    const regex = /^Alias: (?<alias>\S+) *= *(?<url>\S+).*/;
```
mint-thompson, line 17: This regular expression is a little different than the FSH grammar in terms of blank space: any whitespace characters may appear between the non-whitespace tokens, and at least one whitespace character must be used to separate the tokens. For example, the following is valid FSH for defining aliases:
```
Alias: $first =		http://example.org/first
Alias: $second =	http://example.org/second
Alias:	$third  =
					http://example.org/third
```
The set of characters ` \t\r\n\f\u00A0` are what FSH defines as whitespace characters. So, the regular expression should accept all of those. I think that `/Alias:\s+(?<alias>\S+)\s+=\s+(?<url>\S+).*/` should be close to what you want to use, but should be tested to make sure that it works correctly with all of the FSH whitespace characters.

src/processor/AliasProcessor.ts
```diff
@@ -0,0 +1,42 @@
...
+        const aliasFileContentLines = aliasFileContent.split('\n');
+        for (let i = 0; i < aliasFileContentLines.length; i++) {
+          const aliasFileContentLine = aliasFileContentLines[i].trim();
+          if (aliasFileContentLine.startsWith('Alias:')) {
+            const alias = ExportableAlias.fromFSH(aliasFileContentLine);
+            if (alias != null) {
+              aliases.push(alias);
+            }
+          }
+        }
```
mint-thompson, line 27: An alias definition may be spread across multiple lines, like this:
```
Alias: $myAlias =
http://example.org/myAlias
```
So, splitting on `\n` is not going to be completely correct.

mint-thompson (changes requested):
Thank you for creating this PR, @AymericGrassart. I think this will definitely be a useful option to have in GoFSH.

src/processor/AliasProcessor.ts
AymericGrassart: Hi Mint, I have made updates per your comments. Please review the changes and confirm whether you are content with the resolutions.
Regarding the handling of multi-line alias definition, I put a basic parser in place, but wonder whether leveraging the antlr grammar in fsh-sushi would have been a better approach. Let me know your thoughts on that as I am not too familiar with antlr lexers. Thanks!

mint-thompson:
Thank you for updating this PR, @AymericGrassart . There are a few other small things that I think I want to change, and I don't want you to have to take the time to make those changes. Would it be okay with you if I made some updates to this PR?

src/app.ts
```diff
@@ -126,6 +132,12 @@ async function app() {
     process.exit(1);
   }
 
+  // Load alias file
+  let aliases: ExportableAlias[];
+  if (program.aliasFile) {
```
julianxcarter, line 137: I believe `program` should be `programOptions` both here and on line 138

julianxcarter (changes requested):
This looks great @mint-thompson! Bringing the parser over from sushi was a great idea, and it really makes this feature a lot more robust. I have one minor suggestion, but overall I think this is ready to merge.

src/app.ts
```diff
@@ -192,6 +205,11 @@ async function app() {
     clr('║') + ' ├────────────────────┼───────────────────┼────────────────────┤ ' + clr('║'),
     clr('║') + ` │ ${    instNum    } │ ${    invNum    } │ ${    mapNum     } │ ` + clr('║'),
     clr('║') + ' ╰────────────────────┴───────────────────┴────────────────────╯ ' + clr('║'),
+    clr('║') + ' ╭────────────────────┬───────────────────┬────────────────────╮ ' + clr('║'),
+    clr('║') + ' │     Aliases        │                   │                    │ ' + clr('║'),
```
cmoesel, line 209: Nitpicky, but: `Aliases` is not centered properly in the block.

cmoesel: Also, I can't decide for sure if it's actually very helpful to output the number of aliases or not.  I'd be OK w/ just deleting this row too.

Or keep it and center `Aliases` in its box.

Either way.

cmoesel (changes requested):
This is a very cool feature.  I like it.

The issue that Julian brought up definitely needs to be addressed and I added a couple of other suggestions that should at least be considered.

src/processor/Package.ts
```diff
@@ -69,6 +70,17 @@ export class Package {
     } else if (resource instanceof ExportableMapping) {
       checkDuplicateDefinition(this.mappings, resource, 'mapping');
       this.mappings.push(resource);
+    } else if (resource instanceof ExportableAlias) {
+      const duplicateAlias = this.aliases.find(e => (e.alias = resource.alias));
```
julianxcarter, line 74: We're gonna want a comparison operator here instead of the singular `=`. As it is right now, this always results in duplicates being flagged when there is more than one alias in the file.
```suggestion
      const duplicateAlias = this.aliases.find(e => (e.alias === resource.alias));
```

julianxcarter (changes requested):
One more suggestion, but otherwise this looks great to me!

julianxcarter (dismissed):

cmoesel (dismissed):
It's good!  Thank you, @AymericGrassart and @mint-thompson!

mint-thompson (approved):

cmoesel:
Thanks again @AymericGrassart and @mint-thompson.  Usually we squash-commit PRs, but since you each contributed a good amount to this PR, I decided to rebase it and group each of your commits into individual commits -- so there is one commit for @AymericGrassart's work and one commit for @mint-thompson's work.  That way you both get full credit for your work.  This required me to force-push.  I hope you don't mind!  As soon as CI passes, I will merge!

cmoesel (approved):

Issue #157, opened 2021-11-03
Generating invalid FSH from certain styles of extension 


ngfreiter:
These errors are easiest to recreate by downloading the [VRDR package](https://build.fhir.org/ig/HL7/vrdr/downloads.html) and running `gofsh -f` on it.

One set of errors look to be due to a difference in how VRDR represents certain extensions vs how SUSHI processes them. For example, the first screenshot below shows how VRDR represents an extension. The second shows an example from mCODE (built with SUSHI). In the SUSHI example, the `type` of the extension is profiled to the relevant extension (`http://hl7.org/fhir/R4/extension-condition-asserteddate`  in this case). In VRDR, the `type` is just `Extension`, and `url` is constrained. We actually disallowed this sort of behavior in SUSHI previously, and standardized it so SUSHI would only produce extensions using the approach you see in mCODE. So in these cases, I'm not sure there is a way for GoFSH to generate valid FSH without changing the content of the resulting JSON, which we avoid doing in GoFSH.

![image (1)](https://user-images.githubusercontent.com/52747882/140178212-26466e00-0da5-4766-8956-a66f6257a109.png)
![image (2)](https://user-images.githubusercontent.com/52747882/140178244-e1a6c258-c02e-4f7f-92c4-6a6ae7b7008a.png)

For another set of errors, in VRDR, there are certain extensions which have a structure like I’ve shown in the first screenshot. Note that the `sliceName` of the extension is `date-year-absent-reason`, but the `extension.url` is constrained to `year-absent-reason`, which does not match. In the second screenshot (from the Sara Alert IG, built using FSH/SUSHI), you can see that the `sliceName` and `extension.url` match exactly at `contact-of-known-case`. Again this is a decision that is built into SUSHI, when someone defines a simple extension on a complex extension, we automatically set the `url` to be the `sliceName` of the extension. So then trying to re-set the `url` will result in errors. Again, this is a case of essentially something that we can’t really accomplish in FSH, without changing the resulting JSON.

![image (3)](https://user-images.githubusercontent.com/52747882/140178452-7dfbc672-7c22-4e14-b619-790c407d55ca.png)
![image (4)](https://user-images.githubusercontent.com/52747882/140178479-5ac12de9-dff5-4d2d-8bb7-535f7304b3be.png)

I'm not sure what the best approach is in these cases. I don't think we want to purposefully modify the resulting JSON, but I'm not sure how we can generate FSH that will result in the original JSON with no errors.

PR #158, opened 2021-11-08
Better handle unusual extensions

ngfreiter:
## Description

Fixes https://github.com/FHIR/GoFSH/issues/157.

This PR adds two special cases to the `RemoveExtensionURLAssignmentRulesOptimizer`.

The first case occurs when the JSON processed by GoFSH has the pattern:
```json
{
  "id": "Extension.extension:date-month-absent-reason.url",
  "path": "Extension.extension.url",
  "type": [{ "code": "uri" }],
  "fixedUri": "month-absent-reason"
}
```
where the `url` of an inline extension is fixed to a value different that the `sliceName` of the element. This will result in a line of FSH like:
```
* extension[date-year-absent-reason].url = "year-absent-reason" (exactly)
```
which causes an error in SUSHI, since `url` has already been automatically fixed by SUSHI to be `data-year-absent-reason`. This PR updates the optimizer to log an error, and prefer the `sliceName` by removing the rule which fixes `url`. So no rule like the line of FSH above will be generated. Note that this doesn't change the JSON generated by SUSHI, since this rule was ignored by SUSHI as an error anyway. But this change places the error in GoFSH instead of SUSHI, and makes it more clear.

The second case occurs when the JSON processed by GoFSH has the pattern:
```json
{
  "id": "Observation.extension:injuryLocation",
  "path": "Observation.extension",
  "sliceName": "injuryLocation",
  "min": 0,
  "max": "1"
},
{
  "id": "Observation.extension:injuryLocation.url",
  "path": "Observation.extension.url",
  "min": 1,
  "max": "1",
  "fixedUri": "http://hl7.org/fhir/us/vrdr/StructureDefinition/Observation-Location"
},
```
In this case, the URL of the extension is being indicated by the `fixedUri`. This will result in FSH like:
```
* extension contains injuryLocation 0..1
* extension[injuryLocation].url = "http://hl7.org/fhir/us/vrdr/StructureDefinition/Observation-Location" (exactly)
```
But SUSHI processes this as if you are trying to set an inline extension, which is not allowed on profiles, and results in errors. Instead the FSH should look like:
```
extension contains ObservationLocation named injuryLocation 0..1
```
where `ObservationLocation` is the name of the extension with url `http://hl7.org/fhir/us/vrdr/StructureDefinition/Observation-Location`. This PR updates GoFSH to log an error when it encounters this situation, and generate FSH using the `named` keyword, which allows SUSHI to process without error. Note that this is a little unusual, because we are purposefully creating FSH which generates JSON which will differ from the original JSON. But the JSON generated by SUSHI will have the same meaning as the original, and uses the preferred style, so I think it is justified in this case.

## Testing

To test this, I suggest downloading the [VRDR package](https://build.fhir.org/ig/HL7/vrdr/downloads.html). Then on `master`, run `gofsh -f -o build-master` on it. You should see a clean `gofsh` run, but `sushi` errors. Then on this branch, run `gofsh -f -o build-branch`. You should see the errors added in this PR logged from `gofsh`, and `sushi` should finish with only two errors (unrelated to this PR).

Then compare the generated FSH in `build-master` and `build-branch` and ensure it makes sense.

Then compare the SUSHI generated JSON in `build-master` and `build-branch` and ensure it also makes sense.

### DIFF
--- a/src/optimizer/plugins/RemoveExtensionURLAssignmentRules.ts
+++ b/src/optimizer/plugins/RemoveExtensionURLAssignmentRules.ts
@@ -44,9 +44,19 @@ export default {
               } else if (urlRule.value !== item.name) {
                 logger.error(
                   `${sd.name}: Inline extension at "${rule.path}[${item.name}]" has sliceName "${item.name}" but "${urlRule.value}" is assigned to url. ` +
-                    'SUSHI requires that these values match exactly, so the url assignment will be ignored, ' +
-                    `and the sliceName of "${item.name}" will be assigned to the url.`
+                    'SUSHI requires that these values match exactly, so the value assigned to the url will be used as the sliceName.'
                 );
+                // If the value set by the url rule and the sliceName do not match, prefer the value set by the url rule
+                // and replace all occurrences of the old sliceName in other rules
+                const newSliceName = urlRule.value as string;
+                [...sd.rules, ...rule.cardRules, ...rule.flagRules].forEach(r => {
+                  r.path = r.path.replace(
+                    new RegExp(`^${rule.path}\\[${item.name}\\]`),
+                    `${rule.path}[${newSliceName}]`
+                  );
+                });
+
+                item.name = newSliceName;
               }
             }
           });
--- a/src/optimizer/plugins/RemoveExtensionURLAssignmentRules.ts
+++ b/src/optimizer/plugins/RemoveExtensionURLAssignmentRules.ts
@@ -1,7 +1,10 @@
 import { OptimizerPlugin } from '../OptimizerPlugin';
 import { Package } from '../../processor';
 import { ExportableAssignmentRule, ExportableContainsRule } from '../../exportable';
+import { logger } from '../../utils';
 import { pullAt } from 'lodash';
+import { isUri } from 'valid-url';
+import { utils } from 'fsh-sushi';
 
 // Loop over all profiles and extensions, removing assignment rules on inline extensions
 // NOTE: Inline extensions on a profile are allowed by SUSHI, but they are technically not
@@ -11,7 +14,7 @@ export default {
   description:
     'Remove assignment rules on inline extension "url" paths since SUSHI automatically applies these',
 
-  optimize(pkg: Package): void {
+  optimize(pkg: Package, fisher: utils.Fishable): void {
     [...pkg.profiles, ...pkg.extensions].forEach(sd => {
       const rulesToRemove: number[] = [];
       sd.rules.forEach(rule => {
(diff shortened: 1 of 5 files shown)


cmoesel (changes requested):
> This PR updates the optimizer to log an error, and prefer the sliceName by removing the rule which fixes url.

I think we actually should go the other way, preferring the fixed `url` and changing the `sliceName`.  The fixed `url` is used in the actual instance of the extension, so if we change it, we invalidate all existing instances of that extension.  The `sliceName`, on the other hand, is just a detail of the metadata and does not affect the actual serialization of the extension.

Changing the slice name is probably a lot harder, but I think it's the right way to go. Although now that I think of it, it's kind of unfortunate that we rely on the slice name and fixed URL to match -- because I don't think the spec requires it.  So I'd also be open to an approach that loosens things up to allow for this use case and better preserves the round-trip (but that probably requires fixes to SUSHI).

ngfreiter:
Ok @cmoesel, updated to prefer the `url` and not the `sliceName` in https://github.com/FHIR/GoFSH/pull/158/commits/505e9149f31c492d7411696b633964ca8e71fb5b.

cmoesel (commented):
I've tried running the test using VRDR as you suggested -- but I am getting different results than you described.  While I see that your fixes have been applied, I am ending up with **7** SUSHI errors even on this PR branch (vs the **2** you said to expect).

The extra errors all stem from this root cause error:
> error Cannot create injuryLocation extension; unable to locate extension definition for: ObservationLocation.

This is based on this FSH:
```
* extension contains ObservationLocation named injuryLocation 0..1
```

Looking into it, the problem is that `Observation` is defined as a `Profile` with `Parent: Extension` rather than being defined as an `Extension`.  This seems like another issue that we should investigate and fix -- although I am confused that you didn't encounter it yourself.

The other comment I have is that when we produce a _named_ extension (i.e., adding the `type.profile`), I wonder if that should be a _warning_ rather than an error -- since it is semantically the same and actually improves best practice.  Logging it as an error makes it sound bad when, in some ways, it's good.

ngfreiter:
> I've tried running the test using VRDR as you suggested -- but I am getting different results than you described. While I see that your fixes have been applied, I am ending up with **7** SUSHI errors even on this PR branch (vs the **2** you said to expect).
> 
> The extra errors all stem from this root cause error:
> 
> > error Cannot create injuryLocation extension; unable to locate extension definition for: ObservationLocation.
> 
> This is based on this FSH:
> 
> ```
> * extension contains ObservationLocation named injuryLocation 0..1
> ```
> 
> Looking into it, the problem is that `Observation` is defined as a `Profile` with `Parent: Extension` rather than being defined as an `Extension`. This seems like another issue that we should investigate and fix -- although I am confused that you didn't encounter it yourself.
> 
> The other comment I have is that when we produce a _named_ extension (i.e., adding the `type.profile`), I wonder if that should be a _warning_ rather than an error -- since it is semantically the same and actually improves best practice. Logging it as an error makes it sound bad when, in some ways, it's good.

Ah I actually do see those errors. I had fixed that issue in the package early on in my testing, and forgotten I did it. Those errors happen because the `ObservationLocation` SD is for an `Extension`, but has `kind` set to `resource`, which I think is pretty clearly an error. I think in this case we should ignore the `kind` and only trust the `type`. Do you think it should be an error or warning?

cmoesel:
> I think in this case we should ignore the `kind` and only trust the `type`.
What do you mean by "in this case" and where would this change (to ignore the `kind` go)?  Are you suggesting that SUSHI should allow `extension contains` with a profile of an `Extension` that has `kind: resource`?  I think we fish for extensions here, so would that mean changing the fisher?  Might there be negative consequences if the fisher can return extensions w/ `kind: resource`?  Or am I misunderstanding what you're suggesting?

> Do you think it should be an error or warning?

If you're referring to my suggestion that adding `type.profile` could be an error or a warning, I think I'd lean toward _warning_ myself, but I'm willing to be wrong on this one if you disagree.

cmoesel:
Oh wait.  I should have read my own previous comment before commenting again.  I was thinking that the bad extension definition (`kind: resource`) was in a dependency.  But it's in the project, so I suspect you mean to ignore kind and generate and `Extension` rather than a `Profile`.  That makes a lot more sense!

Yeah, I think we should do that. Even if you do want to create a _profile_ on an extension, this would _not_ change its kind.  So I think any definition with `type: Extension` and `kind: resource` is flat out wrong -- so I think it's OK to fix it even if it breaks the round trip...  For this one, I would log an... error?  It' s definitely wronger (?) than the previous issue.

ngfreiter:
Ok @cmoesel, the latest two commits should address your comments.

julianxcarter (approved):

cmoesel (approved):
I've retested this and everything checks out.  Definitely an improvement!  Thanks, @ngfreiter!

PR #159, opened 2021-11-09
Add option to generate FSH without Aliases

### LINKED ISSUE #156: Enable generation of "Stand alone" FSH
The FSH that is generated by GoFSH cannot always stand alone. The main example I am thinking of here is with aliases. The FSH snippet cannot be copy and pasted into a tool like FSHOnline, without copying the corresponding aliases that are used in that FSH. We should consider adding an option which would allow the user to specify that they do not want to break aliases out into their own definitions.

mint-thompson:
Fixes #156 and completes task [CIMPL-846](https://standardhealthrecord.atlassian.net/browse/CIMPL-846).

When the "--no-alias" flag is specified, FSH is produced without any aliases. URLs that do not resolve to a name will be used directly rather than being replaced by an alias.

Due to some normally helpful features of commander, do not add command-line options directly to the Command instance as properties. Instead, retrieve the options using the provided opts() method. This is done to avoid colliding with the existing "alias" property on the Command instance.

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,7 +8,7 @@ For more information about the evolving FSH syntax see the [FHIR Shorthand Refer
 
 # Installation for GoFSH Users
 
-GoFSH requires [Node.js](https://nodejs.org/) to be installed on the user's system.  Users should install Node.js 12 (LTS), although the previous LTS version (Node.js 10) is also expected to work.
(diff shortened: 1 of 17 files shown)


ngfreiter (changes requested):
One minor comment, otherwise looks good!

ngfreiter (approved):

julianxcarter (approved):

PR #166, opened 2021-12-20
SUSHI to 2.2.3 Plus Bonus Dependency Updates

cmoesel:
Bumps SUSHI to 2.2.3 and updates other dependencies reported by npm audit.

Updates to eslint and jest also required changes to their configs.  I removed `no-use-before-define` from the eslint config in favor of the default settings.  I think we're OK since `tsc` catches the important ordering issues.


mint-thompson:
I'm getting one warning from `npm audit` after a fresh install on this branch. The offending package is `glob-parent`, which is a deep dependency of `cpy-cli`. npm provided this page with more information: https://github.com/advisories/GHSA-ww39-953v-wcq6

cmoesel:
Yes, this is a long-standing issue w/ `cpy-cli`.  The dependency is very deep, so it's not simple to fix.  But... since we are in full control over how `cpy-cli` is used, I don't believe we are at risk by not addressing this specific issue.

cmoesel:
In fact, this was also noted in [one of Julia's previous PRs](https://github.com/FHIR/GoFSH/pull/153) updating dependencies for GoFSH.  It's sad that `cpy-cli` has not yet addressed it though. :-/  At some point I guess we can consider switching to a different library.

mint-thompson (dismissed):

cmoesel:
Closing this PR in favor of #168, which contains the same basic changes but without the complicated conflict in package-lock.json.

PR #167, opened 2021-12-20
Reserved path characters removed from filenames

julianxcarter:
Addresses [CIMPL-870](https://standardhealthrecord.atlassian.net/browse/CIMPL-870), replacing reserved path characters with `-` to prevent unsavory file writing behavior. This PR adds this behavior to `makeNameSushiSafe` which previously just removed whitespace from entity names, and renames the function to `sanitizeName`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -13270,6 +13270,14 @@
         }
       }
     },
+    "sanitize-filename": {
+      "version": "1.6.3",
+      "resolved": "https://registry.npmjs.org/sanitize-filename/-/sanitize-filename-1.6.3.tgz",
+      "integrity": "sha512-y/52Mcy7aw3gRm7IrcGDFx/bCk4AhRh2eI9luHOQM86nZsqwiRkkq2GekHXBBD+SmPidc8i2PqtYZl+pWJ8Oeg==",
+      "requires": {
+        "truncate-utf8-bytes": "^1.0.0"
+      }
+    },
     "saxes": {
       "version": "3.1.11",
       "resolved": "https://registry.npmjs.org/saxes/-/saxes-3.1.11.tgz",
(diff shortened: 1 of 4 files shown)


cmoesel (changes requested):
This looks like it affects the actual resource names, not just the filenames.  In other words, if a profile is named `My/Profile` in the original FHIR source, then this will produce FSH that names the profile `My-Profile` instead.  This would negatively impact the round-trip (since it changes the actual name in the definition).

We usually try to keep the round-trip intact, when possible.  Would it be possible to sanitize _only_ the filename when we write it out?  In other words, the FHIR definition name should stay the same (`/` and all), but when we write out the file to disk, we would replace the bad chars in the file name with `-`.  This way we're addressing the problem (preventing unintended path traversal) without affecting the FHIR definitions.

julianxcarter:
> This looks like it affects the actual resource names, not just the filenames. In other words, if a profile is named `My/Profile` in the original FHIR source, then this will produce FSH that names the profile `My-Profile` instead. This would negatively impact the round-trip (since it changes the actual name in the definition).
> 
> We usually try to keep the round-trip intact, when possible. Would it be possible to sanitize _only_ the filename when we write it out? In other words, the FHIR definition name should stay the same (`/` and all), but when we write out the file to disk, we would replace the bad chars in the file name with `-`. This way we're addressing the problem (preventing unintended path traversal) without affecting the FHIR definitions.

That was my mistake! I've updated this to only affect filenames, and leave the actual resources alone.

cmoesel (approved):
This looks great!  I also tested it using a simple project and confirmed it works.  Thanks, @julianxcarter!

cmoesel:
In the interest of getting a release out today, I will merge this with 1 1/2 reviews.  @jafeltra was able to briefly review the code -- and although she did not spend as much time as she normally would, everything "seemed reasonable".  I think that plus my more thorough review is sufficient for this PR.

PR #178, opened 2022-03-14
Removed cpy-cli, added copyfiles, and changed build script

### LINKED ISSUE #171: Replace cpy-cli w/ copyfiles
The `cpy-cli` package has vulnerabilities in its dependencies, but it appears cpy-cli will not fix them.  As a result, we need to move to something else.  This was already done for SUSHI, so we should just do the same here.

See fhir/sushi#1012.

guhanthuran:
Fixes #171 and completes task CIMPL-894

Check https://github.com/FHIR/sushi/pull/1012 for reasoning as to why this dependency was changed. 

To test, run `npm clean-install`, which will clear out your `node_modules` directory and install based on the updated package requirements. Then, run `npm build`, and check to see that the `dist/utils` directory contains all the files from `src/utils`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "GoFSH is a FHIR Shorthand (FSH) decompiler, able to convert formal FHIR definitions from JSON to FSH.",
   "main": "dist/index.js",
   "scripts": {
-    "build": "rm -rf dist && tsc && copyfiles -u 3 \"src/utils/template.html/*\" dist/utils",
+    "build": "rm -rf dist && tsc && copyfiles -u 2 \"src/utils/template.html\" dist/utils",
     "check": "npm run test && npm run lint && npm run prettier",
     "test": "jest --maxWorkers=4 --coverage",
     "test:watch": "npm run test -- --watchAll",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -4,7 +4,7 @@
   "description": "GoFSH is a FHIR Shorthand (FSH) decompiler, able to convert formal FHIR definitions from JSON to FSH.",
   "main": "dist/index.js",
   "scripts": {
-    "build": "rm -rf dist && tsc && cpy src/utils/template.html dist/utils",
+    "build": "rm -rf dist && tsc && copyfiles -u 3 \"src/utils/template.html/*\" dist/utils",
```
mint-thompson, line 7: Since this is just copying one file, I think this needs to be changed to not use a glob.
```suggestion
    "build": "rm -rf dist && tsc && copyfiles -u 2 \"src/utils/template.html\" dist/utils",
```

mint-thompson (changes requested):

cmoesel (dismissed):
See @mint-thompson's suggestion.

mint-thompson (approved):
Thank you!

PR #181, opened 2022-04-11
Integrate FPL

jafeltra:
This PR integrates FPL with GoFSH. The only changes here are to use FPL's `mergeDependency` function instead of the old one from SUSHI and update how the tests load packages and track which packages are loaded similar to how SUSHI's test work now.

This PR requires the updated version of SUSHI ~because it contains the updated `FHIRDefinitions` class~. You can test this PR by locally linking the `separate-package-loading` SUSHI branch on this branch. Until the next SUSHI version is released, I don't think this can be fully updated for the checks to pass here, so I've left this as a draft PR for now, but it can/should still be reviewed now.

Edit: This PR still requires the updated version of SUSHI (3.x), but it is because of how the `Type` enum changed. [846ef42](https://github.com/FHIR/GoFSH/pull/181/commits/846ef42976acd4078213f6839a5047ba6b22078e) updates to use FPL's `FHIRDefinitions` class and implements its own `fishForMetadata` method, but there are still places in SUSHI where functions are called using SUSHI's version of `Type` but GoFSH (and FPL) are expecting the updated `Type` values. Therefore, this PR needs to have a SUSHI version who has the same version of the `Type` enum as GoFSH and FPL. This can still be tested locally by linking an updated version of SUSHI here, but it shouldn't be merged until we can update the branch's version of SUSHI.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1723,6 +1723,14 @@
       "resolved": "https://registry.npmjs.org/at-least-node/-/at-least-node-1.0.0.tgz",
       "integrity": "sha512-+q/t7Ekv1EDY2l6Gda6LLiX14rU9TV20Wa3ofeQmwPFZbOMo9DXrLbOjFaaclkXKWidIaopwAObQDqwWtGUjqg=="
     },
+    "axios": {
+      "version": "0.21.4",
+      "resolved": "https://registry.npmjs.org/axios/-/axios-0.21.4.tgz",
+      "integrity": "sha512-ut5vewkiu8jjGBdqpM44XxjuCjq9LAKeHVmoVfHVzy8eHgxxq8SbAVQNovDA8mVi05kP0Ea/n/UzcSHcTJQfNg==",
+      "requires": {
+        "follow-redirects": "^1.14.0"
+      }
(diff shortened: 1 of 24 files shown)


julianxcarter (approved):
Ran this with the `next` branch on SUSHI and everything works well!

cmoesel (changes requested):
When I used `npm link` all of the tests failed pretty miserably (due to antlr errors).  Instead I did `npm pack` on sushi:next and then installed that into GoFSH.  Then the tests all passed.  I don't think there's anything to change here; I'm just noting that.

I experimented w/ this using US Core and mCODE.  Technically, everything went well; there were no differences in the output files.

But... just to be sure, I cleared my fhir cache before using this PR branch.  Everything still worked, but that was when I noticed that I wasn't getting any logs from the FPL; just long pauses when it downloaded dependencies.  When the version on master has to download dependencies, it emits logs telling you what's going on.  I think that would be nice here too -- so you probably need to just pass in a log function?

cmoesel (approved):
Thanks for integrating the logger!  It looks good!  Nice work!

PR #182, opened 2022-04-26
Cimpl 919 support passing fhir version as arg

### LINKED ISSUE #177: Support passing fhirVersion as argument
In some cases, users might point GoFSH at a set of instances that don't declare they're FHIR version.  In this case, it would be nice for GoFSH to allow the user to pass in the required FHIR version as an argument.  Since `-f` and `-v` are already taken, perhaps it could be `-u`/`--useFHIRVersion`. E.g.:
```
gofsh /path/to/resources -u 5.0.0-snapshot1
```

See: https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/GoFSH.20and.20R5/near/274566527

guhanthuran:
Fixes #177 

To test:
Try out the -u option, which lets the user decide what FHIR version to use.
For example: `ts-node src/app.ts [path_to_file] -u 5.0.0`

Suggestions for testing: try an empty input file, an populated file with missing FHIR Version, and a file where the file FHIR version is different from the input FHIR version. 

Warning: this comes pre-merged with Julia's changes, so take a look at `app.ts`, `FHIRProcessor.ts`, and `ConfigurationEtractor.ts` if you want the main changes regarding the `-u` option. 

One more note: should the error on line 48 in `ConfigurationEtractor.ts` be a warning or is it good as an error? Would appreciate your input. (I'm personally leaning towards warn since GoFSH can still work with a differing input FHIR version in some cases.)

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,7 +29,8 @@
     "Julian Carter <jacarter@mitre.org",
     "Nick Freiter <nfreiter@mitre.org>",
     "Chris Moesel <cmoesel@mitre.org>",
-    "Mint Thompson <mathompson@mitre.org>"
+    "Mint Thompson <mathompson@mitre.org>",
+    "Guhan B. Thuran <gthuran@mitre.org>"
   ],
   "license": "Apache-2.0",
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "GoFSH is a FHIR Shorthand (FSH) decompiler, able to convert formal FHIR definitions from JSON to FSH.",
   "main": "dist/index.js",
   "scripts": {
-    "build": "rm -rf dist && tsc && cpy src/utils/template.html dist/utils",
+    "build": "rm -rf dist && tsc && copyfiles -u 2 \"src/utils/template.html\" dist/utils",
     "check": "npm run test && npm run lint && npm run prettier",
     "test": "jest --maxWorkers=4 --coverage",
     "test:watch": "npm run test -- --watchAll",
@@ -51,7 +51,7 @@
     "@types/valid-url": "^1.0.3",
     "@typescript-eslint/eslint-plugin": "^5.8.0",
     "@typescript-eslint/parser": "^5.8.0",
-    "cpy-cli": "^3.1.1",
+    "copyfiles": "^2.4.1",
     "eslint": "^8.5.0",
     "eslint-config-prettier": "^6.10.1",
     "jest": "^27.4.5",
--- a/src/app.ts
+++ b/src/app.ts
@@ -7,6 +7,7 @@ import chalk from 'chalk';
 import { pad, padStart, padEnd } from 'lodash';
 import { fhirdefs, fhirtypes, utils } from 'fsh-sushi';
 import {
+  determineCorePackageId,
   ensureOutputDir,
   getInputDir,
   getAliasFile,
@@ -14,7 +15,6 @@ import {
   getResources,
   loadExternalDependencies,
   writeFSH,
-  useGivenFhirVersion,
   logger,
   stats,
   fshingTrip,
@@ -23,6 +23,7 @@ import {
 } from './utils';
 import { Package, AliasProcessor } from './processor';
 import { ExportableAlias } from './exportable';
+import { isSupportedFHIRVersion } from 'fsh-sushi/dist/utils';
 
 const FSH_VERSION = '2.0.0';
 
@@ -118,8 +119,11 @@ async function app() {
   const dependencies = programOptions.dependency?.map((dep: string) => dep.trim());
 
   // Use specified FHIR Version
-  //logger.info(`${programOptions.useFHIRVersion}`);
-  useGivenFhirVersion(programOptions.useFHIRVersion);
+  const specifiedFHIRVersion = programOptions.useFHIRVersion;
+  if (specifiedFHIRVersion && !isSupportedFHIRVersion(specifiedFHIRVersion)) {
+    logger.error(`Specified FHIR version is invalid: ${specifiedFHIRVersion}`);
+    process.exit(1);
+  }
 
   // Load FhirProcessor and config object
   const fileType = programOptions.fileType?.toLowerCase() ?? 'json-only';
@@ -152,18 +156,14 @@ async function app() {
   } as ProcessingOptions;
 
   const processor = getFhirProcessor(inDir, defs, fileType);
-  //pass in here
-  const config = processor.processConfig(dependencies);
+  const config = processor.processConfig(dependencies, specifiedFHIRVersion);
 
   // Load dependencies from config for GoFSH processing
   const allDependencies =
     config.config.dependencies?.map(
       (dep: fhirtypes.ImplementationGuideDependsOn) => `${dep.packageId}@${dep.version}`
     ) ?? [];
-  //check fhirVersion below (we can exit)
-  const fhirPackageId = config.config.fhirVersion[0].startsWith('4.0')
-    ? 'hl7.fhir.r4.core'
-    : 'hl7.fhir.r5.core';
+  const fhirPackageId = determineCorePackageId(config.config.fhirVersion[0]);
   allDependencies.push(`${fhirPackageId}@${config.config.fhirVersion[0]}`);
   const dependencyDefs = loadExternalDependencies(defs, allDependencies);
 
--- a/src/app.ts
+++ b/src/app.ts
@@ -14,6 +14,7 @@ import {
   getResources,
   loadExternalDependencies,
   writeFSH,
+  useGivenFhirVersion,
   logger,
   stats,
   fshingTrip,
@@ -71,6 +72,7 @@ async function app() {
       'specify an existing FSH file containing aliases to be loaded.'
     )
     .option('--no-alias', 'output FSH without generating Aliases')
+    .option('-u, --useFHIRVersion <fhirVersion>', 'specifiy which FHIR version to run instances on')
     .version(getVersion(), '-v, --version', 'print goFSH version')
     .on('--help', () => {
       console.log('');
@@ -115,6 +117,10 @@ async function app() {
   // Trim empty spaces from command line dependencies
   const dependencies = programOptions.dependency?.map((dep: string) => dep.trim());
 
+  // Use specified FHIR Version
+  //logger.info(`${programOptions.useFHIRVersion}`);
+  useGivenFhirVersion(programOptions.useFHIRVersion);
+
   // Load FhirProcessor and config object
   const fileType = programOptions.fileType?.toLowerCase() ?? 'json-only';
   if (!['json-only', 'xml-only', 'json-and-xml'].includes(fileType)) {
@@ -146,6 +152,7 @@ async function app() {
   } as ProcessingOptions;
 
   const processor = getFhirProcessor(inDir, defs, fileType);
+  //pass in here
   const config = processor.processConfig(dependencies);
 
   // Load dependencies from config for GoFSH processing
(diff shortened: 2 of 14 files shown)


cmoesel:
@guhanthuran -- you said:

> Warning: this comes pre-merged with Julia's changes, so take a look at app.ts, FHIRProcessor.ts, and ConfigurationEtractor.ts if you want the main changes regarding the -u option.

But I don't think I see any of Julia's changes in this PR. I assume you're referring to Julia's changes that were merged already (so now they don't show up here since they're already in `master`).

test/extractor/ConfigurationExtractor.test.ts
```diff
@@ -59,6 +59,33 @@ describe('ConfigurationExtractor', () => {
...
+      const result = ConfigurationExtractor.process([ig, ...resources], '5.0.0');
+      expect(result.config.fhirVersion).toEqual(['5.0.0']);
+    });
+
+    it('should be able to handle a specified FHIR version when FHIR version is missing', () => {
+      const ig = JSON.parse(
+        fs.readFileSync(path.join(__dirname, 'fixtures', 'missing-fhir-version-ig.json'), 'utf-8')
+      );
+      const result = ConfigurationExtractor.process([ig, ...resources], '5.0.0');
+      expect(result.config.fhirVersion).toEqual(['5.0.0']);
```
cmoesel, line 75: Here I would also check the loggerSpy to confirm no errors/warnings were logged.

cmoesel (changes requested):
I tested this out manually and everything worked as advertised.  That said, I still have a few recommendations to consider.  Some things _should_ be fixed, but some are a matter of preference, so it's up to you.

src/utils/Processing.ts
```diff
@@ -151,9 +151,18 @@ export function getLakeOfFHIR(inDir: string, fileType: string): LakeOfFHIR {
...
   const docs: WildFHIR[] = [];
+  let pluralJsonFiles = '';
+  let pluralXmlFiles = '';
+
+  if (jsonFiles.length !== 1) {
+    pluralJsonFiles = 's';
+  }
+  if (xmlFiles.length !== 1) {
+    pluralXmlFiles = 's';
+  }
```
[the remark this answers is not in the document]
guhanthuran: I think I might keep this one as is, since this part was with Julia's code not mine.

src/extractor/ConfigurationExtractor.ts
```diff
@@ -36,9 +36,25 @@ export class ConfigurationExtractor {
...
+    }
+    if (specifiedFHIRVersion) {
+      // fhirVersion can sometimes not be detected
+      if (fhirVersion?.length && fhirVersion.indexOf(specifiedFHIRVersion) === -1) {
+        // should this be error or warn?
+        logger.error(
+          `FHIR Version mismatch error: specified version is: ${specifiedFHIRVersion} while detected version is: ${fhirVersion.join(
+            ', '
+          )}`
+        );
```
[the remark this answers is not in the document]
guhanthuran: Great - made this a warning and added that sentence.

src/extractor/ConfigurationExtractor.ts
```diff
@@ -36,9 +36,27 @@ export class ConfigurationExtractor {
...
+      // If there is a detected version, check if it matches the specified version
+      if (fhirVersion?.length && fhirVersion.indexOf(specifiedFHIRVersion) === -1) {
+        logger.warn(
+          `FHIR Version mismatch error: specified version is: ${specifiedFHIRVersion} while detected version is: ${fhirVersion.join(
+            ', '
+          )}`
+        );
+        logger.warn(
+          `GoFSH will use the specified version (${specifiedFHIRVersion}), but this may result in additional processing errors.`
+        );
```
cmoesel, line 53: Instead of logging this as a _second_ warning, I'd just add this sentence to the end of the warning above -- because it's really the same basic thing.  That way it doesn't get logged w/ a warning count of 2.

cmoesel (changes requested):
Thanks for addressing my comments!  I just have one more small little tweak!

cmoesel (dismissed):
Looks good to me!  We just need a second reviewer now!

src/app.ts
```diff
@@ -116,6 +120,13 @@ async function app() {
   // Trim empty spaces from command line dependencies
   const dependencies = programOptions.dependency?.map((dep: string) => dep.trim());
 
+  // Use specified FHIR Version
+  const specifiedFHIRVersion = programOptions.useFHIRVersion;
+  if (specifiedFHIRVersion && !utils.isSupportedFHIRVersion(specifiedFHIRVersion)) {
+    logger.error(`Specified FHIR version is invalid: ${specifiedFHIRVersion}`);
```
jafeltra, line 126: Does it make sense to say "Specified FHIR version is not supported" here instead? Maybe it is a valid FHIR version but just one that we don't support?

src/extractor/ConfigurationExtractor.ts
```diff
@@ -36,9 +36,24 @@ export class ConfigurationExtractor {
...
-        : ['4.0.1'];
+      if (utils.isSupportedFHIRVersion(fhirVersionFromResources)) {
+        fhirVersion = [fhirVersionFromResources];
+      }
+    }
+    if (specifiedFHIRVersion) {
+      // If there is a detected version, check if it matches the specified version
+      if (fhirVersion?.length && fhirVersion.indexOf(specifiedFHIRVersion) === -1) {
+        logger.warn(
+          `FHIR Version mismatch error: specified version is: ${specifiedFHIRVersion} while detected version is: ${fhirVersion.join(
```
jafeltra, line 47: Since this is a warning now, does it make sense to update the text here to say "FHIR Version mismatch warning: ..." instead?

src/utils/Processing.ts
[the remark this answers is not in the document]
jafeltra: I don't think I added this code. I could be wrong and I like what it does, so I can take credit if we want haha. But I just want to make sure we intentionally added it here. The test included below looks good too, so I don't think there's any harm and probably there is only good! I do like Chris's helper function though.

test/extractor/ConfigurationExtractor.test.ts
```diff
@@ -59,6 +59,41 @@ describe('ConfigurationExtractor', () => {
...
+      expect(loggerSpy.getAllMessages('warn')).toHaveLength(2);
+      expect(loggerSpy.getMessageAtIndex(-2, 'warn')).toMatch(
+        'ImplementationGuide missing properties needed to generate configuration file: fhirVersion. These properties will be inferred based on FHIR definitions.'
+      );
+      expect(loggerSpy.getLastMessage('warn')).toMatch(
+        'FHIR Version mismatch error: specified version is: 5.0.0 while detected version is: 4.0.1. GoFSH will use the specified version (5.0.0), but this may result in additional processing errors.'
+      );
+    });
+
+    it('should log an error when the specified FHIR version is different from detected FHIR version', () => {
```
jafeltra, line 86: This is also a really small suggestion to make the test description more accurate now that it is a warning.

```suggestion
    it('should log a warning when the specified FHIR version is different from detected FHIR version', () => {
```

test/extractor/ConfigurationExtractor.test.ts
```diff
@@ -59,6 +59,41 @@ describe('ConfigurationExtractor', () => {
...
+      const ig = JSON.parse(
+        fs.readFileSync(path.join(__dirname, 'fixtures', 'simple-ig.json'), 'utf-8')
+      );
+      const result = ConfigurationExtractor.process([ig, ...resources], '5.0.0');
+      expect(result.config.fhirVersion).toEqual(['5.0.0']);
+      expect(loggerSpy.getLastMessage('warn')).toMatch(
+        'FHIR Version mismatch error: specified version is: 5.0.0 while detected version is: 4.5.0. GoFSH will use the specified version (5.0.0), but this may result in additional processing errors.'
+      );
+    });
```
jafeltra, line 95: I think it might be good to include one more test here to cover the case when a FHIR Version is not detected and not specified so it defaults to `4.0.1`. Does that make sense to you to include? Or do you feel like that's covered elsewhere?

jafeltra (commented):
I had just a few comments. Pretty much all of them are very small changes and one is a suggestion for another test. Let me know what you think!

src/app.ts
cmoesel: I agree.  `not supported` is better than `invalid`.

src/extractor/ConfigurationExtractor.ts
cmoesel: I agree here too.  Good catch, @jafeltra!

test/extractor/ConfigurationExtractor.test.ts
cmoesel: I like that.

package.json
```diff
@@ -29,7 +29,8 @@
     "Julian Carter <jacarter@mitre.org",
     "Nick Freiter <nfreiter@mitre.org>",
     "Chris Moesel <cmoesel@mitre.org>",
-    "Mint Thompson <mathompson@mitre.org>"
+    "Mint Thompson <mathompson@mitre.org>",
+    "Guhan B. Thuran <gthuran@mitre.org>"
```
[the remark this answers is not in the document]
guhanthuran: 🥳

test/extractor/ConfigurationExtractor.test.ts
guhanthuran: I believe this was covered in the `should be able to handle a specified FHIR version when FHIR version is missing` test case in `ConfigurationExtractor.test.ts`. Let me know if we can get more specific.

src/extractor/ConfigurationExtractor.ts
guhanthuran: Done.

src/app.ts
guhanthuran: Done.

test/extractor/ConfigurationExtractor.test.ts
jafeltra: I think that case covers when a FHIR version isn't present in the resources but it is specified directly to the `ConfigurationExtractor.process` method. I was thinking of one that used those same resources that are missing a FHIR version but there is no user provided FHIR version either. Something like this:

```
const ig = JSON.parse(
  fs.readFileSync(path.join(__dirname, 'fixtures', 'missing-fhir-version-ig.json'), 'utf-8')
);
const result = ConfigurationExtractor.process([ig, ...resources]);
expect(result.config.fhirVersion).toEqual(['4.0.1']);
// ... other expects for the appropriate warnings
```

jafeltra (commented):
This looks good, I just had one really small comment that should be really easy to fix.

test/api/FhirToFsh.test.ts
```diff
@@ -96,14 +98,16 @@ describe('fhirToFsh', () => {
       /Invalid logLevel: 11. Valid levels include: silly, debug, verbose, http, info, warn, error, silent/
     );
     expect(results.warnings).toHaveLength(0);
+    // expect(results.warnings[0].message).toMatch('Could not determine FHIR version. Using 4.0.1.');
```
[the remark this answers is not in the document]
guhanthuran: Done!

jafeltra:
Everything here looks really good! When I was looking at the tests again after you added the new tests, I thought of another question. When there is an ImplementationGuide that is missing the `fhirVersion` property, there is a warning that is logged that says

```
ImplementationGuide missing properties needed to generate configuration file: fhirVersion. These properties will be inferred based on FHIR definitions.
```

Even when the `-u, --useFHIRVersion` option is used, that warning is logged. To me, that warning feels a little confusing since someone might be intentionally using the `-u` option knowing there is no FHIR version. I think we would still want the warning in the case where the `-u` option is not used and there are other properties it will apply to regardless of any CLI options used, so it might be too complicated to try to change it and not worth the effort, but I wanted to at least ask. @guhanthuran do you agree that it feels confusing? Is it worth changing the wording of the warning or when it is logged? Maybe this is even out of scope of this task? Maybe @cmoesel has thoughts on this too.

guhanthuran:
Make sure that we are logging whenever FHIR version whenever it is changed - add info statements so that user is aware.

cmoesel (approved):
Awesome.  Thank you @guhanthuran.  I think our users will appreciate this!

PR #184, opened 2022-05-31
Ignores ig-r4.json

### LINKED ISSUE #179: GoFSH should ignore ig-r4.json
When running GoFSH on some packages, it will report an error because it detects two `ImplementationGuide` files with the same `id`.  This is because the IG Publisher will produce an `ig-r4.json` file representing the same information as the original `ImplementationGuide` file (but conforming to R4).

For example, when running GoFSH on US Core 4.0.0, we get the following error:
> error Encountered 1 definition(s) with the same resourceType and id as a previous definition. FHIR definitions should have unique resourceType and id. The following duplicate definitions will not be processed by GoFSH:
>  - /Users/cmoesel/.fhir/packages/hl7.fhir.us.core#4.0.0/package/ig-r4.json (ImplementationGuide/hl7.fhir.us.core) matches /Users/cmoesel/.fhir/packages/hl7.fhir.us.core#4.0.0/package/ImplementationGuide-hl7.fhir.us.core.json

GoFSH should ignore the `ig-r4.json` file so this does not happen.

guhanthuran:
Fixes #179 

To test:
Run a test where the ig-r4.json has the same `id` as an ImplementationGuide. If it throws an error, this fix is breaking, if it is not, it's working.


jafeltra (approved):
Looks good to me! The code and test look good, and I confirmed the error in US Core is removed.

PR #187, opened 2022-06-16
Added logs for runtime arguments

guhanthuran:
Added logs for the various runtime args, if you can please just check my spelling! 😅


src/app.ts
```diff
@@ -99,6 +99,45 @@ async function app() {
...
+  if (programOptions.installedSushi) {
+    logger.info('  --installed-sushi');
+  }
+  if (programOptions.metaProfile) {
+    logger.info(`  --meta-profile ${programOptions.metaProfile}`);
+  }
+  if (programOptions.aliasFile) {
+    logger.info(`  --alias-file ${programOptions.aliasFile}`);
+  }
+  if (programOptions.alias) {
```
jafeltra, line 130: The logic for this one actually needs to be flipped. When `programOptions.alias` is `true`, we are using aliases and we have _not_ used the `--no-alias` flag. The value of `programOptions.alias` gets flipped to `false` when you do use `--no-alias`. So we want to only print the option information when we used `--no-alias`, so instead I think we need to check `if(!programOptions.alias)`.

src/app.ts
```diff
@@ -99,6 +99,45 @@ async function app() {
...
+    logger.info('  --no-alias');
+  }
+  if (programOptions.useFHIRVersion) {
+    logger.info(`  --useFHIRVersion ${programOptions.useFHIRVersion}`);
+  }
+  if (programOptions.out) {
+    logger.info(`  --out ${path.resolve(programOptions.out)}`);
+  }
```
jafeltra, line 138: This is just a question for now, but I realized that the `ensureOutputDir` function always logs the output director. Therefore, when you use the `-o --out` option, we end up logging the output directory twice - once here and once from within the `ensureOutputDir`. SUSHI's `ensureOutputDir` function works a little differently, so it only logs the output directory if it isn't set using `-o --out`, so it is always logged and only logged once. To make that work in GoFSH, we would need to update how `ensureOutputDir` works. Is that something we want to do? It might be a little out of scope of this PR. Or do we not mind the duplicate log? One does log the full path and one logs a relative path, so it might be okay to keep both. I'd like other opinions before actually asking for a change.

src/app.ts
```diff
@@ -99,6 +99,45 @@ async function app() {
...
+  if (programOptions.installedSushi) {
+    logger.info('  --installed-sushi');
+  }
+  if (programOptions.indent) {
+    logger.info('  --indent');
+  }
+  if (programOptions.installedSushi) {
+    logger.info('  --installed-sushi');
+  }
```
jafeltra, line 123: This is duplicated and can be removed.

src/app.ts
```diff
@@ -99,6 +99,45 @@ async function app() {
...
+  if (programOptions.style) {
+    logger.info(`  --style ${programOptions.style}`);
+  }
+  if (programOptions.fshingTrip) {
+    logger.info('  --fshing-trip');
+  }
+  if (programOptions.installedSushi) {
+    logger.info('  --installed-sushi');
+  }
```
jafeltra, line 117: A logged message for `-t, --file-type` is missing.

jafeltra (changes requested):
This is looking good. I noticed a few things that need to be updated and had one question for us to consider.

src/app.ts
guhanthuran: I changed the --out log message to not include the whole file directory, do you think this approach is a bit better? 

`if (programOptions.out) {
    logger.info('  --out');
  }`

jafeltra: Hm I think it would be good to keep the file path. Especially since it's the full resolved path, I think it's good context to display. It also aligns with SUSHI.

After thinking about it more, I think it's better to err on the side of more information. So I think rather than update how `ensureOutputDir` works, we can just log the path twice. So if you wouldn't mind, can you add back in the `path.resolve` portion?

guhanthuran: Sure - I'll keep it in.

src/app.ts
```diff
@@ -99,6 +99,45 @@ async function app() {
...
+  if (programOptions.fileType) {
+    logger.info(`  --file-type ${programOptions.fileType}`);
+  }
+  if (programOptions.indent) {
+    logger.info('  --indent');
+  }
+  if (programOptions.metaProfile) {
+    logger.info(`  --meta-profile ${programOptions.metaProfile}`);
+  }
+  if (!programOptions.aliasFile) {
```
jafeltra, line 127: Sorry I just noticed this, but the negation go added to the wrong `if` statement. This needs to be on the `alias` statement, not the `aliasFile` statement. Can you change this line back to just say `if (programOptions.aliasFile)` and add the negation to the one below?

src/app.ts
jafeltra: Thanks for swapping it back!

jafeltra (changes requested):

src/app.ts
guhanthuran: Done.

jafeltra (approved):
Looks good!

julianxcarter (approved):

jafeltra:
@guhanthuran once you click the "update branch" button, this is good to merge!

PR #191, opened 2022-07-12
Improve FSHing Trip for input files w/ arbitrary file names

cmoesel:
Prior to this PR, the fshing trip option assumed that input files used the `{type}-{id}.json` naming convention. As a result, it would fail to match up input files having arbitrary names w/ their corresponding output files (post-SUSHI).

Now the fshing trip will match up input files based on the filename that SUSHI would use if it were naming them; which results in inputs and outputs being properly matched up.  You can test this by running GoFSH with the `-f` option on one or more files w/ unconventional names (for example, [resprate.profile.json](http://hl7.org/fhir/R4/resprate.profile.json)). On master (or release), the resulting diff will treat the input file and corresponding output file as separate diffs with no relationship to each other.  On this branch, it will properly align them, despite the input file name.

While testing, I also noticed that fshing trip was putting `validation-oo.json` and `validation-summary.json` into the report.  These files should not be reported on, so I added them to the list of files to ignore.  You can test this by running GoFSH with the `-f` option against a package in your FHIR cache (like mCODE or US Core).


jafeltra (approved):
Looks good to me and works as expected.

julianxcarter (approved):

PR #194, opened 2022-08-25
1.6.2

jafeltra:
Version bump and new SUSHI for next GoFSH release.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
   "name": "gofsh",
-  "version": "1.6.1",
+  "version": "1.6.2",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
--- a/package-lock.json
+++ b/package-lock.json
@@ -2829,9 +2829,9 @@
       "optional": true
     },
     "fsh-sushi": {
-      "version": "2.6.0",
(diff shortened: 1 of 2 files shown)


mint-thompson (approved):
🪕

cmoesel (approved):
Yes, indeed, the number `2` comes after the number `1`.  Well done!

jafeltra:
> Yes, indeed, the number `2` comes after the number `1`. Well done!

The tricky part is there are two `1`s and yes I did accidentally name the branch `2.6.2` at first 😅

cmoesel:
> > Yes, indeed, the number `2` comes after the number `1`. Well done!
> 
> The tricky part is there are two `1`s and yes I did accidentally name the branch `2.6.2` at first 😅

Oh! Yeah, actually, in my haste, I think I misremembered the change as being 2.6.1 to 2.6.2 (in which case my comment would make more sense).  I actually wasn't intentionally calling out your initial mistake (.2 instead of .1).  I hadn't even noticed it, but in fact, had moreso bought into it.  Oops!

cmoesel:
Wait.  I am so confused.  Actually, now I see that the GoFSH version really did bump from `1.6.1` to `1.6.2`, so my initial comment was in the right context.  When I just relooked at it now, I was looking at the SUSHi dependency version.  You're right, too many `1`s.

Anyway, the release notes look good.  I'm merging it!

PR #198, opened 2022-09-29
Update logging statements to include more information

### LINKED ISSUE #197: Some errors in GoFSH do not indicate what resource/file the error is associated with
Best way to reproduce this is to download the latest R5 FHIR package and run GoFSH against it -- then look at the reported errors and find the ones that don't indicate resource or file.

Or, if you have time, review all logger error messages in the code manually.

cmoesel:
Some logging statements did not provide enough information to help the user debug the issue. Several statements have now been updated to include useful details for identifying the source of the problem being reported.

I would have liked to include filenames where possible, but the current structure of the code does not allow for that, as file names are lost pretty early in the process. Perhaps we can rework it some day, but for now, this PR should be enough of an improvement to satisfy the original request.

Fixes #197

### DIFF
--- a/src/extractor/AddElementRuleExtractor.ts
+++ b/src/extractor/AddElementRuleExtractor.ts
@@ -1,12 +1,15 @@
 import { isEmpty } from 'lodash';
-import { ProcessableElementDefinition } from '../processor';
+import { ProcessableElementDefinition, ProcessableStructureDefinition } from '../processor';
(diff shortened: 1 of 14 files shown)


julianxcarter (approved):
The changes here look good to me! I also did a check of all of our warning and error statements to see if there was anything else to expand and didn't find anything.

PR #205, opened 2022-12-20
Resolve contentReference when getting types

### LINKED ISSUE #204: ValueSet expansion.contains rules should not be optimized into a single assignment
When processing a ValueSet, the `expansion.contains` element has all of the parts that would be found on a `Coding`-type element. Thus, the `CombineCodingAndQuantityValuesOptimizer` combines them into a single assignment rule. However, `expansion.contains` is a `BackboneElement`, and SUSHI does not correctly process assignment rules against it.

The [expansion example ValueSet](https://www.hl7.org/fhir/r4/valueset-example-expansion.json) provides an example of this behavior, with GoFSH creating rules like this one:
```
* ^expansion.contains[=].contains[=] = http://loinc.org#2093-3 "Cholesterol [Mass/volume] in Serum or Plasma"
```

See also https://chat.fhir.org/#narrow/stream/215610-shorthand/topic/Valueset.20expansion

mint-thompson:
Fixes #204 and completes task [CIMPL-1048](https://standardhealthrecord.atlassian.net/browse/CIMPL-1048).

CombineCodingAndQuantityValuesOptimizer has a set list of types for which it will optimize rules. Resolve contentReference on elements to more accurately determine the type of the element at the rule's path.

Previously, since contentReference wasn't resolved, any element with a contentReference (and therefore no type) on a non-Instance would get optimized. This is why, on a ValueSet, `expansion.contains` would not get optimized (since it has a type), but the deeper versions like `expansion.contains.contains` would get optimized (since they have a contentReference to `expansion.contains`).

### DIFF
--- a/src/optimizer/utils.ts
+++ b/src/optimizer/utils.ts
@@ -165,7 +165,9 @@ function getTypesForPath(
     // NOTE: Normalize the path to remove indices and/or slice references
     const element = instanceOfSD.findElementByPath(path.replace(/\[[^\]]+\]/g, ''), fisher);
     if (element?.contentReference) {
-      return instanceOfSD.findElement(element.contentReference.slice(1))?.type;
+      return instanceOfSD.findElement(
+        element.contentReference.slice(element.contentReference.indexOf('#') + 1)
+      )?.type;
     } else {
       return element?.type;
     }
--- a/src/optimizer/utils.ts
+++ b/src/optimizer/utils.ts
@@ -164,6 +164,10 @@ function getTypesForPath(
     const instanceOfSD = fhirtypes.StructureDefinition.fromJSON(instanceOfDef);
     // NOTE: Normalize the path to remove indices and/or slice references
     const element = instanceOfSD.findElementByPath(path.replace(/\[[^\]]+\]/g, ''), fisher);
-    return element?.type;
+    if (element?.contentReference) {
+      return instanceOfSD.findElement(element.contentReference.slice(1))?.type;
+    } else {
+      return element?.type;
+    }
   }
 }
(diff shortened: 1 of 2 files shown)


src/optimizer/utils.ts
```diff
@@ -164,6 +164,10 @@ function getTypesForPath(
     const instanceOfSD = fhirtypes.StructureDefinition.fromJSON(instanceOfDef);
     // NOTE: Normalize the path to remove indices and/or slice references
     const element = instanceOfSD.findElementByPath(path.replace(/\[[^\]]+\]/g, ''), fisher);
-    return element?.type;
+    if (element?.contentReference) {
+      return instanceOfSD.findElement(element.contentReference.slice(1))?.type;
```
cmoesel, line 168: Since you are using `.slice(1)`, I think this assumes that the `contentReference` starts with a `#`.  But... I _think_ it is possible for a `contentReference` to start with a type id or URL (and _then_ `#element`).

Just to be safe, we should either:
* also test that the `contentReference` starts with `#`, OR
* do `.slice(element.contentReference.indexOf('#') + 1)`

I _think_ the 2nd one is safe, as I don't think contentReferences cross resources -- so the element id should exist, even if it references a different profile.  I'm not 100% sure, but I think it's a risk I'm willing to take.  That said, I'm fine w/ either of the above, so you can decide.

cmoesel (changes requested):
This seems to fix the problem. Thanks! I've added only one small suggestion.

I can't remember if we talked about this, but is there a reason we check for CODE_IMPOSTER_PATHS when processing instance paths, but not caret paths? Should we update the code so it uses the CODE_IMPOSTER_PATHS in both cases?

jafeltra (commented):
This looks good to me! I didn't notice anything besides what Chris had pointed out, so once those are addressed, I'm good to approve.

cmoesel (approved):
Looks good! Thank you!

PR #209, opened 2023-02-03
Update PathRule optimization

### LINKED ISSUE #208: Indenting not working correctly
I just did a quick test on US Core 5.0.1 and got some weird results when indenting... For example, look at this example instance it produced:
```
Instance: 10-minute-apgar-color
InstanceOf: USCoreObservationSurveyProfile
Title: "10 minute Apgar Color Example"
Description: "This example of a US Core Survey Observation Profile illustrates its use to directly capture individual surveys assessment items as an observations. It is not derived from a FHIR QuestionnaireResponse."
Usage: #example
* meta.extension[0]
* meta.extension[=].url = "http://hl7.org/fhir/StructureDefinition/instance-name"
* meta
  * extension[=].valueString = "10 minute Apgar Color Example"
  * extension[+]
    * url = "http://hl7.org/fhir/StructureDefinition/instance-description"
    * valueMarkdown = "This example of a US Core Survey Observation Profile illustrates its use to directly capture individual surveys assessment items as an observations. It is not derived from a FHIR QuestionnaireResponse."
* status = #final
* category = $observation-category#survey "Survey"
  * text = "Survey"
* code = $loinc#32401-2 "10 minute Apgar Color"
  * text = "Apgar color score"
* subject = Reference(infant-example)
* effectiveDateTime = "2016-05-18T22:33:22+00:00"
* performer = Reference(practitioner-1) "Ronald Bone"
* valueCodeableConcept.coding.extension
  * url = "http://hl7.org/fhir/StructureDefinition/ordinalValue"
* valueCodeableConcept.coding
  * extension.valueDecimal = 2
* valueCodeableConcept
  * coding = $loinc#LA6724-4 "Good color all over"
  * text = "2. Good color all over"
```

Note the weirdness with the `meta` paths and `valueCodeableConcept` paths.  Something odd is definitely happening. It seems there might be a correlation between weirdness and the presence of extensions in the path -- so perhaps there is some optimizer that is interacting poorly with it.

jafeltra:
Fixes #208.

This PR updates how path rules are calculated in order to better determine when they are needed. This fixes the issue reported in #208. The problem in the reported issue was due to a path with multiple parts that didn't exactly match the previous rule, but did partially match the rule, so a path rule for the first part of the path was added incorrectly.

The `AddPathRuleOptimizer` does a bit more than just add path rules now due to the new implementation. It adds path rules as necessary, but it also reorders rules so that if an Assignment Rule or other type of rule can be used to set context (when we simplify the paths later), that rule is used first and no path rule is needed. It might be worth updating the name of the optimizer, but I wasn't sure so I haven't done that yet.

The refactor not only fixes the problem in the #208, but it also fixes some other issues that seem to have been introduced recently. I tested this branch by running a round trip on US Core. The comparison generated has some differences, so I also compared the `fsh-generated` folder from a round trip on this branch with a round trip on `master` and GoFSH `v1.6.3` (the latest published version). There are a few issues in the round trip on `master` that are resolved in this branch. If anyone reviewing this would like more details on what I compared, let me know.

### DIFF
--- a/src/optimizer/plugins/AddPathRulesOptimizer.ts
+++ b/src/optimizer/plugins/AddPathRulesOptimizer.ts
@@ -1,10 +1,91 @@
 import { Package } from '../../processor';
 import { OptimizerPlugin } from '../OptimizerPlugin';
 import { ProcessingOptions } from '../../utils';
-import { ExportablePathRule } from '../../exportable';
+import {
+  ExportableAddElementRule,
+  ExportableMappingRule,
+  ExportablePathRule,
+  ExportableSdRule
+} from '../../exportable';
 import ConstructInlineInstanceOptimizer from './ConstructInlineInstanceOptimizer';
 import SimplifyArrayIndexingOptimizer from './SimplifyArrayIndexingOptimizer';
 import SimplifyRulePathContextsOptimizer from './SimplifyRulePathContextsOptimizer';
+
+// NOTE: These are the rule types that are allowed on the entities that are optimized
+// by this optimizer. Type is created to allow specificity and convenience.
+type AllowedRule =
+  | ExportableMappingRule
+  | ExportablePathRule
+  | ExportableAddElementRule
+  | ExportableSdRule;
+
+type PathNode = {
+  path: string;
+  rule?: AllowedRule;
+  children: PathNode[];
+};
+
+// Create a tree with a node for each path part for all rules
+function buildPathTree(parent: PathNode, rule: AllowedRule) {
+  // We want to process each part, so split at each . not within []
+  // Because we want to consider each slice and index within the path as a
+  // unique path, we don't want to use the full parseFSHPath.
+  const pathParts = rule.path.split(/\.(?![^\[]*\])/g);
+  let currentPath = '';
+  let lastParent: PathNode;
+  pathParts.forEach(part => {
+    currentPath = currentPath === '' ? part : `${currentPath}.${part}`; // build the path back up
+    lastParent = parent.children.find(child => child.path === currentPath);
+    if (lastParent == null) {
+      // If this path hasn't been seen, create a node for it in the tree
+      const newNode: PathNode = { path: currentPath, children: [] };
+      // If the current path is also the rule path, we want to include the rule on the node
+      if (currentPath === rule.path) {
+        newNode.rule = rule;
+      }
+      parent.children.push(newNode);
+
+      // Use the current path's node as the parent for the next path part
+      parent = newNode;
+    } else {
+      // If the current path is also the rule path but that path's node was previously added to the tree,
+      // we still want to add the rule to node so the rule is tracked
+      if (currentPath === rule.path) {
+        lastParent.rule = rule;
+      }
+
+      // Use the current path's node as the parent for the next path part
+      parent = lastParent;
+    }
+  });
+}
+
+// Build the list of rules that will create optimal context.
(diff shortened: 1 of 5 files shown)


test/optimizer/plugins/AddPathRulesOptimizer.test.ts
```diff
@@ -263,5 +372,251 @@ describe('optimizer', () => {
...
+
+    it('should optimize order for rules that require multiple rules at the same path', () => {
+      // Profile: MultipleRulesPerPathObs
+      // Parent: Observation
+      // * category 1..1 MS
+      // * category only CodeableConcept
+      // * category.coding 1..* MS
+      // * category.coding only Coding
+      // * category.coding.system 1..1 MS
+      // * category.coding.system only uri
```
mint-thompson, line 539: In this test, you put the `only` rules at the end, so I think this FSH should reflect that.

src/optimizer/plugins/AddPathRulesOptimizer.ts
```diff
@@ -1,11 +1,92 @@
...
+      // If the current path is also the rule path but that path's node was previously added to the tree,
+      // we still want to add the rule to node so the rule is tracked
+      if (currentPath === rule.path) {
+        currentPathNode.rule = (currentPathNode.rule ?? []).concat(rule);
+      }
+
+      // Use the current path's node as the parent for the next path part
+      parent = currentPathNode;
+    }
+  });
```
mint-thompson, line 60: I'm thinking a bit about the implementation inside this `forEach`. There are two main things I'm thinking about:
- The main difference between the two paths from your `if` statement is whether you have to create a new node. Apart from that, they do the same thing.
- The condition for adding `rule` to a node is true on the last iteration of the loop. Put another way, it could be done after the loop finishes, rather than having a check on every iteration.
I might be trying to get too clever here, though. Let me know if you want to talk about this.

mint-thompson (changes requested):
This looks basically excellent. I have only minor questions about the implementation. But, it sounds like this helps with a lot of issues that came up when running against USCore.

test/optimizer/plugins/AddPathRulesOptimizer.test.ts
jafeltra: Ah good catch. I had it in the expected order. I've updated the comment, but I realized I didn't have a comment above the expect statements that has the expected FSH. Do you think this would be helpful to include? I was thinking the expect statements were enough, but I realized most other tests in this file do have the second commented FSH block.

cmoesel (dismissed):
This approach worked really well! Nicely executed! I also appreciate the slew of tests.

I am approving this as-is, but have also noted a potential simplification of one function. If you like the simplification, feel free to incorporate. Otherwise, feel free to keep it as-is!

src/optimizer/plugins/AddPathRulesOptimizer.ts
```diff
@@ -1,11 +1,82 @@
...
+      if (node.rules.length > 0) {
+        rules.push(...node.rules);
+      }
+      // If this path has a child, continue processing the child node.
+      if (node.children.length > 0) {
+        createAndOrganizeRules(node.children, rules);
+      }
+    }
+  });
+}
```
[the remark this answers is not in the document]
jafeltra: Ohh much simpler! Thanks! I like it, I'll add it.

mint-thompson (approved):
Excellent work, thank you!

vadi2:
This is a great improvement - giant questionnaires are now readable. Thanks @guhanthuran @jafeltra!

PR #213, opened 2023-03-27
Create inline instances with numeric ids

### LINKED ISSUE #203: GoFSH should support creating inline instances that have ids that are numbers
_*Note:* This issue is currently blocked by [this bug](https://github.com/FHIR/sushi/issues/1184) in SUSHI. If SUSHI cannot assign instances with an id that is a number, it is good that GoFSH does not create instances with ids that are numbers. We should not fix this bug before the SUSHI bug is fixed._

Currently, when GoFSH tries to create an inline instance, it checks if the Instance's `id` is a number. If it is a number, it does not set the `Instance` keyword to that numeric id and instead auto-generates a string to be used with the `Instance` keyword. It then creates an assignment rule that sets the id to the numeric value. This allows SUSHI to use the string that is used with the `Instance` keyword when assigning the inline instance, but the actual instance JSON preserves the correct `id`.

For example, when the following JSON is converted to FSH, the `id` is set via an assignment rule and an auto-generated string is used with the `Instance` keyword.

```
{
  "resourceType": "Condition",
  "id": "ExampleCondition",
  "contained": [
    {
      "resourceType": "Patient",
      "id": "123",
      "name": [
        {
          "given": [
            "Jane"
          ],
          "family": "Doe"
        }
      ]
    }
  ],
  "subject": {
    "reference": "#123"
  }
}
```

This behavior is likely implemented because SUSHI cannot currently assign instances that have a numeric id. Even though FHIR allows an `id` to just be a number, SUSHI does not properly recognize the number as an `id` of a resource and instead just thinks it is a number. This is a separate issue [here](https://github.com/FHIR/sushi/issues/1184). Once that issue is fixed, GoFSH does not need to prevent numeric `id`s from being set with the `Instance` keyword.

mint-thompson:
Fixes #203 and completes task [CIMPL-1053](https://standardhealthrecord.atlassian.net/browse/CIMPL-1053).

The fix is pretty straightforward: remove the condition that checks to make sure `id` isn't numeric. Sometimes, changes are not so tricky. 

IMPORTANT NOTE: Do not merge this in until there has been an official SUSHI release that supports parsing Instances with numeric IDs!

### DIFF
--- a/test/optimizer/plugins/ConstructInlineInstanceOptimizer.test.ts
+++ b/test/optimizer/plugins/ConstructInlineInstanceOptimizer.test.ts
@@ -225,12 +225,10 @@ describe('optimizer', () => {
         optimizer.optimize(myPackage, fisher);
 
         expect(myPackage.instances).toHaveLength(3);
-        const expectedRule1 = cloneDeep(containedId1);
-        expectedRule1.path = 'id';
         assertExportableInstanceWithDifferentName(
           myPackage.instances[1],
-          'Inline-Instance-for-Foo-1',
-          '123',
+          '0123',
+          '0123',
           'Observation',
           'Inline',
           undefined,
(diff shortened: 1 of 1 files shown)


jafeltra (dismissed):
Oh. Nice. Sometimes changes aren't so tricky.

cmoesel (approved):
This is good! But I'm also converting to draft so we don't accidentally merge before SUSHI is released.

image

PR #215, opened 2023-04-12
Use name of local CodeSystem when optimizing assigned code

### LINKED ISSUE #190: --no-aliases:  doesn't handle local code
References to codesystems that are defined LOCAL to an IG still generate $<name> references.

mint-thompson:
Fixes #190 and completes task [CIMPL-989](https://standardhealthrecord.atlassian.net/browse/CIMPL-989).

A url for the system of an assigned code used in an AssignmentRule or CaretValueRule will be turned into the name of a locally defined CodeSystem when possible. A local name can be used regardless of the option to create aliases.

The existing `optimizeURL` function is used here, since it does exactly what we need.

### DIFF
--- a/src/optimizer/plugins/ResolveValueRuleURLsOptimizer.ts
+++ b/src/optimizer/plugins/ResolveValueRuleURLsOptimizer.ts
@@ -1,4 +1,4 @@
-import { fshtypes } from 'fsh-sushi';
+import { fshtypes, utils } from 'fsh-sushi';
 import { OptimizerPlugin } from '../OptimizerPlugin';
 import { Package } from '../../processor';
 import {
@@ -6,16 +6,16 @@ import {
   ExportableCaretValueRule,
   ExportableRule
 } from '../../exportable';
-import { resolveAliasFromURL } from '../utils';
+import { optimizeURL } from '../utils';
 import CombineCodingAndQuantityValuesOptimizer from './CombineCodingAndQuantityValuesOptimizer';
-import { ProcessingOptions } from '../../utils';
+import { ProcessingOptions, MasterFisher } from '../../utils';
 
 export default {
   name: 'resolve_value_rule_urls',
   description: 'Replace URLs in caret and assignment rules with their names or aliases',
   runAfter: [CombineCodingAndQuantityValuesOptimizer.name],
 
-  optimize(pkg: Package): void {
+  optimize(pkg: Package, fisher: MasterFisher, options: ProcessingOptions = {}): void {
     [
       ...pkg.instances,
       ...pkg.profiles,
@@ -28,13 +28,15 @@ export default {
           (rule instanceof ExportableAssignmentRule || rule instanceof ExportableCaretValueRule) &&
           rule.value instanceof fshtypes.FshCode
         ) {
-          rule.value.system =
-            resolveAliasFromURL(rule.value.system, pkg.aliases) ?? rule.value.system;
+          rule.value.system = optimizeURL(
+            rule.value.system,
+            pkg.aliases,
+            [utils.Type.CodeSystem],
+            fisher,
+            options.alias ?? true
+          );
         }
       });
     });
-  },
-  isEnabled(options: ProcessingOptions): boolean {
-    return options.alias === true;
   }
 } as OptimizerPlugin;
--- a/src/optimizer/plugins/ResolveValueRuleURLsOptimizer.ts
+++ b/src/optimizer/plugins/ResolveValueRuleURLsOptimizer.ts
@@ -11,9 +11,9 @@ import CombineCodingAndQuantityValuesOptimizer from './CombineCodingAndQuantityV
 import { ProcessingOptions, MasterFisher } from '../../utils';
 
 export default {
-  name: 'resolve_value_rule_system_system_urls_for_codes',
+  name: 'resolve_value_rule_system_urls_for_codes',
   description:
-    'Replace URLs in code values in caret and assignment rules with their names or aliases',
+    'Replace system URLs in code values in caret and assignment rules with their names or aliases',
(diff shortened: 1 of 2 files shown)


src/optimizer/plugins/ResolveValueRuleURLsOptimizer.ts
```diff
@@ -1,21 +1,21 @@
...
 } from '../../exportable';
-import { resolveAliasFromURL } from '../utils';
+import { optimizeURL } from '../utils';
 import CombineCodingAndQuantityValuesOptimizer from './CombineCodingAndQuantityValuesOptimizer';
-import { ProcessingOptions } from '../../utils';
+import { ProcessingOptions, MasterFisher } from '../../utils';
 
 export default {
   name: 'resolve_value_rule_urls',
   description: 'Replace URLs in caret and assignment rules with their names or aliases',
```
cmoesel, line 15: This isn't the fault of this PR, but... this optimizer seems to be poorly named / described, because it's much narrower than it suggests. It doesn't resolve all value rule URLs; it only resolves the system URLs of codes in value rules. It may be more accurate to call it `resolve_value_rule_system_system_urls_for_codes` (I almost said `resolve_value_rule_code_system_urls`, but that can also be misunderstood).

src/optimizer/plugins/ResolveValueRuleURLsOptimizer.ts
```diff
@@ -28,13 +28,15 @@ export default {
...
         ) {
-          rule.value.system =
-            resolveAliasFromURL(rule.value.system, pkg.aliases) ?? rule.value.system;
+          rule.value.system = optimizeURL(
+            rule.value.system,
+            pkg.aliases,
+            [utils.Type.CodeSystem],
+            fisher,
+            options.alias ?? true
+          );
```
cmoesel, line 37: I love that we're re-using this, but it may be too aggressive. The request was to use _local_ code system names instead of aliases -- but this replaces _all_ code system URLs with their names (including those from core FHIR and dependency IGs). As an example, here is a screenshot of a diff of the aliases in US Core between `master` and this branch:

image

We probably want to think about if this is what we really want. I think best practice is generally to use URLs for external systems and names for local systems. That's the direction I'm leaning, but if you disagree then I think it's worth a discussion!

cmoesel (commented):
This one did a bit more than I expected, so I've noted my expectations vs what it does -- and have asked for your thoughts.

src/optimizer/plugins/ResolveValueRuleURLsOptimizer.ts
mint-thompson: I think you're right: this change might be too aggressive in reducing the number of aliases. I'll try to adjust this to use names only for systems in the lake.

src/optimizer/plugins/ResolveValueRuleURLsOptimizer.ts
```diff
@@ -8,14 +8,15 @@ import {
 } from '../../exportable';
 import { resolveAliasFromURL } from '../utils';
 import CombineCodingAndQuantityValuesOptimizer from './CombineCodingAndQuantityValuesOptimizer';
-import { ProcessingOptions } from '../../utils';
+import { ProcessingOptions, MasterFisher } from '../../utils';
 
 export default {
-  name: 'resolve_value_rule_urls',
-  description: 'Replace URLs in caret and assignment rules with their names or aliases',
+  name: 'resolve_value_rule_system_system_urls_for_codes',
```
cmoesel, line 14: Doh. I'm sorry. You followed my instructions too closely! I know I said `resolve_value_rule_system_system_urls_for_codes`, but the double-`system` was a typo. I think it should just be `resolve_value_rule_system_urls_for_codes`

src/optimizer/plugins/ResolveValueRuleURLsOptimizer.ts
```diff
@@ -8,14 +8,15 @@ import {
...
+  description:
+    'Replace URLs in code values in caret and assignment rules with their names or aliases',
```
cmoesel, line 16:
```suggestion
    'Replace system URLs in code values in caret and assignment rules with their names or aliases',
```

cmoesel (changes requested):
So this one is my mistake... I fed you a bad suggested name. Sorry!

cmoesel (changes requested):
I'm not sure what's up, but when I run this against US Core 5.0.1 and compare to the output of running `master` against US Core 5.0.1 -- I'm seeing all the `urn:`-based code systems get dropped:
image

jafeltra:
I reviewed this, and I didn't notice anything else beyond the `urn:` aliases Chris mentioned.

src/optimizer/plugins/ResolveValueRuleURLsOptimizer.ts
```diff
@@ -28,13 +29,17 @@ export default {
...
-          rule.value.system =
-            resolveAliasFromURL(rule.value.system, pkg.aliases) ?? rule.value.system;
+          const localSystem = fisher.lakeOfFHIR.fishForFHIR(
+            rule.value.system,
+            utils.Type.CodeSystem
+          );
+          if (localSystem?.name) {
+            rule.value.system = localSystem.name;
+          } else if (options.alias ?? true) {
+            rule.value.system = resolveAliasFromURL(rule.value.system, pkg.aliases);
```
cmoesel, line 39: I think maybe this needs to be the same as it was before:
```ts
rule.value.system =
            resolveAliasFromURL(rule.value.system, pkg.aliases) ?? rule.value.system;
```

cmoesel (approved):
FANtastic.

image

jafeltra (approved):

PR #219, opened 2023-04-27
Handle numeric element names when importing resources

mint-thompson:
Completes task [CIMPL-1085](https://standardhealthrecord.atlassian.net/browse/CIMPL-1085).

Elements are allowed to have numeric names. When flattening an object, array indices get put into square brackets, but numeric element names do not. This keeps the paths intact when checking types and writing assignment or caret value rules.

One interesting side effect of the new flattening function is that empty arrays are no longer possible as values. As a result, the `isFSHValueEmpty` check now never returns true. So, either we can remove that check, or we need to come up with a case where it will return true.

This PR is staying in a draft state until there is an official SUSHI release that supports numeric paths.


cmoesel (changes requested):
This seems like a good way to solve this problem!

Regarding `isFSHValueEmpty`... I can't think of a way to force an empty value. Maybe there is some exceptional situation where it could happen? If we can't think of anything, I guess we could just go ahead and remove it. It probably isn't doing a lot of harm, but leaving dead code around isn't great either, so... I guess we remove it until and unless someone figures out a way to make it happen!

Although you have a test showing that the `getPathValuePairs` function now treats numeric keys and numeric indices correctly, I think it would be nice to have a test that shows numeric keys actually _work_ in the full context of GoFSH. Could you consider adding an extractor test or two?

Also... I noticed that `.tool-versions` still indicates a Node 12 version. Could you bump that up to the same as what's on SUSHI's `next` branch? That should be `16.19.0`.

mint-thompson:
@cmoesel Thank you for the suggestions! I've made updates accordingly. Additionally, I figured out a case where `isFSHValueEmpty` would still help us out: an array that contains an empty object. I added another test case to demonstrate it on a StructureDefinition, which covers the previously uncovered code.

jafeltra (approved):
This makes sense to me. I ran this against US Core and everything seemed to be as expected. The only thing I noticed was that it looks like this branch is one commit behind `next-v3` since the other PR for numeric instance names was merged. I'm not sure if you want to update it now or let the merge handle it.

PR #221, opened 2023-05-18
Special handling for CodeableReference type constraints

### LINKED ISSUE #220: Cannot roundtrip R5 workflow reason extension
Roundtripping [workflow-reason](http://hl7.org/fhir/extensions/StructureDefinition-workflow-reason.profile.json.html) from R5 doesn't work, both in fshonline and gofsh on the CLI:

```
gofsh --indent --useFHIRVersion 5.0.0
sushi
// ...
info  Converting FSH to FHIR resources...
error Cannot bind value set to Reference; must be coded (code, Coding, CodeableConcept, Quantity, CodeableReference), or the data types (string, uri).
  File: .../gofsh/input/fsh/extensions/WorkflowReason.fsh
  Line: 33
```

The [R4 version](http://hl7.org/fhir/r4/extension-workflow-reasoncode.json.html) of the extension does pass without errors (just a warning).

mint-thompson:
Fixes #220 and completes task [CIMPL-1100](https://standardhealthrecord.atlassian.net/browse/CIMPL-1100).

CodeableReference elements have to be handled a bit carefully because they are sort of like Reference elements, but the FSH Reference keyword usually tries to make things a Reference type. This can lead to issues when a choice element has a CodeableReference type. Make separate rules for constraining the type to CodeableReference and for constraining the target profiles for CodeableReference.

When round-tripping the file in the issue, the only differences are in the ordering of properties. The starting and ending JSON are logically equal.

One thing I noticed that may or may not be in scope to do here is that `OnlyRuleExtractor.process` can't really handle a case where an element with type `Reference` or `CodeableReference` has both `profile` and `targetProfile` defined. As far as I know, nothing in the spec prohibits this, so that will need to be addressed.


test/extractor/AddElementRuleExtractor.test.ts
```diff
@@ -60,14 +71,81 @@ describe('AddElementRuleExtractor', () => {
...
+        element,
+        looseSD
+      );
+      const expectedElement = new ExportableAddElementRule('cheese[x]');
+      expectedElement.min = 0;
+      expectedElement.max = '*';
+      expectedElement.short = 'Cheese is versatile';
+      expectedElement.definition = 'You can put cheese on a lot of things. Sometimes you melt it.';
+      expectedElement.types.push({ type: 'string' }, { type: 'CodeableReference' });
+      const expectedOnly = new ExportableOnlyRule('cheeseCodeableReference');
```
cmoesel, line 123: Oh... I hadn't thought of this scenario, but... this is actually a problem. The only way to apply the reference to the `CodeableReference` is via a type slice path, but... slices aren't allowed in logical models! SUSHI doesn't yell, but the IG Publisher does:
> Rule sdf-1: 'Element paths must be unique unless the structure is a constraint' Failed

So, creating a type slice here is not a viable option. Although we _could_ use a caret rule assignment, that's awful -- and we can't expect our users to do that. We may have to introduce `CodeableReference(A or B)` after all... I'll talk to Mark.

cmoesel (changes requested):
I hadn't thought too much about how this works w/ AddElement rules. In particular, I hadn't fully thought out the need for something like this when `Reference` and `CodeableReference` are involved:
```
* med[x] 0..* Reference(Medication or Immunization) or CodeableReference "Medication or Immunization" "Medication or Immunization"
* medCodeableReference only Reference(Medication or Immunization)
```
Unfortunately, the second rule is not valid in a logical or resource because you cannot introduce slices in a logical or resource (only in profiles and extensions). So we're at an impasse. We'll need to discuss this further.

mint-thompson:
This PR is on hold until there are some changes in SUSHI to support a `CodeableReference` keyword.

mint-thompson:
closed, there will be a new PR that uses the `CodeableReference` keyword.

PR #224, opened 2023-06-08
v3.0.0

cmoesel:
I think that since we changed the Node.js requirements, we need to call this a major version (it's technically a breaking change). If anyone disagrees, however, I'm willing to have that conversation.


mint-thompson (approved):

cmoesel:
Oops. We only should have been on v2.0.0. I removed this erroneous commit and replaced it w/ v2.0.0.

PR #225, opened 2023-07-06
Update how plugin module is loaded to support FSH Online

jafeltra:
This PR updates how the plugin module is loaded in order to work in FSH Online. For some reason, even with the workaround we previously had, FSH Online was unable to load the plugins module after the GoFSH 2.0.0 release. I'm not sure exactly why things changes, but the root of the change seems to be trying to import a module set in a variable, rather than a string. The way the compiled code changed made it so that, even with the old workaround in place, a variable was still be passed into the function that does the import.

This update makes it so that, if no alternative plugin directory is provided, the default `'plugins'` directory is imported via a string, rather than a variable whose value is not changed. Using this default is the behavior GoFSH uses in the `getResources` function, which FSH Online also calls, so FSH Online will only ever use the default behavior now.

This PR should be reviewed in combination with [FSH Online #136](https://github.com/FSHSchool/FSHOnline/pull/136) since its purpose is to resolve an issue there.

### DIFF
--- a/src/optimizer/loadOptimizers.ts
+++ b/src/optimizer/loadOptimizers.ts
@@ -8,21 +8,25 @@ import { logger } from '../utils/GoFSHLogger';
  * If there is a circular dependency in the plugins, an error will be logged and the resulting order is undetermined.
  * @param folder - the folder to load optimizers from.  Currently only used in tests.  CLI uses default value.
  */
-export async function loadOptimizers(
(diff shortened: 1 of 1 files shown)


cmoesel (approved):
This works and seems to be the best/cleanest solution to a problem we don't fully understand. Thanks!

mint-thompson (approved):

PR #227, opened 2023-07-11
To v2.0.1

jafeltra:
This PR preps for the next GoFSH release with the following steps:
- Bumped SUSHI to the latest 3.1.0
- Ran `npm audit fix` to fix what could be fixed
- Bumped the version to 2.0.1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -23,7 +23,7 @@
         "lodash": "^4.17.21",
         "readline-sync": "^1.4.10",
         "sanitize-filename": "^1.6.3",
-        "semver": "^7.3.2",
+        "semver": "^7.5.4",
         "temp": "^0.9.1",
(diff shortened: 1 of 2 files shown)


cmoesel (changes requested):
I noticed that `semver` still causes issues w/ audit. I'd recommend you do the same as I did for SUSHI and use overrides to force it to the latest. I reviewed the change logs and the only notable differences between 6 and 7 was a change in requirements for Node.js -- so it should be safe.

In this case, since we're using `semver` directly, I'd start by updating it directly:
```
npm install semver@latest
```
and then add the override referencing the dependency version:
```
"overrides": {
    "semver": "$semver"
  }
```

cmoesel (approved):
It's good!

PR #228, opened 2023-07-14
Load the same automatic dependencies as SUSHI

mint-thompson:
Completes task [CIMPL-1113](https://standardhealthrecord.atlassian.net/browse/CIMPL-1113).

Use SUSHI's function for loading automatic dependencies. There is a minor discrepancy in the expected type, but the FHIR definitions are only passed in so they can be passed to fhir-package-loader's mergeDependency function. So, the ts-ignore here should be safe. Add a note indicating the reason for the ts-ignore and the conditions under which it can be removed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "diff": "^5.0.0",
     "diff2html": "^3.1.18",
     "fhir": "^4.10.0",
-    "fhir-package-loader": "^0.3.0",
+    "fhir-package-loader": "^0.4.0",
     "flat": "^5.0.2",
     "fs-extra": "^9.0.1",
     "fsh-sushi": "^3.1.0",
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
     "diff": "^5.0.0",
     "diff2html": "^3.1.18",
     "fhir": "^4.10.0",
-    "fhir-package-loader": "^0.4.0",
+    "fhir-package-loader": "^0.5.0",
     "flat": "^5.0.2",
     "fs-extra": "^9.0.1",
     "fsh-sushi": "^3.1.0",
--- a/src/app.ts
+++ b/src/app.ts
@@ -5,22 +5,21 @@ import fs from 'fs-extra';
 import program from 'commander';
 import chalk from 'chalk';
 import { pad, padStart, padEnd } from 'lodash';
-import { fhirtypes, utils } from 'fsh-sushi';
+import { utils } from 'fsh-sushi';
 import {
-  determineCorePackageId,
   ensureOutputDir,
   FHIRDefinitions,
   getInputDir,
   getAliasFile,
   getFhirProcessor,
   getResources,
-  loadExternalDependencies,
   writeFSH,
   logger,
   stats,
   fshingTrip,
   getRandomPun,
-  ProcessingOptions
+  ProcessingOptions,
+  loadExternalDependencies
 } from './utils';
 import { Package, AliasProcessor } from './processor';
 import { ExportableAlias } from './exportable';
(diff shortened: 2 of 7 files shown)


src/app.ts
```diff
@@ -208,6 +208,12 @@ async function app() {
     ) ?? [];
   const fhirPackageId = determineCorePackageId(config.config.fhirVersion[0]);
   allDependencies.push(`${fhirPackageId}@${config.config.fhirVersion[0]}`);
+  await utils.loadAutomaticDependencies(
+    config.config.fhirVersion[0],
+    config.config.dependencies ?? [],
+    // @ts-ignore TODO: this can be removed once SUSHI changes the type signature for this function to use FPL's FHIRDefinitions type
+    defs
+  );
   const dependencyDefs = loadExternalDependencies(defs, allDependencies);
```
jafeltra, line 217: I have a question about the location of this new function. Do you know if it is possible/if it is useful to include the `loadAutomaticDependencies` within the `loadExternalDependencies` function? To me, it seems like it logically makes sense there since they are part of the set of external dependencies that are loaded, and it seems like SUSHI loads the automatic dependencies from within its `loadExternalDependencies` function. That would also allow us to add a test to ensure the automatic dependencies are actually included, since the function wouldn't just be within `app.ts`.

That said, I'm curious what your thoughts are on this.

src/app.ts
```diff
@@ -208,6 +208,12 @@ async function app() {
     ) ?? [];
   const fhirPackageId = determineCorePackageId(config.config.fhirVersion[0]);
   allDependencies.push(`${fhirPackageId}@${config.config.fhirVersion[0]}`);
+  await utils.loadAutomaticDependencies(
+    config.config.fhirVersion[0],
+    config.config.dependencies ?? [],
+    // @ts-ignore TODO: this can be removed once SUSHI changes the type signature for this function to use FPL's FHIRDefinitions type
```
jafeltra, line 214: Did we say there should be a PR for SUSHI to do this as part of this overall task (but no release just for this)?

jafeltra (commented):
I just had two pretty small questions. Let me know what you think.

src/app.ts
mint-thompson: We did! Just created [CIMPL-1143](https://standardhealthrecord.atlassian.net/browse/CIMPL-1143) for this.

jafeltra: Oh nice, I lost track of that task! Thanks!

src/utils/Processing.ts
```diff
@@ -112,7 +113,26 @@ export function writeFSH(resources: Package, outDir: string, style: string): voi
...
+export async function loadExternalDependencies(
+  defs: FHIRDefinitions,
+  config: ExportableConfiguration
+) {
+  const allDependencies =
+    config.config.dependencies?.map(
+      (dep: fhirtypes.ImplementationGuideDependsOn) => `${dep.packageId}@${dep.version}`
+    ) ?? [];
+  const fhirPackageId = determineCorePackageId(config.config.fhirVersion[0]);
+  allDependencies.push(`${fhirPackageId}@${config.config.fhirVersion[0]}`);
```
jafeltra, line 125: I'm actually not sure which way is what we want, but I noticed that in the old `fhirToFsh` function, we only pushed on the FHIR version if it wasn't already included in the dependency list. I think it makes sense to have that logic here, but I also noticed we did _not_ have that logic in `app.ts`. It might be good to avoid trying to load that package twice, but I'm curious what you think.

test/api/FhirToFsh.test.ts
```diff
@@ -111,52 +111,6 @@ describe('fhirToFsh', () => {
...
-      ]
-    });
-
-    expect(loadSpy.mock.calls).toHaveLength(1);
-    expect(loadSpy.mock.calls[0][1]).toEqual([
-      'hl7.fhir.us.core@3.1.0',
-      'hl7.fhir.us.mcode@1.0.0',
-      'hl7.fhir.r4.core@4.0.1'
-    ]);
-  });
```
jafeltra, line 137: I'm wondering if it might still be useful to have this test (or something similar) even though the functionality is now moved out to the `loadExternalDependencies` function. There don't seem to be any other tests that check that the `dependencies` property on the configuration option is used, and it seems like that might be good. Or maybe a simpler test to just check that `loadExternalDependencies` is called from `fhirToFsh` is sufficient?

jafeltra (commented):
This looks really good. I like the refactor, especially since you noticed the `fhirToFsh` piece! I just had two small questions.

src/utils/Processing.ts
mint-thompson: I think it makes sense to not try to load the package twice, so I'll add in that check.

jafeltra (approved):
This looks good to me! Thanks for making those updates!

package.json
```diff
@@ -73,7 +73,7 @@
     "diff": "^5.0.0",
     "diff2html": "^3.1.18",
     "fhir": "^4.10.0",
-    "fhir-package-loader": "^0.3.0",
+    "fhir-package-loader": "^0.4.0",
```
cmoesel, line 76: FPL is at v0.5.0 now, so... since you're bumping this anyway, might as well bump it to the latest.

cmoesel (changes requested):
This looks good. I just have one tiny request...

jafeltra (approved):
I still approve

cmoesel (approved):
Just what the Dr. ordered.

PR #229, opened 2023-07-14
Extract more specific types in OnlyRules with target profiles

### LINKED ISSUE #220: Cannot roundtrip R5 workflow reason extension

mint-thompson:
Fixes #220 and completes task [CIMPL-1100](https://standardhealthrecord.atlassian.net/browse/CIMPL-1100).

Reference, CodeableReference, and canonical types can all be specified with target profiles. When extracting an OnlyRule, set the corresponding boolean value on the extracted type along with the target profiles. This takes advantage of the newly available isCodeableReference flag on SUSHI's OnlyRuleType.

The changes in this PR ended up going a bit beyond the original task description, as I noticed that GoFSH was handling canonical types with target profiles by creating caret rules. But, everything appears to round-trip cleanly, so I thought it would be best to clean up all these types together.

### DIFF
--- a/src/extractor/OnlyRuleExtractor.ts
+++ b/src/extractor/OnlyRuleExtractor.ts
@@ -1,3 +1,4 @@
+import { fshrules } from 'fsh-sushi';
 import { ProcessableElementDefinition } from '../processor';
 import { ExportableOnlyRule } from '../exportable';
 import { getPath } from '../utils';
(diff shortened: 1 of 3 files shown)


cmoesel (approved):
Looks good, works well, so much simpler than when we were avoiding a `CodeableReference` keyword. I like it!

I tested it on US Core, but also made my own small IG that constrained elements of `canonical` and `CodeableReference` -- and it all checks out.

jafeltra (approved):
This makes sense to me. I also tested this on the example provided in the original issue, and the warning is resolved and the JSON round trip seems correct.

PR #230, opened 2023-08-01
Don't warn on AssignmentRules for Logical/Resource

### LINKED ISSUE #226: GoFSH should not warn on fixed values in LMs
GoFSH currently warns on fixed/pattern values in logical models. This is because the FHIR spec technically says they are not allowed ([here](https://hl7.org/fhir/R5/elementdefinition.html#interpretation)). Since implementing this, however, we've been told that it is indeed allowed and asked to relax the behavior in SUSHI to allow for it. We should also relax it in GoFSH so it no longer warns. In addition, we should check if GoFSH warns on constraining inherited elements and also remove those warnings.

mint-thompson:
Fixes #226 and completes task [CIMPL-1139](https://standardhealthrecord.atlassian.net/browse/CIMPL-1139).

Update condition for detecting inherited elements on Logical/Resource definitions so that rules are extracted correctly.

### DIFF
--- a/src/processor/StructureDefinitionProcessor.ts
+++ b/src/processor/StructureDefinitionProcessor.ts
@@ -128,9 +128,12 @@ export class StructureDefinitionProcessor {
         element.sliceName &&
         !/\[x]:[a-z][a-z0-9]*[A-Z][A-Za-z0-9]*$/.test(element.id) &&
         ancestorSliceDefinition == null;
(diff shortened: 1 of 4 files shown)


jafeltra (approved):
These changes make sense to me!

cmoesel (approved):
I did not have time to test this, but I've reviewed the code (twice!) and it looks good to me. Makes sense. Since I have 1 foot out the door and Julia already reviewed and tested, I think 1.5 approvals is good enough!

PR #231, opened 2023-08-16
Don't combine rules on inline ValueSet

### LINKED ISSUE #218: GoFSH should not combine rules on compose.include.concept on Instances of ValueSets
GoFSH correctly does not combine rules for `ValueSet.compose.include.concept`. However, If the ValueSet is contained in another resource, GoFSH will make an inline Instance for it. When it does, the Instance _does_ combine the rules on `compose.include.concept`.

This [FSH Online example](https://fshschool.org/FSHOnline/#/share/3N4rdAJ) demonstrates the problem. If you run GoFSH, the Questionnaire with the contained ValueSet is created correctly. If you run SUSHI on the resulting Questionnaire resource, an inline Instance is created for the ValueSet, and the FSH code syntax is used to combine `compose.include.concept`. Running SUSHI on this FSH results in an error because the types do not match.

mint-thompson:
Fixes #218 and completes task [CIMPL-1095](https://standardhealthrecord.atlassian.net/browse/CIMPL-1095).

Construct inline instances before combining coding and Quantity values. This way, rules that assign to elements in a ValueSet instance that shouldn't be combined will be correctly detected as having a type that does not use combined rules.

### DIFF
--- a/src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
+++ b/src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
@@ -15,6 +15,7 @@ import AddReferenceKeywordOptimizer from './AddReferenceKeywordOptimizer';
 import SimplifyInstanceNameOptimizer from './SimplifyInstanceNameOptimizer';
 import { MasterFisher, logger, ProcessingOptions } from '../../utils';
 import { utils } from 'fsh-sushi';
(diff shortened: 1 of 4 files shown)


jafeltra (approved):
Ohhh this was so simple and nice! I tested it out with the example in the issue and with the test case added, and everything looks good!

PR #232, opened 2023-08-17
update SUSHI to v3.3.1

mint-thompson:
Completes task [CIMPL-1151](https://standardhealthrecord.atlassian.net/browse/CIMPL-1151)

This gets GoFSH up-to-date with SUSHI in preparation for updating both in FSH Online. Note that this does _not_ update the `antlr4` dependency which was updated in SUSHI. GoFSH's usage of `antlr4` is sufficiently limited that it makes sense to wait until the performance improvement in `antlr4` is available, at which point the dependency will be updated in both SUSHI and GoFSH.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "fhir-package-loader": "^0.5.0",
     "flat": "^5.0.2",
     "fs-extra": "^9.0.1",
-    "fsh-sushi": "^3.1.0",
+    "fsh-sushi": "^3.3.1",
     "ini": "^1.3.8",
     "lodash": "^4.17.21",
     "readline-sync": "^1.4.10",
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "fhir-package-loader": "^0.5.0",
         "flat": "^5.0.2",
         "fs-extra": "^9.0.1",
-        "fsh-sushi": "^3.1.0",
+        "fsh-sushi": "^3.3.1",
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -67,7 +67,7 @@
     "typescript": "^4.7.4"
   },
   "dependencies": {
-    "antlr4": "~4.8.0",
+    "antlr4": "~4.13.0",
```
jafeltra, line 70: Why do we want to include an antlr update here? I thought we wanted to continue to avoid updating to the version with performance issues since we did not yet need any other functionality from the updates. Did something change?

jafeltra: I did try running the current version of GoFSH and this version against US Core and they both took about the same time, so maybe it doesn't make much of a difference. But I'm just curious if there was anything in particular that convinced you to update antlr4 to 4.13.0 now.

jafeltra (commented):
I just had a question about updating the antlr version, but everything works as expected.

package.json
mint-thompson: I feel like it's simpler to use the same version of `antlr4` in both tools, but you're right that we don't really need any new features.

jafeltra (approved):
Looks good to me! I tested it out with US Core and don't see any changes and the speed is comparable. I also looked in node_modules and the top level `antlr4` for GoFSH has the old version and no workaround is needed and the `fsh-sushi/antlr4` has the workaround in place.

PR #240, opened 2023-09-06
better performance on SimplifyArrayIndexingOptimizer

mrcerf:
makes a huge performance improvement on large codesystems


cmoesel (approved):
This change makes a lot of sense and I can see why it is more efficient. I'm not sure why we used a `filter` here in the first place, but maybe we used to do more with `filteredElements` later? That said, the solution you've implemented makes perfect sense now! Thank you for this!

mint-thompson (approved):

PR #241, opened 2023-09-07
improve perofrmance of combine_coding_and_quantity_values

mrcerf:
should be faster on large resources

### DIFF
--- a/src/optimizer/plugins/CombineCodingAndQuantityValuesOptimizer.ts
+++ b/src/optimizer/plugins/CombineCodingAndQuantityValuesOptimizer.ts
@@ -53,8 +53,9 @@ export default {
       const rules: fshrules.Rule[] = def.rules; // <-- this assignment makes TypeScript happier in the next chunk of code
       const typeCache: Map<string, fhirtypes.ElementDefinitionType[]> = new Map();
 
-      const ruleMap: {[path: string]: {[caretPath: string]: ExportableCaretValueRule[]}} = {};
-      rules.filter(r => r instanceof ExportableCaretValueRule)
+      const ruleMap: { [path: string]: { [caretPath: string]: ExportableCaretValueRule[] } } = {};
+      rules
+        .filter(r => r instanceof ExportableCaretValueRule)
         .forEach((r: ExportableCaretValueRule) => {
           ruleMap[r.path] = ruleMap[r.path] || {};
           ruleMap[r.path][r.caretPath] = ruleMap[r.path][r.caretPath] || [];
@@ -85,7 +86,8 @@ export default {
             `${basePath}.display`,
             `${basePath}.unit`,
             `${basePath}.value`
-          ].filter(sibling => ruleMap[rule.path] && ruleMap[rule.path][sibling] && ruleMap[rule.path][sibling].length)
+          ]
+            .filter(sibling => ruleMap[rule.path] && ruleMap[rule.path][sibling]?.length)
             .map(sibling => ruleMap[rule.path][sibling][0]);
           if (siblings.length) {
             const systemSibling = siblings.find(sibling => sibling.caretPath.endsWith('.system'));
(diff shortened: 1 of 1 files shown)


mint-thompson (approved):
This change looks fine to me. Thank you for submitting this PR!

src/optimizer/plugins/CombineCodingAndQuantityValuesOptimizer.ts
```diff
@@ -53,6 +53,15 @@ export default {
       const rules: fshrules.Rule[] = def.rules; // <-- this assignment makes TypeScript happier in the next chunk of code
       const typeCache: Map<string, fhirtypes.ElementDefinitionType[]> = new Map();
 
+      const ruleMap: { [path: string]: { [caretPath: string]: ExportableCaretValueRule[] } } = {};
+      rules
+        .filter(r => r instanceof ExportableCaretValueRule)
+        .forEach((r: ExportableCaretValueRule) => {
+          ruleMap[r.path] = ruleMap[r.path] || {};
```
cmoesel, line 60: I think that in TypeScript, there is a preference for using nullish-coalescing `??` over `||` for this type of assignment. This is mainly to avoid overwriting "falsey" non-null values. This probably doesn't apply here, but since we're in TypeScript anyway, I think it's a good practice to use `??` anyway.

src/optimizer/plugins/CombineCodingAndQuantityValuesOptimizer.ts
```diff
@@ -53,6 +53,15 @@ export default {
...
+          ruleMap[r.path][r.caretPath] = ruleMap[r.path][r.caretPath] || [];
```
cmoesel, line 61: Same comment re: `||` and `??`.

src/optimizer/plugins/CombineCodingAndQuantityValuesOptimizer.ts
```diff
@@ -53,6 +53,15 @@ export default {
...
+          ruleMap[r.path][r.caretPath].push(r);
```
cmoesel, line 62: Actually... is there a reason you are storing this as an array? I see below that you only ever access `[0]` -- and because of the way GoFSH works, I don't think you'd ever have any more than one anyway. So couldn't you replace 
```ts
ruleMap[r.path][r.caretPath] = ruleMap[r.path][r.caretPath] || [];
ruleMap[r.path][r.caretPath].push(r);
```
with the more simple (and slightly more efficient):
```ts
ruleMap[r.path][r.caretPath] = r;
```

src/optimizer/plugins/CombineCodingAndQuantityValuesOptimizer.ts
```diff
@@ -72,18 +81,14 @@ export default {
...
             `${basePath}.value`
-          ];
-          const siblings = rules.filter(
-            otherRule =>
-              otherRule instanceof ExportableCaretValueRule &&
-              rule.path === otherRule.path &&
-              siblingPaths.includes(otherRule.caretPath)
-          ) as ExportableCaretValueRule[];
+          ]
+            .filter(sibling => ruleMap[rule.path] && ruleMap[rule.path][sibling]?.length)
```
cmoesel, line 90: I think that
```
ruleMap[rule.path] && ruleMap[rule.path][sibling]?.length
```
can be further simplified to:
```
ruleMap[rule.path]?.[sibling]?.length
```
Of course if you change the value in the sibling map to be a rule instead of an array, you can also remove the `.length` at the end.

cmoesel (changes requested):
Thank you for this contribution! I think it's a very smart approach. I like it!

I made a few minor suggestions for your consideration.

mint-thompson (approved):

cmoesel (approved):
Looks good! Thanks for making GoFSH better!

cmoesel:
GitHub Actions seem to be stuck, but I ran the tests and lint locally -- and it's all good -- so I will force the merge.

Issue #234, opened 2023-08-29
Support FSH 3.0 Context keyword


cmoesel:
GoFSH should properly translate JSON extension contexts to use the [Context](https://hl7.org/fhir/uv/shorthand/2023Sep/reference.html#defining-extensions) keyword.

PR #244, opened 2023-09-28
Extract extension contexts to keyword

mint-thompson:
Completes #234 and task [CIMPL-1163](https://standardhealthrecord.atlassian.net/browse/CIMPL-1163).

When processing an Extension definition, set the contexts directly on the ExportableExtension object. When doing so, convert paths back to FSH-style paths. When extracting caret value rules on an Extension definition, do not extract any rules for contexts.

Add optimizer plugin to resolve URLs in contexts.

This ended up being a bit simpler than I had anticipated in regards to differentiating between core and non-core resources. Since SUSHI does the hard work of actually resolving contexts, GoFSH gets to use the name of anything that it can fish up, no matter where it came from.

### DIFF
--- a/src/optimizer/plugins/RemoveDefaultExtensionContextRulesOptimizer.ts
+++ b/src/optimizer/plugins/RemoveDefaultExtensionContextRulesOptimizer.ts
@@ -1,37 +1,22 @@
 import { OptimizerPlugin } from '../OptimizerPlugin';
 import { Package } from '../../processor';
-import { ExportableCaretValueRule } from '../../exportable';
-import { isEqual, pullAt } from 'lodash';
-import { fshtypes } from 'fsh-sushi';
-const { FshCode } = fshtypes;
+import { isEqual } from 'lodash';
 
 export default {
   name: 'remove_default_extension_context_rules',
   description: 'Remove extension contexts matching the default context that SUSHI generates',
 
   optimize(pkg: Package): void {
-    // * ^context[0].type = #element
-    const DEFAULT_TYPE = new ExportableCaretValueRule('');
-    DEFAULT_TYPE.caretPath = 'context[0].type';
-    DEFAULT_TYPE.value = new FshCode('element');
-    // * ^context[0].expression = "Element"
-    const DEFAULT_EXPRESSION = new ExportableCaretValueRule('');
-    DEFAULT_EXPRESSION.caretPath = 'context[0].expression';
-    DEFAULT_EXPRESSION.value = 'Element';
     // Loop through extensions looking for the default context type (and removing it)
     pkg.extensions.forEach(sd => {
-      const numContexts = sd.rules.filter(
-        r =>
-          r instanceof ExportableCaretValueRule &&
-          r.path === '' &&
-          /^context\[\d+]\.type$/.test(r.caretPath)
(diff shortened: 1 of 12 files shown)


test/optimizer/plugins/ResolveContextURLsOptimizer.test.ts
```diff
@@ -0,0 +1,195 @@
...
+      ]);
+      expect(myPackage.aliases).toEqual([
+        {
+          alias: '$UnknownObservation',
+          url: 'https://demo.org/StructureDefinition/UnknownObservation'
+        }
+      ]);
+    });
+
+    it('should not change an unquoted context when the context value is a URL of an unknown resource with a path and the alias option is false', () => {
```
jafeltra, line 136: Super minor, but I think the description of this test (unknown resource with a path) actually goes with the test below it and that test's description (unknown resource (no path)) goes here.

jafeltra (commented):
The code here looks really good. I left one minor comment about test descriptions inline.

I ran this branch on US Core, which led me to two questions:
1. This came up when I compared the round trip results from US Core and I saw differences that I wasn't expecting. However, this might actually be an issue with SUSHI, and not this branch. It came up in `USCoreEthnicityExtension`, `USCoreRaceExtension`, and `USCoreTribalAffiliationExtension`. They all have a context like this:
```json
{
  "type": "element",
  "expression": "FamilyMemberHistory"
}
```

This branch creates a `Context: FamilyMemberHistory` in the FSH, but then when SUSHI outputs that context into JSON, it creates a context like this:

```json
{
  "expression": "http://hl7.org/fhir/StructureDefinition/DiagnosticReport-geneticsFamilyMemberHistory",
  "type": "extension"
}
```
Assuming the `Context` keyword looks correct, maybe this is just a SUSHI issue.

2. I wanted to double check that this was expected since there didn't seem to be a caret value rule created for the context before this branch, but `USCoreJurisdictionExtension` has a context array this like this:

```json
"context": [
  {
    "type": "element",
    "expression": "Element"
  }
]
```
The released version of GoFSH didn't create any caret rules for this context, but this branch does create `Context: Element`. I think that Context keyword is correct, but the missing caret value rule made me not sure, so I wanted to check.

mint-thompson:
@jafeltra Thank you for the review!
1. That change in USCore is a SUSHI bug, similar to the one recently found for Instances of `Identifier`: if you look at `StructureDefinitionExporter.setContext`, you'll see that it fishes for extensions before any other type.
2. It looks like that's the default context that SUSHI sets when no Context keyword is used. There's a corresponding `RemoveDefaultExtensionContextRulesOptimizer` in GoFSH. This optimizer probably doesn't do much of anything anymore. So, the optimizer should change to instead check the context on each extension, and if the only context is the default, remove it so that no `Context` keyword is used.

jafeltra (approved):
Looks good to me! Thanks for updating that optimizer and answering my questions!

src/optimizer/plugins/RemoveDefaultExtensionContextRulesOptimizer.ts
cmoesel, line 1: Ooooh. This optimizer is _way_ nicer now!

PR #247, opened 2023-10-18
Track code caret rules when simplifying array indexing

### LINKED ISSUE #245: French ICD-10 CodeSystem not usable with Sushi after goFSH
Hello,

I used goFSH from a JSON CodeSystem resource, but the result is not usable with Sushi.

I am attaching the resources that will allow you to reproduce the bug.

I use goFSH 2.1.2, sushi 3.4.0 and nodejs 20.8.0 or 18.15.0.

[CodeSystem-terminologie-cim-10.zip](https://github.com/FHIR/GoFSH/files/12804917/CodeSystem-terminologie-cim-10.zip)

Best regards,

mint-thompson:
Fixes #245 and completes task [CIMPL-1190](https://standardhealthrecord.atlassian.net/browse/CIMPL-1190).

A code caret rule will have an empty path and a non-empty path array. Track them using the path array so that caret paths receive the correct soft indices.

Additionally, update the optimizer to use the now-correct SUSHI functions for path operations.

### DIFF
--- a/src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
+++ b/src/optimizer/plugins/SimplifyArrayIndexingOptimizer.ts
@@ -26,13 +26,20 @@ export default {
       const caretPathMap: Map<string, Map<string, number>> = new Map();
       const ruleArr: fshrules.Rule[] = def.rules;
       const parsedPaths = ruleArr.map((rule: fshrules.Rule) => {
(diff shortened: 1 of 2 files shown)


cmoesel (approved):
Good find! And thanks for deleting all that code we don't need anymore!

jafeltra (approved):
This makes sense to me! Good call on deleting those old functions. I tested the round trip and SUSHI is much happier now.

PR #248, opened 2023-10-20
Simplify code caret path handling in SimplifyArrayIndexingOptimizer

cmoesel:
Your PR #247 works is easy to read as it is, so you can choose to take or leave this PR. But... I noticed some repetition of code, so I wanted to see if I could make it DRYer. I think I did, and in the process was also able to revert back the `parsedPath` object to its original form. [Compared to master](https://github.com/FHIR/GoFSH/compare/master...cimpl-1190-code-system-caret-paths-cmm#diff-ffcb2d5a1dad70041f41fbeb4545dc12c8c4212a674d1143b2aab54637b75c03) it introduces fewer changes than the existing PR. I would have put this in a GitHub PR _suggestion_, but since it spanned a bunch of changes, I couldn't.


mint-thompson (approved):
This is absolutely a much better implementation. Thank you for taking the time to revise my code.

PR #250, opened 2023-11-15
Extract characteristics for Logicals

### LINKED ISSUE #238: Support FSH 3.0 Characteristics keyword
GoFSH should support the new [Characteristics](https://hl7.org/fhir/uv/shorthand/2023Sep/reference.html#defining-logical-models) keyword for logical models.

mint-thompson:
Fixes #238 and completes task [CIMPL-1167](https://standardhealthrecord.atlassian.net/browse/CIMPL-1167).

Logical models use special extensions to set their type characteristics. Check for these extensions when extracting keywords. Do not create caret rules for these extensions.


cmoesel (changes requested):
Everything that's here looks good, but if you run it on the CLI against a logical JSON w/ characteristics, you don't get any of the characteristics in the output FSH. It looks like the Logical class's `toFSH()` method needs to be updated in SUSHI. Foiled again by `toFSH()`!

jafeltra (commented):
This looks really good to me. I tested it out with the branch in SUSHI that updates the `toFSH` function and the Characteristics are there and look good.

One small question I had -- does it make sense to include in either or both of the tests that extensions that are not the characteristics extensions are not included in the characteristics keyword and still have regular rules created?

cmoesel (approved):
This looks good! I like Julia's idea to add a non-context extension to the test JSON to prove it still handles those correctly, but I also recognize that we have other tests that handle normal extensions (and they still pass), so I'm willing to accept this without that change.

jafeltra:
> but I also recognize that we have other tests that handle normal extensions (and they still pass)

I'm guessing we have test cases _somewhere_ that handle extensions on Structure Definitions generically (although I haven't pinpointed exactly which test case covers this), but I don't think we have any tests that test if a Logical has non-characteristics extensions. We only have test cases that have extensions that are all characteristics. Maybe this is testing more than we need to still, but wanted to ask specifically about that. Or maybe I'm just missing something and we have already covered that through something else.

jafeltra (approved):
This looks good! Thanks for updating those tests.

cmoesel (approved):
It's good! Thanks!

PR #251, opened 2023-11-22
Extract rules for Invariants. Use keyword for Invariant description, rules for everything else

### LINKED ISSUE #235: Support FSH 3.0 Invariant Rules
GoFSH should support FSH 3.0 [Invariant Rules](https://hl7.org/fhir/uv/shorthand/2023Sep/reference.html#defining-invariants). This is particularly important for supporting extensions, but is also an opportunity to encourage a best practice of using `Invariant` and `Description` keywords, but rules for everything else. E.g.:
```
Invariant:   us-core-6
Description: "Patient.name.given or Patient.name.family or both SHALL be present"
* severity = #error
* expression = "family.exists() or given.exists()"
* xpath = "f:given or f:family"
```

mint-thompson:
Fixes #235 and completes task [CIMPL-1164](https://standardhealthrecord.atlassian.net/browse/CIMPL-1164).

When extracting an Invariant, create assignment rules for properties that are not represented using keywords. Use the same functions as the InstanceProcessor for this. Add the paths for these elements to the returned list of paths so that caret rules will not be created for these elements.

When producing FSH for an Invariant, use the Description keyword for the description element. For all other metadata elements, use assignment rules instead. These rules will always come before any other rules on the Invariant.


cmoesel (commented):
This generally looks good, but I'm wondering if the `metadataToFSH()` code should actually be in SUSHI rather than here.

jafeltra (commented):
This looks really good and works nicely! I only had a similar question to Chris's question, but I didn't notice anything else. I tried it on US Core and the few invariants with additional rules on them are now extract to the invariant!

cmoesel (approved):
This looks good -- and running it against US Core gets the expected results on invariants.

jafeltra (approved):
Looks good!

PR #254, opened 2024-02-28
Create Aliases with supported characters (and warn on provided unsupported aliases)

jafeltra:
This PR updates that aliases that GoFSH creates to ensure that only supported characters are included in the Alias name. To do this, I replaced any unsupported character with `-`. The PR also makes it so that when provided Aliases use unsupported characters, GoFSH logs the same warning that SUSHI now does.

I couldn't figure out a more elegant way to ignore the leading `$` in alias names with the regex, so if I missed something obvious, I'm happy to make the regex ignore that character so the full replacement can be done with `alias.replace(...)`

### DIFF
--- a/src/optimizer/utils.ts
+++ b/src/optimizer/utils.ts
@@ -11,6 +11,8 @@ import {
   ExportableAlias
 } from '../exportable';
 
+const unsupportedAliasCharRegex = /[^a-zA-z0-9_\-\.]/g;
+
 export function optimizeURL(
   url: string,
   aliases: ExportableAlias[],
@@ -101,6 +103,9 @@ export function resolveAliasFromURL(url: string, aliases: ExportableAlias[]): st
       counterPart += 1;
       alias = `${aliasPart}_${counterPart}`;
     }
+
+    // Replace unsupported characters, but allow $ at the start
+    alias = `$${alias.substring(1).replace(unsupportedAliasCharRegex, '-')}`;
 
     aliases.push(new ExportableAlias(alias, url));
     return alias;
--- a/src/processor/AliasProcessor.ts
+++ b/src/processor/AliasProcessor.ts
@@ -8,6 +8,8 @@ import { InputStream, CommonTokenStream } from 'antlr4';
 import FSHLexer from 'fsh-sushi/dist/import/generated/FSHLexer';
 import FSHParser from 'fsh-sushi/dist/import/generated/FSHParser';
 
+const aliasRegex = /^\$?[a-zA-z0-9_\-\.]*$/;
+
 export class AliasProcessor {
   static process(aliasFile: string): ExportableAlias[] {
     // Load aliases from alias-file option.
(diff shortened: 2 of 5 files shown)


mint-thompson (approved):
I think the implementation you went with for replacing unsupported characters is reasonable. The only other way I can think of implementing this is with a [replacement function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replace#specifying_a_function_as_the_replacement) that checks the match offset, but I don't like that as much as your implementation.

src/optimizer/utils.ts
```diff
@@ -102,6 +104,9 @@ export function resolveAliasFromURL(url: string, aliases: ExportableAlias[]): st
       alias = `${aliasPart}_${counterPart}`;
     }
 
+    // Replace unsupported characters, but allow $ at the start
+    alias = `$${alias.substring(1).replace(unsupportedAliasCharRegex, '-')}`;
```
cmoesel, line 108: Wouldn't it be better to do this on line 98 above? E.g.
```ts
const aliasPart = `$${rawAlias.replace(unsupportedAliasCharRegex, '-')}`;
```
This avoids prepending `$` twice and needing to use `substring(1)`. More importantly, however, I think it works better w/ the existing alias chech on lines 101 - 105. We want to do that check using the final alias (post-replacements), not the pre-replacement alias.

src/processor/AliasProcessor.ts
```diff
@@ -8,6 +8,8 @@ import { InputStream, CommonTokenStream } from 'antlr4';
 import FSHLexer from 'fsh-sushi/dist/import/generated/FSHLexer';
 import FSHParser from 'fsh-sushi/dist/import/generated/FSHParser';
 
+const aliasRegex = /^\$?[a-zA-z0-9_\-\.]*$/;
```
cmoesel, line 11:
```suggestion
const aliasRegex = /^\$?[a-zA-z0-9_\-\.]+$/;
```
I think we want to require at least one character besides the `$`, right?

cmoesel (changes requested):
I like it. It's pretty clean, which is always welcome!

I've made a couple of comments that I think should be addressed before we merge this.

src/optimizer/utils.ts
jafeltra: Oh.. yeah. I'm not sure why I didn't think of that earlier.

src/processor/AliasProcessor.ts
jafeltra: Well... yes that seems reasonable. However, I took this regex from SUSHI. So I think that probably means SUSHI should also require the Alias has at least one other character. So I can update this one, but that probably means I should update SUSHI's as well.

src/optimizer/utils.ts
jafeltra: Added in [9dcef55](https://github.com/FHIR/GoFSH/pull/254/commits/9dcef55564c7708af4d1a0690c721cb62b567329).

src/processor/AliasProcessor.ts
jafeltra: Updated in [b7e0415](https://github.com/FHIR/GoFSH/pull/254/commits/b7e04152c06aa03e829171edbc4f4a46d22708fb). I'll make a small PR in SUSHI (although I should probably run another regression, so that PR is coming eventually).

cmoesel (approved):
Looks good to me! Thank you!

mint-thompson (approved):

PR #255, opened 2024-03-11
Do not create blank alias names + Update ANTLR

jafeltra:
This PR makes a small update to no longer create empty string Alias names. This would happen when a URL that GoFSH tried to make an alias for ended in a `/`, such as `http://example.org/test/url/`. If the URL was just `http://example.org/`, GoFSH already properly handled using the hostname as an alias instead.

I also combined this with updating to the same ANTLR version that SUSHI is using, `4.13.1-patch-1`, because the update didn't require any code changes. I'm happy to separate these out if that is preferred. I ran the tests, and they still work with the new antlr version, and there were no changes when running this branch on US Core.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "2.3.0",
       "license": "Apache-2.0",
       "dependencies": {
-        "antlr4": "~4.8.0",
+        "antlr4": "~4.13.1-patch-1",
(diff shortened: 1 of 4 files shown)


mint-thompson (approved):

cmoesel (approved):
Looks good!

PR #256, opened 2024-03-14
Warn when temporal primitive is not formatted correctly

mint-thompson:
Completes task [CIMPL-1241](https://standardhealthrecord.atlassian.net/browse/CIMPL-1241). See also [FSHOnline #152](https://github.com/FSHSchool/FSHOnline/issues/152).

The FHIR types dateTime, date, time, and instant all have required formats. If a value of one of these types does not match the format, emit a warning. The rule to set this value is still created.

Something I'm thinking about for this is the usefulness of the warning message. Right now, it has the rule path, but no indication of what entity has the incorrectly formatted value. I could send along the entity name to `getFshValue` to use in the message, since by the time the value is returned, the information about the FHIR type is gone. Please let me know what you think about this warning message.

### DIFF
--- a/src/extractor/CaretValueRuleExtractor.ts
+++ b/src/extractor/CaretValueRuleExtractor.ts
@@ -24,10 +24,17 @@ export class CaretValueRuleExtractor {
     const remainingFlatElementArray = flatElementArray.filter(
       ([key]) => !input.processedPaths.includes(key)
     );
+    const entityPath = path === '' ? structDef.name : `${structDef.name}.${path}`;
     remainingFlatElementArray.forEach(([key], i) => {
       const caretValueRule = new ExportableCaretValueRule(path);
       caretValueRule.caretPath = key;
-      caretValueRule.value = getFSHValue(i, remainingFlatElementArray, 'ElementDefinition', fisher);
+      caretValueRule.value = getFSHValue(
+        i,
+        remainingFlatElementArray,
+        'ElementDefinition',
+        entityPath,
+        fisher
+      );
       // If the value is empty, we can't use it. Log an error and give up on trying to use this key.
       if (isFSHValueEmpty(caretValueRule.value)) {
         logger.error(
@@ -207,7 +214,7 @@ export class CaretValueRuleExtractor {
         }
         const caretValueRule = new ExportableCaretValueRule('');
         caretValueRule.caretPath = key;
-        caretValueRule.value = getFSHValue(i, flatArray, 'StructureDefinition', fisher);
+        caretValueRule.value = getFSHValue(i, flatArray, 'StructureDefinition', input.name, fisher);
         if (isFSHValueEmpty(caretValueRule.value)) {
           logger.error(
             `Value in StructureDefinition ${input.name} for element ${key} is empty. No caret value rule will be created.`
@@ -241,7 +248,13 @@ export class CaretValueRuleExtractor {
       }
       const caretValueRule = new ExportableCaretValueRule('');
       caretValueRule.caretPath = key;
-      caretValueRule.value = getFSHValue(i, flatArray, resourceType, fisher);
+      caretValueRule.value = getFSHValue(
+        i,
+        flatArray,
+        resourceType,
+        input.name ?? input.id,
+        fisher
+      );
       if (isFSHValueEmpty(caretValueRule.value)) {
         logger.error(
           `Value in ${resourceType} ${
@@ -272,7 +285,7 @@ export class CaretValueRuleExtractor {
     flatArray.forEach(([key], i) => {
       const caretValueRule = new ExportableCaretValueRule('');
       caretValueRule.caretPath = key;
-      caretValueRule.value = getFSHValue(i, flatArray, 'Concept', fisher);
+      caretValueRule.value = getFSHValue(i, flatArray, 'Concept', entityName, fisher);
       caretValueRule.isCodeCaretRule = true;
       caretValueRule.pathArray = [...pathArray];
       if (isFSHValueEmpty(caretValueRule.value)) {
--- a/src/extractor/CaretValueRuleExtractor.ts
+++ b/src/extractor/CaretValueRuleExtractor.ts
@@ -285,7 +285,13 @@ export class CaretValueRuleExtractor {
     flatArray.forEach(([key], i) => {
       const caretValueRule = new ExportableCaretValueRule('');
       caretValueRule.caretPath = key;
-      caretValueRule.value = getFSHValue(i, flatArray, 'Concept', entityName, fisher);
+      caretValueRule.value = getFSHValue(
+        i,
+        flatArray,
+        'Concept',
+        `${entityName} ${pathArray.join('.')}`,
+        fisher
+      );
       caretValueRule.isCodeCaretRule = true;
       caretValueRule.pathArray = [...pathArray];
       if (isFSHValueEmpty(caretValueRule.value)) {
--- a/src/processor/InstanceProcessor.ts
+++ b/src/processor/InstanceProcessor.ts
@@ -114,7 +114,7 @@ export class InstanceProcessor {
         i,
         flatInstanceArray,
         instanceOfJSON.type,
-        target.name,
+        target.id,
         fisher
       );
       // if the value is empty, we can't use that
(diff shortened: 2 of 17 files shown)


cmoesel (changes requested):
This works well. Thank you!

In regard to your question, I think we _should_ report out the name of the entity for which we found the problem; otherwise it's more difficult for the user to find the issue.

In the extractor, we already have other cases where we do this by passing in the SD with the ED. In getFSHValue, it looks like you'll need to add an additional argument (as you noted).

src/processor/InstanceProcessor.ts
```diff
@@ -110,7 +110,13 @@ export class InstanceProcessor {
     const flatInstanceArray = toPairs(getPathValuePairs(inputJSON));
     flatInstanceArray.forEach(([path], i) => {
       const assignmentRule = new ExportableAssignmentRule(path);
-      assignmentRule.value = getFSHValue(i, flatInstanceArray, instanceOfJSON.type, fisher);
+      assignmentRule.value = getFSHValue(
+        i,
+        flatInstanceArray,
+        instanceOfJSON.type,
+        target.name,
```
jafeltra, line 117: It looks like `target.name` is `{name}-of-{instanceOf}`, so the warning message says something like `MyObservation-of-Observation`, which is not super familiar to users and I don't think appears in the output FSH or input JSON. I didn't check enough to know if either of these are consistently available, but from my quick check, it looked like it might be possible to use `target.id` and `target.instanceOf` separately to get a little more readable version of the name in the warning, so the warning could have something like `MyObservation of Observation`.

jafeltra (commented):
It is a little bit of a bummer that we need the set of warnings in 3 different places, but that isn't really because of your branch, so oh well I guess.

I left a few questions in line, let me know if you want to chat about any of them.

src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -272,7 +285,7 @@ export class CaretValueRuleExtractor {
     flatArray.forEach(([key], i) => {
       const caretValueRule = new ExportableCaretValueRule('');
       caretValueRule.caretPath = key;
-      caretValueRule.value = getFSHValue(i, flatArray, 'Concept', fisher);
+      caretValueRule.value = getFSHValue(i, flatArray, 'Concept', entityName, fisher);
```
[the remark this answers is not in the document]
mint-thompson: Ah, it should definitely be including more path information here! The full `entityPath` should accomplish that, so I'll update this.

src/processor/InstanceProcessor.ts
mint-thompson: `id` definitely seems like the more useful information to give, since many Instances aren't going to have meaningful names. I'll update this to give a better reference.

src/processor/InstanceProcessor.ts
```diff
@@ -110,7 +110,13 @@ export class InstanceProcessor {
     const flatInstanceArray = toPairs(getPathValuePairs(inputJSON));
     flatInstanceArray.forEach(([path], i) => {
       const assignmentRule = new ExportableAssignmentRule(path);
-      assignmentRule.value = getFSHValue(i, flatInstanceArray, instanceOfJSON.type, fisher);
+      assignmentRule.value = getFSHValue(
+        i,
+        flatInstanceArray,
+        instanceOfJSON.type,
+        target.id,
```
jafeltra, line 117: I think you can technically have ids that are the same if the resource type is different. Do you think it is worth making the name that gets logged here be something like `{target.instanceOf} {target.id}` or `{target.id} of {target.instanceOf`?

test/extractor/CaretValueRuleExtractor.test.ts
```diff
@@ -1042,5 +1042,19 @@ describe('CaretValueRuleExtractor', () => {
...
+      const caretRules = CaretValueRuleExtractor.process(element, looseSD, defs);
+      const expectedRule = new ExportableCaretValueRule('component.valueDateTime');
+      expectedRule.caretPath = 'patternDateTime';
+      expectedRule.value = '2013-01-01 00:00:00.000';
+      expect(caretRules).toHaveLength(1);
+      expect(caretRules[0]).toEqual<ExportableCaretValueRule>(expectedRule);
+      expect(loggerSpy.getAllMessages('warn')).toHaveLength(1);
+      expect(loggerSpy.getLastMessage('warn')).toMatch(
+        /Value 2013-01-01 00:00:00\.000 on ObservationWithCaret\.component\.valueDateTime element patternDateTime is not a valid FHIR dateTime/s
+      );
```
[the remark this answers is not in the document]
jafeltra: I think I'd still like confirmation that running the `caret-value-profile.json` with this branch gives the same warning message as this test is expecting. I am assuming I'm doing something wrong, but I sort of just want to know that it is me.

src/extractor/CaretValueRuleExtractor.ts
```diff
@@ -285,7 +285,13 @@ export class CaretValueRuleExtractor {
     flatArray.forEach(([key], i) => {
       const caretValueRule = new ExportableCaretValueRule('');
       caretValueRule.caretPath = key;
-      caretValueRule.value = getFSHValue(i, flatArray, 'Concept', entityName, fisher);
+      caretValueRule.value = getFSHValue(
+        i,
+        flatArray,
+        'Concept',
+        `${entityName} ${pathArray.join('.')}`,
```
jafeltra, line 292: I just want to double check on what the `pathArray` provides in this case to make sure it's what we want. For the original example reported on FSH Online, the `pathArray` just contains the code, so the warning message becomes:
> Value 2023-12-31 00:00:00.000 on MaterCS #297375 element property[1].valueDateTime is not a valid FHIR dateTime

For a code system that has hierarchical codes, the warning message could becomes something like this:
> Value 2013/01/01 on AnteaterCS #Anteater.#Tamandua.#SouthernTamandua element property[0].valueDateTime is not a valid FHIR dateTime

I think they're both technically correct and an improvement from before because they do point you to the code where the issue is. But the element paths are much more accurate, showing the actual property names and indices you would see in the JSON. Hopefully someone would know to look in the `concept` array (and within nested `concept` arrays when needed, but it just felt a little different.

All that said, maybe `pathArray` is as close as we can get here, but I wanted to double check.

jafeltra (commented):
I just had a couple more questions.

src/extractor/CaretValueRuleExtractor.ts
mint-thompson: This call to `getFSHValue` happens within `CaretValueRuleExtractor.processConcept`. In this function, we have the `pathArray`, but not the actual JSON path. For both CodeSystems and ValueSets, we stop tracking JSON paths more or less immediately. To get the JSON path here would require passing it along through several function calls. While the concept path doesn't directly tell you where to look, it does let you know which concept to search for in your input file, which I think is probably still enough information to quickly get at the value that generated the warning.

test/extractor/CaretValueRuleExtractor.test.ts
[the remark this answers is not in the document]
mint-thompson: The warnings are different, and there's a good reason why: when actually running GoFSH, the element first goes through the `AssignmentRuleExtractor`, which generates a warning. The path is marked as processed, and the `CaretValueRuleExtractor` never looks at the value. I'll modify the test to make this a bit less confusing.

src/extractor/CaretValueRuleExtractor.ts
jafeltra: That makes sense, thanks for the extra explanation. Good to know that this is the best we can do, and it is still helpful.

test/extractor/CaretValueRuleExtractor.test.ts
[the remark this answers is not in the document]
jafeltra: Ahhh okay, that makes sense why I was seeing different behavior. If you don't think it is necessary to edit the test, I don't want to make more work if it is only because I wasn't sure how things were working all together.

jafeltra (approved):
This looks good to me! Thanks for making those updates.

cmoesel (approved):
Thanks Mint! I bet this ended up hitting a lot more lines of code than you were expecting! Those pesky error messages!

PR #270, opened 2024-10-04
Construct inline instances from within more FSH entity types

### LINKED ISSUE #252: GoFSH handles inline CodeSystems poorly
Given:
```
{
  "resourceType": "ValueSet",
  "id": "example-valueset",
  "contained": [
    {
      "resourceType": "CodeSystem",
      "id": "example-codesystem",
      "url": "http://example.org/codesystem",
      "version": "1.0.0",
      "content": "complete",
      "status": "active",
      "concept": [
        {
          "code": "example-code-1",
          "display": "Example Code 1"
        },
        {
          "code": "example-code-2",
          "display": "Example Code 2"
        },
        {
          "code": "example-code-3",
          "display": "Example Code 3"
        }
      ]
    }
  ],
  "url": "http://example.org/valueset",
  "version": "1.0.0",
  "status": "active",
  "compose": {
    "include": [
      {
        "system": "#example-codesystem"
      }
    ]
  }
}
```
GoFSH produces:
```
ValueSet: ExampleValueset
Id: example-valueset
* ^contained.resourceType = "CodeSystem"
* ^contained.id = "example-codesystem"
* ^contained.url = "http://example.org/codesystem"
* ^contained.version = "1.0.0"
* ^contained.content = #complete
* ^contained.concept[0] = #example-code-1 "Example Code 1"
* ^contained.concept[+] = #example-code-2 "Example Code 2"
* ^contained.concept[+] = #example-code-3 "Example Code 3"
* ^url = "http://example.org/valueset"
* ^version = "1.0.0"
* ^status = #active
* include codes from system #example-codesystem
```
There are multiple issues with this:
1. Instead of defining the whole CodeSystem inline (which SUSHI currently errors on, see [SUSHI#1404](https://github.com/FHIR/sushi/issues/1404)), it should probably create a separate `Instance` w/ `Usage: #inline`.
2. Since `contained.concept` is _not_ a `code/Coding/CodeableConcept`, I don't think it's valid to assign a FSH code (and SUSHI rightfully complains). If using `Instance`/`caret` syntax for `contained.concept`, the `code` and `display` need to be assigned separately.
3. SUSHI doesn't handle `include codes from system #example-codesystem` well (due to the `#`). SUSHI probably _should_ handle this, but we need to track [SUSHI#1403](https://github.com/FHIR/sushi/issues/1403) and only export the `#` if SUSHI supports it. It may be better practice to just export the id (although SUSHI also doesn't handle this correctly per [SUSHI#1402](https://github.com/FHIR/sushi/issues/1402)).

mint-thompson:
**Description:**
Logical models, Resources, CodeSystems, and ValueSets are now checked for rules that could be used to define an inline instance. Since this means those entities will now contain rules that assign inline instances, check those entity types when simplifying instance names.

**Testing Instructions:**
Confirm that the new test cases demonstrate that inline instances are created for rules contained on the newly added FSH entity types. Confirm that the sample JSON in the related issue now results in an inline instance of CodeSystem. Note that a round trip will not be successful with the sample JSON, since `#example-codesystem` is not a value that should be used for referring to a contained CodeSystem. The correct output for this JSON is shown in https://github.com/FHIR/sushi/pull/1520.

**Related Issue:**
Fixes #252

### DIFF
--- a/src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
+++ b/src/optimizer/plugins/ConstructInlineInstanceOptimizer.ts
@@ -31,9 +31,16 @@ export default {
 
   optimize(pkg: Package, fisher: MasterFisher, options: ProcessingOptions = {}): void {
     const inlineInstances: ExportableInstance[] = [];
-    // Only construct inline instances in instances, profiles, and extensions.
     // If we add support for other types, update SimplifyInstanceNameOptimizer as well.
-    [...pkg.instances, ...pkg.profiles, ...pkg.extensions].forEach(resource => {
+    [
+      ...pkg.instances,
+      ...pkg.profiles,
+      ...pkg.extensions,
+      ...pkg.logicals,
+      ...pkg.resources,
+      ...pkg.codeSystems,
+      ...pkg.valueSets
+    ].forEach(resource => {
       const ruleType =
         resource instanceof ExportableInstance
           ? ExportableAssignmentRule
--- a/src/optimizer/plugins/SimplifyInstanceNameOptimizer.ts
+++ b/src/optimizer/plugins/SimplifyInstanceNameOptimizer.ts
@@ -35,7 +35,15 @@ export default {
     });
 
     // Fix up inline assignments in types that may have them (see: ConstructInlineInstanceOptimizer)
-    [...pkg.instances, ...pkg.profiles, ...pkg.extensions].forEach(resource => {
+    [
+      ...pkg.instances,
+      ...pkg.profiles,
+      ...pkg.extensions,
+      ...pkg.logicals,
+      ...pkg.resources,
+      ...pkg.codeSystems,
+      ...pkg.valueSets
+    ].forEach(resource => {
       const inlineAssignmentRules = resource.rules.filter(
         rule =>
           (rule instanceof ExportableAssignmentRule || rule instanceof ExportableCaretValueRule) &&
--- a/test/optimizer/plugins/ConstructInlineInstanceOptimizer.test.ts
+++ b/test/optimizer/plugins/ConstructInlineInstanceOptimizer.test.ts
@@ -5,9 +5,13 @@ import { Package } from '../../../src/processor';
 import {
   ExportableAssignmentRule,
   ExportableCaretValueRule,
+  ExportableCodeSystem,
   ExportableExtension,
   ExportableInstance,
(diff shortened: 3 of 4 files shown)


jafeltra (approved):
This looks pretty good to me. Those comments pointing to the two optimizers were very helpful!

I just have one question for my own understanding about what you said in the PR description about that third item on the issue:

> Note that a round trip will not be successful with the sample JSON, since `#example-codesystem` is not a value that should be used for referring to a contained CodeSystem.

Do you mean that _right now_ `#example-codesystem` is not a valid value to be used to refer to a contained CodeSystem? We do still have FHIR/SUSHI#1403 to support that, so should we create an optimizer for GoFSH for that extension pattern once that syntax is supported in SUSHI?

I think any answer to that question doesn't affect this PR, so I'm going to approve in the meantime!

mint-thompson:
> Do you mean that _right now_ `#example-codesystem` is not a valid value to be used to refer to a contained CodeSystem? We do still have [FHIR/sushi#1403](https://github.com/FHIR/sushi/issues/1403) to support that, so should we create an optimizer for GoFSH for that extension pattern once that syntax is supported in SUSHI?

Yes, that's correct. The JSON in the linked issue is not correct according, since it should be using the contained CodeSystem's `url` as the value of `system`, and should also contain the `valueset-system` extension as seen in https://github.com/FHIR/sushi/pull/1520. With that JSON, the fshing trip is successful.

As a related issue: right now, the generated FSH for the correct ValueSet will be an Instance, not a ValueSet. This is because the presence of the `valueset-system` extension looks like something that would need to be manually assigned. But, since we know SUSHI will put that in there when we need it, we don't need rules for it as long as we can turn it into ValueSet definition. Do you think making that change should be part of this PR, or should it be a separate issue?

mint-thompson:
> As a related issue: right now, the generated FSH for the correct ValueSet will be an Instance, not a ValueSet. This is because the presence of the `valueset-system` extension looks like something that would need to be manually assigned. But, since we know SUSHI will put that in there when we need it, we don't need rules for it as long as we can turn it into ValueSet definition. Do you think making that change should be part of this PR, or should it be a separate issue?

After some discussion with the team, I think it's fair to say that this optimization is a separate issue. I've created #271 to track this.

cmoesel (approved):
This looks good to me as well. Thank you, @mint-thompson!

PR #272, opened 2024-10-11
Support FHIR R6

jafeltra:
**Description:** Add support for using FHIR R6 versions. This mostly just switched from using our old regex if/else statement to using the nice util function from SUSHI for different supported FHIR versions.

**Testing Instructions:** Make sure the tests pass and you can run GoFSH on a project that uses a FHIR R6 version.

**Related Issue:** N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -18,7 +18,7 @@
         "fhir-package-loader": "^1.0.0",
         "flat": "^5.0.2",
         "fs-extra": "^11.2.0",
-        "fsh-sushi": "^3.11.0",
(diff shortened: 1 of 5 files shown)


cmoesel (approved):
This looks good! I tested it against a homemade R6 IG that has a profile on DeviceAlert (which is new in R6) and everything went swimmingly! Thanks, @jafeltra!

mint-thompson (approved):
Tests look fine, and I was able to process JSON with R6 definitions. Thank you!

PR #275, opened 2024-11-06
Process ValueSet that uses components from a contained system

### LINKED ISSUE #271: Optimize ValueSets that refer to contained CodeSystems
When a ValueSet refers to a contained CodeSystem, the [valueset-system](https://hl7.org/fhir/extensions/StructureDefinition-valueset-system.html) extension will be present on the `ValueSet.compose.include.system` element. Since the rules on a FSH ValueSet don't allow for setting extensions at this path, GoFSH will produce a FSH Instance of ValueSet rather than a FSH ValueSet. However, this extension is automatically added by SUSHI when a ValueSet compose rule refers to a contained CodeSystem. Therefore: if the only rules on a ValueSet that would normally require GoFSH to create a FSH Instance are for these extensions, then it's safe to produce a FSH ValueSet instead.

mint-thompson:
**Description:**
When a ValueSet component references a contained system, the valueset-system extension should be present at compose.system.extension. When it is present, it is safe to throw it out, since SUSHI will add it back in automatically. When it is removed, also remove the extension list and containing underscore property if those become empty. This will often allow the input to be processed as a ValueSet.

This change is simpler than I expected, but this is largely due to the work on SUSHI to better handle various cases involving contained resources.

**Testing Instructions:**
Confirm that the new test and accompanying test fixture accurately reflect a ValueSet that references a contained system. Run a fshing trip with input that includes a ValueSet that references a contained system and confirm that the ValueSet is unchanged. 

**Related Issue:**
Fixes #271.

### DIFF
--- a/src/processor/ValueSetProcessor.ts
+++ b/src/processor/ValueSetProcessor.ts
@@ -1,5 +1,5 @@
 import { utils, fhirtypes, fshtypes } from 'fsh-sushi';
-import { capitalize, compact, difference } from 'lodash';
+import { capitalize, compact, difference, isEmpty } from 'lodash';
 import { flatten } from 'flat';
 import { ExportableValueSet } from '../exportable';
 import {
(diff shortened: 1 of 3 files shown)


jafeltra (approved):
This change makes sense to me and works as described!

cmoesel (approved):
I like it!

PR #277, opened 2024-12-20
Use new FPL API

mint-thompson:
**Description:**
The FPL version 2 package includes API changes. Update classes, functions, and tests that include loaded dependencies to use the new API. Most of this is acquired through SUSHI's FHIRDefinitions class, which extends FPL's BasePackageLoader class. The main thing to watch out for is that FHIRDefinitions must be initialized before use, which is an async operation.

The LakeOfFHIR class is updated to have an instance of FHIRDefinitions for access to its fishing methods. Therefore, it also needs to be initialized asynchronously. There may be some possible redesigns of the LakeOfFHIR class in order to more fully make use of the new FPL API.

**Testing Instructions:**
~~This currently needs the `new-fpl` branch of SUSHI as a dependency. Until this branch is part of a SUSHI release, you'll need to make that dependency yourself. Pull down the `new-fpl` branch of SUSHI, build it, `npm pack` it into a `.tgz` file, then use that as the dependency. Until the dependency is available, this PR will be a draft.~~
Update 2024/12/24: the updates to SUSHI and FPL are now available in published releases. The dependencies in this PR are updated accordingly.

There are a lot of changes to the tests due to the FPL API changes. There should be no changes in output FSH resulting from this change, so try it out with some FHIR resources and make sure the FSH you get is the same before and after the changes.

**Related Issue:**
N/A


src/processor/LakeOfFHIR.ts
```diff
@@ -5,11 +5,16 @@ import { CONFORMANCE_AND_TERMINOLOGY_RESOURCES } from './InstanceProcessor';
...
-import { FHIRDefinitions, logger } from '../utils';
+import { FHIRDefinitions, logger, logMessage } from '../utils';
+import { InMemoryVirtualPackage } from 'fhir-package-loader';
 
 // Like FSHTank in SUSHI, but it doesn't contain FSH, it contains FHIR.  And who ever heard of a tank of FHIR?  But a lake of FHIR...
 export class LakeOfFHIR implements utils.Fishable {
-  constructor(public docs: WildFHIR[]) {}
+  readonly defs: FHIRDefinitions;
+
+  constructor(public docs: WildFHIR[]) {
```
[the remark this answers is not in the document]
cmoesel: Ah. I see now that we need it for `restockLake` in the tests, so I guess it probably needs to stay as-is.

test/optimizer/plugins/ResolveValueSetComponentRuleURLsOptimizer.test.ts
```diff
@@ -36,49 +33,73 @@ describe('optimizer', () => {
...
       const myPackage = new Package();
       myPackage.add(valueset);
+      await restockLake(
+        lake,
+        path.join(__dirname, 'fixtures', 'simple-codesystem.json'),
+        path.join(__dirname, 'fixtures', 'unsupported-codesystem.json'),
+        path.join(__dirname, 'fixtures', 'simple-valueset.json'),
+        path.join(__dirname, 'fixtures', 'unsupported-valueset.json')
+      );
+      fisher = new MasterFisher(lake, defs);
```
cmoesel, line 50: I expect that I am missing something, but... why do we have to constantly restock the lake in every test? Why don't we stock it in `beforeAll` like we used to? I don't see any modifications to the resources in the lake -- but maybe I'm missing it.

If this does indeed need to happen in every test, I think you should extract it into a function like `getRestockedFisher` (or something like that) since it looks like this logic is repeated a lot across the tests.

test/utils/Processing.test.ts
```diff
@@ -743,9 +756,10 @@ describe('Processing', () => {
 
     it('should load FHIR R6 prerelease if specified', () => {
       const defs = new FHIRDefinitions();
+      // jest.spyOn(defs, 'loadPackage').mockImplementation(replacementLoadPackage);
```
cmoesel, line 759: Can this commented code be deleted?

cmoesel (changes requested):
This looks good, and thankfully doesn't seem to have affected _too many_ things. I've left a few comments for your consideration. I've also noticed that when I run the tests, I'm now getting a lot of log messages about loading a virtual package. I think the spy used to suppress those but it seems it is not anymore. Have you looked into it?

test/optimizer/plugins/ResolveValueSetComponentRuleURLsOptimizer.test.ts
mint-thompson: The set of resources is slightly different in some of the tests, which is why I pulled it out of `beforeAll`. I'll make a helper function like you suggested.

mint-thompson:
I've made updates based on review comments from @cmoesel , and since the published SUSHI and FPL dependencies are now available, I've marked this PR as ready for review.

cmoesel (dismissed):
Thank you, Mint. This looks great and works well when running against other IGs.

src/processor/LakeOfFHIR.ts
```diff
@@ -111,32 +116,48 @@ export class LakeOfFHIR implements utils.Fishable {
...
   fishForFHIR(item: string, ...types: utils.Type[]) {
     // The simplest approach is just to re-use the FHIRDefinitions fisher.  But since this.docs can be modified by anyone at any time
     // the only safe way to do this is by rebuilding a FHIRDefinitions object each time we need it.  If this becomes a performance
     // concern, we can optimize it later -- but performance isn't a huge concern in GoFSH. Note also that this approach may need to be
     // updated if we ever need to support fishing for Instances.
-    const defs = new FHIRDefinitions();
-    this.docs.forEach(d => defs.add(d.content));
-    return defs.fishForFHIR(item, ...types);
+    return this.defs.fishForFHIR(item, ...types);
   }
```
jafeltra, line 145: Is this comment still applicable since we're not re-building a FHIRDefinitions object each time anymore?

src/processor/LakeOfFHIR.ts
```diff
@@ -111,32 +116,48 @@ export class LakeOfFHIR implements utils.Fishable {
...
   fishForMetadata(item: string, ...types: utils.Type[]): utils.Metadata {
     // The simplest approach is just to re-use the FHIRDefinitions fisher.  But since this.docs can be modified by anyone at any time
     // the only safe way to do this is by rebuilding a FHIRDefinitions object each time we need it.  If this becomes a performance
     // concern, we can optimize it later -- but performance isn't a huge concern in GoFSH. Note also that this approach may need to be
     // updated if we ever need to support fishing for Instances.
-    const defs = new FHIRDefinitions();
-    this.docs.forEach(d => defs.add(d.content));
-    return defs.fishForMetadata(item, ...types);
+    return this.defs.fishForMetadata(item, ...types);
   }
```
jafeltra, line 153: Similar question about the comment here as above.

jafeltra (commented):
I just left a couple minor questions I had while reviewing. I'm not sure they'll even require any updating.

jafeltra (approved):
Looks good to me!

cmoesel (approved):
Looks good to me too!

PR #279, opened 2024-12-31
v2.5.0

cmoesel:
**Description:** This bumps the version to 2.5.0. I decided to call this a minor release because although it doesn't add any significant new features, it's integration of the new FPL is still pretty significant. It didn't feel like something that should be snuck into a patch release.

**Testing Instructions:** Ensure tests pass.

**Related Issue:** N/A

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "gofsh",
-  "version": "2.4.0",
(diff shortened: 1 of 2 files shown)


jafeltra (approved):
I think this makes sense as a minor release. This isn't a bugfix, so I think it is fine to not be a patch.

PR #281, opened 2025-03-28
Dependency Updates March 2025

cmoesel:
**Description:** This PR updates dependencies to their latest versions. These updates did not require any changes to the code. I've regression tested using US Core and the results are equivalent to the currently released version.

**Testing Instructions:** Ensure GoFSH works as expected.


jafeltra (approved):
Works as expected!

PR #285, opened 2025-12-27
Dependency Updates (2025)

cmoesel:
**Description:** This PR updates all dependencies that can be reasonably updated. See DEPENDENCY-NOTES.md for exceptions. As part of this PR, I've also bumped our preferred Node version to v22, but we still maintain support for Node 18 at the moment.

I've also turned off coverage for CI runs.

**Testing Instructions:** Tests should pass and GoFSH should continue to function as normal.

**Related Issue:** N/A


cmoesel:
Bypassing rules to merge this since it is mainly dependency updates and tests/regressions have passed.

